PR #57, opened 2021-10-27
SCP-2945: Only script outputs in unbalanced transactions

sjoerdvisscher:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-ledger/src/Ledger/Constraints/OffChain.hs
```diff
@@ -163,14 +162,37 @@ pubKey pk = mempty { slPubKeyHashes = Map.singleton (pubKeyHash pk) pk }
...
+    = Nothing
+
+fromScriptOutput :: ScriptOutput -> ChainIndexTxOut
+fromScriptOutput (ScriptOutput vh v dh) =
+    Tx.ScriptChainIndexTxOut (Address.scriptHashAddress vh) (Left vh) (Left dh) v
+
+instance Pretty ScriptOutput where
+    pretty ScriptOutput{scriptOutputValidatorHash, scriptOutputValue} =
+        hang 2 $ vsep ["-" <+> pretty scriptOutputValue <+> "addressed to", pretty scriptOutputValidatorHash]
+
```
koslambrou, line 187: Maybe put this in `Ledger.Tx` together with `ChainIndexTxOut`?

plutus-contract/src/Wallet/Emulator/Wallet.hs
```diff
@@ -268,7 +268,7 @@ validateTxAndAddFees feeCfg slotCfg ownTxOuts utx = do
     -- Balance and sign just for validation
     tx <- handleBalanceTx ownTxOuts utx
     signedTx <- handleAddSignature tx
-    let utxoIndex        = Ledger.UtxoIndex $ unBalancedTxUtxoIndex utx <> (toTxOut <$> ownTxOuts)
+    let utxoIndex        = Ledger.UtxoIndex $ fmap toTxOut $ (U.fromScriptOutput <$> unBalancedTxUtxoIndex utx) <> ownTxOuts
```
sjoerdvisscher, line 271: @j-mueller After thinking some more this here seems to be the clue that there is no failure. The PubKeys are only unavailable in the lookups. But the wallet knows how to lookup its own outputs (`ownTxOuts` here) so `getUnspentOutput` just works!

j-mueller: `getUnspentOutput` works, and it gives you an unspent output from the wallet. But there is no guarantee that the wallet selects the same unspent output when `balanceTx` is called a second time, for the transaction that creates the token by spending this specific output. 

The UTXO could have been spent in the meantime, or maybe the wallet decides that there is a different collection of UTXOs that is better suited in the second call to `balanceTx`.

j-mueller: Because of this, I'm inclined to think that we have to go back to creating a script output (with the PubKey contract) in a separate transaction before we can set up the state machine. What do you think @sjoerdvisscher  @silky  ?

sjoerdvisscher: @j-mueller I agree, but that’s a separate issue I think? So we can merge this one?

j-mueller: Yes, let's merge it!

PR #68, opened 2021-10-31
Chain Index app: query the tip of the node for first response before start

kk-hainq:
This work is part of #69.

The printed slot number is only half helpful. The primary purpose of this query is to get the first response of the node for potential errors before opening the DB and starting the chain index.

When the socket does not exist:
```
Query the tip of the local node: 
plutus-chain-index: Network.Socket.connect: <socket: 78>: does not exist (No such file or directory)
```

When the network magics do not match:
```
Query the tip of the local node: 
plutus-chain-index: HandshakeError (Refused NodeToClientV_9 "version data mismatch: NodeToClientVersionData {networkMagic = NetworkMagic {unNetworkMagic = 1097911063}} /= NodeToClientVersionData {networkMagic = NetworkMagic {unNetworkMagic = 1097911062}}")
```

When it works:
```
Query the tip of the local node:
SlotNo 41320486
```

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-chain-index/app/Main.hs
+++ b/plutus-chain-index/app/Main.hs
@@ -34,8 +34,10 @@ import qualified Cardano.BM.Configuration.Model  as CM
 import           Cardano.BM.Setup                (setupTrace_)
 import           Cardano.BM.Trace                (Trace, logDebug, logError)
 
-import           Cardano.Api                     (ChainPoint)
+import           Cardano.Api                     (ChainPoint, ChainTip (..), ConsensusModeParams (..),
+                                                  LocalNodeConnectInfo (..), getLocalChainTip)
 import           Cardano.Protocol.Socket.Client  (ChainSyncEvent (..), runChainSync)
+import           Cardano.Protocol.Socket.Type    (epochSlots)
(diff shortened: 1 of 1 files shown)


plutus-chain-index/app/Main.hs
```diff
@@ -135,6 +137,17 @@ main = do
       putStrLn "\nChain Index config:"
       print (pretty config)
 
+      -- The printed slot number is only half helpful.
+      -- The primary purpose of this query is to get the first response of the node for potential errors before opening the DB and starting the chain index.
+      -- See #69.
+      putStrLn "\nQuery the tip of the local node:"
```
sjoerdvisscher, line 143:
```suggestion
      putStr "\nThe tip of the local node: "
```

sjoerdvisscher (approved):
Thanks! I just have a small suggestion regarding the message.

plutus-chain-index/app/Main.hs
kk-hainq: Thanks! I've updated the message as suggested.

PR #72, opened 2021-10-31
Chain Index: users can configure to only store txs from a block no onward

kk-hainq:
This PR is a quick proof-of-concept for #73 on allowing users to choose which batch of transactions to store in the DB. In this specific instance, the application can choose to only store transactions from Alonzo onward. Given that tip and UTXO processing is unaffected, most of these configurations are pretty safe to the functioning of the chain index?

I attempted this as a dApp developer wanting to configure the chain index, hence keeping as many decisions in `app` as possible instead of extending the chain index effects.

If this makes sense, I would love to clean it and add a few tests first. A possible next step will then be listing exclusive script addresses to track.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


silky:
@kk-hainq Thanks for proposing this! It looks interesting.

What about instead of specifying an era, we specify a block number, and only start storing transactions from that point forward? This would make the code a bit more generic. Then, you can provide some hints on how to run that for the specific eras, etc.

Thoughts?

kk-hainq:
@silky Awesome! it is not just more generic and clean for the code and users, it allows dApps to specify the block that they launch at for swift syncing before that! I have updated the code as suggested.

plutus-chain-index-core/src/Plutus/ChainIndex/Effects.hs
```diff
@@ -78,7 +78,11 @@ makeEffect ''ChainIndexQueryEffect
 data ChainIndexControlEffect r where
 
     -- | Add a new block to the chain index by giving a new tip and list of tx.
-    AppendBlock :: Tip -> [ChainIndexTx] -> ChainIndexControlEffect ()
+    --
+    -- The extra `Bool` determines if we want to store this batch of transactions.
+    -- If `True` then yes, else only handle the tip and UTXOs.
+    -- This, for example, enables applications to skip unneeded pre-Alonzo transactions.
+    AppendBlock :: Tip -> [ChainIndexTx] -> Bool -> ChainIndexControlEffect ()
```
silky, line 85: To make a code a bit more descriptive, can we get an enum instead of the `Bool`; something like:

```
data BlockStoreOption = Save | DontSave
```

(feel free to come up with a better name!) Then it's a bit clearer at a glance what the `appendBlock ... True` calls are doing.

kk-hainq: Sorry I couldn't get back to this earlier. I've added 3 more commits addressing your suggestions. it does look way better going forward now, thanks!

silky (approved):

silky (approved):

silky:
Thanks @kk-hainq ! :)

PR #76, opened 2021-11-01
SCP-2880: Transaction outputs should contain a minimum of 2 Ada.

koslambrou:
Enforce constraint from the real Cardano network where each transaction output should contain a minimum amount (here 2) of Ada.

* Changed the transaction validation code in Ledger.Index to check this minimum deposit.

* Changed the transaction construction code in Ledger.Constraints.OffChain to add enough Ada for the minimum deposit to all transaction outputs of the unbalanced tx.

* Changed the emulator Wallet handler to correctly handle this constraints when balancing a transaction.

* Adapted the contracts in plutus-use-cases to work with this new constraint.

* The command `cabal run plutus-use-cases-scripts – ./tmp transactions -p ./plutus-use-cases/scripts/protocol-parameters.json` does not produce warnings anymore.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-ledger/src/Ledger/Constraints/OnChain.hs
```diff
@@ -40,10 +41,12 @@ checkOwnOutputConstraint
...
     let hsh = V.findDatumHash (Datum $ toBuiltinData ocDatum) scriptContextTxInfo
         checkOutput TxOut{txOutValue, txOutDatumHash=Just svh} =
-            txOutValue == ocValue && hsh == Just svh
+               Ada.fromValue txOutValue >= Ada.fromValue ocValue
+            && Value.noAdaValue txOutValue == Value.noAdaValue ocValue
+            && hsh == Just svh
         checkOutput _       = False
     in traceIfFalse "L1" -- "Output constraint"
-    $ any checkOutput (V.getContinuingOutputs ctx)
+    $ null (V.getContinuingOutputs ctx) || any checkOutput (V.getContinuingOutputs ctx)
```
koslambrou, line 49: Mmm... not sure if this is correct.

plutus-ledger/src/Ledger/Index.hs
```diff
@@ -374,6 +400,11 @@ checkFeeIsAda t =
 minFee :: Tx -> V.Value
 minFee = const (Ada.lovelaceValueOf 10)
 
+-- | TODO Should be calculated based on the maximum script size permitted on
+-- the Cardano blockchain.
+maxFee :: Ada
+maxFee = Ada.lovelaceOf 1_000_000
```
koslambrou, line 406: This is currently used for testings purposes in plutus-use-cases

plutus-use-cases/src/Plutus/Contracts/Future.hs
```diff
@@ -606,20 +598,6 @@ escrowParams client future ftos FutureSetup{longPK, shortPK, contractStart} =
...
-        getOutcome (Folds.Done a) = a
-        getOutcome e              = Haskell.error $ "not finished: " <> Haskell.show e
-    in
-    either (Haskell.error . Haskell.show) (getOutcome . S.fst')
-        $ Freer.run
-        $ Freer.runError @Folds.EmulatorFoldErr
-        $ Stream.foldEmulatorStreamM fld
-        $ Stream.takeUntilSlot 10
-        $ Trace.runEmulatorStream def setupTokensTrace
-
```
koslambrou, line 622: This was moved to the test file.

plutus-ledger/src/Ledger/Constraints/OffChain.hs
```diff
@@ -328,6 +331,23 @@ mkTx
...
+    => m ()
+adjustMinimumAda = do
+    unbalancedTx . tx . Tx.outputs %= (fmap adjustTxOut . filter (not . Value.isZero . txOutValue))
+
+  where
+    adjustTxOut :: TxOut -> TxOut
+    adjustTxOut txOut =
+      let missingLovelace = max 0 (Ledger.minAdaTxOut - Ada.fromValue (txOutValue txOut))
+       in txOut { txOutValue = txOutValue txOut <> Ada.toValue missingLovelace }
+
```
koslambrou, line 350: Instead of doing it like that, create a function which will be called after the tx was constructed which would adjust the tx outputs to min of 2Ada. Called manually.

plutus-ledger/src/Ledger/Constraints/OnChain.hs
```diff
@@ -40,10 +41,12 @@ checkOwnOutputConstraint
 checkOwnOutputConstraint ctx@ScriptContext{scriptContextTxInfo} OutputConstraint{ocDatum, ocValue} =
     let hsh = V.findDatumHash (Datum $ toBuiltinData ocDatum) scriptContextTxInfo
         checkOutput TxOut{txOutValue, txOutDatumHash=Just svh} =
-            txOutValue == ocValue && hsh == Just svh
+               Ada.fromValue txOutValue >= Ada.fromValue ocValue
```
koslambrou, line 44: Revert to old implementation when adjusting constraints using manually function as said in another comment.

plutus-use-cases/src/Plutus/Contracts/Auction.hs
```diff
@@ -130,7 +132,7 @@ auctionTransition AuctionParams{apOwner, apAsset, apEndTime} State{stateData=old
                 newState =
                     State
                         { stateData = Ongoing HighestBid{highestBid = newBid, highestBidder = newBidder}
-                        , stateValue = apAsset <> Ada.toValue newBid -- and lock the new bid in the script output
+                        , stateValue = apAsset <> Ada.toValue newBid <> Ada.toValue Ledger.minAdaTxOut -- and lock the new bid in the script output
```
koslambrou, line 135: Call `adjustValue` manually.

plutus-use-cases/src/Plutus/Contracts/Governance.hs
```diff
@@ -199,7 +201,7 @@ contract params = forever $ mapError (review _GovError) endpoints where
 
     initLaw = endpoint @"new-law" $ \bsLaw -> do
         let mph = Scripts.forwardingMintingPolicyHash (typedValidator params)
-        void $ SM.runInitialise theClient (GovState (toBuiltin bsLaw) mph Nothing) mempty
+        void $ SM.runInitialise theClient (GovState (toBuiltin bsLaw) mph Nothing) (Ada.toValue Ledger.minAdaTxOut)
```
koslambrou, line 204: Revert to old implementation. Should be fixed automatically when resolving comment from above.

plutus-ledger/src/Ledger/Constraints/OnChain.hs
koslambrou: Added another condition below to make it more precise. However, I can't revert it to the old implementation. Here's why. If 10 Ada was locked in the script address, and the user creates a tx to redeem 9 Ada, the wallet emulator balancing will adjust the transaction to only redeem 8 Ada (so that we can leave 2 Ada in the script address), and then the tx validation will fail with the old `checkOwnOutputConstraint`, because the constraint has an output of 1 Ada and the actual tx has an output of 2 Ada.

plutus-use-cases/src/Plutus/Contracts/Auction.hs
```diff
@@ -130,7 +132,9 @@ auctionTransition AuctionParams{apOwner, apAsset, apEndTime} State{stateData=old
                 newState =
                     State
                         { stateData = Ongoing HighestBid{highestBid = newBid, highestBidder = newBidder}
-                        , stateValue = apAsset <> Ada.toValue newBid -- and lock the new bid in the script output
+                        , stateValue = Value.noAdaValue oldStateValue
+                                    <> Ada.toValue (Ada.fromValue oldStateValue - highestBid)
+                                    <> Ada.toValue newBid -- and lock the new bid in the script output
```
koslambrou, line 137: This new stateValue calculation is more precise, and works better with the StateMachine now that it adjustes the Ada per tx output.

plutus-contract/src/Plutus/Contract/StateMachine.hs
```diff
@@ -430,7 +432,11 @@ runInitialiseWith customLookups customConstraints StateMachineClient{scInstance}
             <> Constraints.unspentOutputs utxo
             <> customLookups
     utx <- either (throwing _ConstraintResolutionError) pure (Constraints.mkTx lookups constraints)
-    submitTxConfirmed utx
+    let adjustedUtx = Constraints.adjustUnbalancedTx utx
+    unless (utx == adjustedUtx) $
+      logWarn @Text $ "Plutus.Contract.StateMachine.runInitialise: "
+                    <> "Found a transaction output value with less than the minimum amount of Ada. Adjusting ..."
```
koslambrou, line 438: Not sure about that warning

koslambrou:
@sjoerdvisscher @j-mueller What's mostly left is maybe reorganize some of the functions in `Plutus.Contract.Request` when that the user can now manually adjust an `UnbalancedTx`. Any suggestions are welcome.

plutus-contract/src/Plutus/Contract/StateMachine/OnChain.hs
```diff
@@ -129,13 +129,13 @@ mkValidator (StateMachine step isFinal check threadToken) currentState input ptx
...
                             newConstraints
-                                { txOwnOutputs=
+                                { txOwnOutputs =
                                     [ OutputConstraint
                                         { ocDatum = newData
                                           -- Check that the thread token value is still there
                                         , ocValue = newValue <> threadTokenValueInner threadToken (ownHash ptx)
                                         }
-                                    ]
+                                    | not (isZero newValue)]
```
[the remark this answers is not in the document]
koslambrou: This was needed because of the GameStateMachine contract.

Here's an use case example which explains the mod:

- `Wallet1` locks 8 Ada to the script address
- `Wallet2` call the `guess` endpoint with the correct secret and wants the full 8 Ada.
- In `GameStateMachine.transition`, the `(Locked, Guess)` transition will generate a new `stateValue` of `mempty` (8 locked Ada - 8 requested Ada).
- Then, `StateMachine.runStep` for the `(Locked, Guess)` transition is executed and it creates an unbalanced tx containing an output of `mempty` Value to the script address. 
- `Wallet.Emulator.Wallet` will validate the balanced transaction and report the error `GameSMError (SMCContractError (WalletError (ValidationError (ScriptFailure (EvaluationError ["L1","Ld","S5","PT5"] "CekEvaluationFailure")))))`

The validation that is used is part of `StateMachine.OnChain.mkValidator` that I modified above. It will call `Ledger.Constraints.OnChain.checkOwnOutputConstraint`. Since the new `stateValue` is `mempty`, `getContinuingOutputs ctx` return an empty, list therefore `checkOwnOutputConstraint` returns `False` (because of `List.any`, thus the validation fails.

Therefore, my mod is simply to not create an output constraint is the new value is empty.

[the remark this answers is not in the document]
sjoerdvisscher: But now it doesn't check for the thread token anymore. And what happens if the value is larger than zero but less than 2 Ada?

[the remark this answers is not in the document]
koslambrou: > But now it doesn't check for the thread token anymore. 
You're right. I should replace it with `not (isZero (newValue <> threadTokenValueInner thread (ownHash ptx))`. I think this is valid. Note that the `runStep` and `mkStep` function will adjust the new state value to contain min of 2Ada is it's not empty. Also note, that I adjusted the `mkStep` function as well. Great job find that.

> And what happens if the value is larger than zero but less than 2 Ada?
So if `Wallet1` locks 8 Ada, and `Wallet2` correctly guesses the secret and want 7Ada, then the new state value will be 1Ada. `mkStep` will generate an adjusted contraint which adds the missing ada (output of 2Ada + thread token), and `runStep` will adjust the outputs of created unbalanced tx.
Makes sense?

plutus-ledger/src/Ledger/Constraints/OnChain.hs
sjoerdvisscher: What I meant during the review was a function that adjusts the OutputConstraints, instead of adjusting the unbalanced tx. I'm not sure if that's still a useful thing to do.

plutus-contract/src/Plutus/Contract/StateMachine.hs
```diff
@@ -517,12 +529,14 @@ mkStep client@StateMachineClient{scInstance} input = do
...
-                                }
-                            | not isFinal ]
+                            let newStateValue = stateValue newState <> SM.threadTokenValueOrZero scInstance
+                                missingLovelace = max 0 (Ledger.minAdaTxOut - Ada.fromValue (stateValue newState))
+                             in [ OutputConstraint
+                                    { ocDatum = stateData newState
+                                      -- Add the thread token value back to the output
+                                    , ocValue = newStateValue <> Ada.toValue missingLovelace
+                                  }
+                                | not isFinal && not (Value.isZero newStateValue) ]
```
sjoerdvisscher, line 539: Ah, this is what I was missing. I don't think we want to throw away the outputconstraint when the value is zero, since we still want to propagate the ocDatum.

plutus-ledger/src/Ledger/Constraints/OnChain.hs
koslambrou: Ahhh ok. Mmm... I'll check that out. Might be better this way than  to adjust unbalanced tx

plutus-contract/src/Plutus/Contract/StateMachine.hs
koslambrou: Then what do we do if the output value is zero? Simply adjusting it to 2 Ada doesn't make much sense, since that would mean that I could never use all funds in a script address.

sjoerdvisscher: And similarly it would not be possible to use all the funds but 1 Ada in a script address right? Maybe we need to discuss this with people who have thought about this more...

plutus-ledger/src/Ledger/Constraints/OnChain.hs
koslambrou: Tried to write a function which adjusts the TxConstraints instead of UnbalancedTx, but it doesn't seem possible. I had the constraint `mustSpendAtLeast <Guess token>` in the `GameStateMachine` contract. In this scenario, there's a tx output which contains the `Guess` token, but I couldn't modify the `TxConstraints` such that a 2Ada would be associated with the token.

plutus-contract/src/Plutus/Contract/StateMachine/OnChain.hs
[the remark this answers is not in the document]
koslambrou: Alright, I managed to find a way to revert this change. What I needed to do was adapt the GameStateMachine contract such that when the locked funds are 0, then we change the state to final, so there will be no more txs.

plutus-contract/src/Plutus/Contract/StateMachine.hs
koslambrou: So currently, we need to adapt contracts like `GameStateMachine` such that if the locked funds are 0, then we need to transition of a final state. Of course, if want all the funds - 1, then the adjusting function will change that to `funds - 2`.

sjoerdvisscher (approved):
Looks good and and impressive job! (Although I have a nagging feeling this won't be the last of it, since it is all quite fiddly)

koslambrou:
> Looks good and and impressive job! (Although I have a nagging feeling this won't be the last of it, since it is all quite fiddly)

Thanks :)

True, it certainly feels fiddly.

PR #77, opened 2021-11-01
PAB: More consistent slot configs

kk-hainq:
This PR attempts to fix a misconfigured slot config issue we found while working with PAB. It started with a wrong Shelley launch time in `plutus-pab/plutus-pab.yaml.sample`, which leads to weird off-chain validation errors when we use `mustValidateIn` constraints.

The quick fix was to update the config file with the "correct" Shelley launch date, but this PR provides the general fix that removes the bug for good.

The flow starts with `mustValidateIn`, which adds the validity range to the `UnbalancedTx`:
https://github.com/input-output-hk/plutus-apps/blob/942bd8c6de6a2d5981d91c704b0258bddd9d9d7c/plutus-ledger/src/Ledger/Constraints/OffChain.hs#L505-L506

Which is eventually balanced here:
https://github.com/input-output-hk/plutus-apps/blob/942bd8c6de6a2d5981d91c704b0258bddd9d9d7c/plutus-contract/src/Wallet/Emulator/Wallet.hs#L202-L208

We can see that the original validity interval is translated to a slot range using the PAB's slot config before being set to `UnbalancedTx`'s internal `tx`'s validity range.

Fast forward through `validateTxAndAddFees`, then `validateTransactionOffChain`, then `checkMintingScripts`, then `mkTxInfo` we end up re-converting the slot range, but with the default slot config!
https://github.com/input-output-hk/plutus-apps/blob/ecc528ec38f42dee056ccb072abd41a64b640425/plutus-ledger/src/Ledger/Index.hs#L396

When the PAB's slot config does not match the default, this mismatch modifies the intended validity interval, and off-chain validates the transaction with it, which effectively makes the config not configurable.
https://github.com/input-output-hk/plutus-apps/blob/f4893f8713d299020d9e6d059dd956f30c11dca1/plutus-ledger/src/Ledger/TimeSlot.hs#L54-L56

Luckily the addition of `SlotConfig` to the existing API is pretty natural and matches previous works.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-chain-index/src/Plutus/ChainIndex/Config.hs
+++ b/plutus-chain-index/src/Plutus/ChainIndex/Config.hs
@@ -58,7 +58,7 @@ defaultConfig = ChainIndexConfig
   , cicSecurityParam = 2160
   , cicSlotConfig =
       SlotConfig
-        { scSlotZeroTime = 1591566291000
+        { scSlotZeroTime = 1596059091000
         , scSlotLength   = 1000
         }
   }
--- a/plutus-contract/src/Wallet/Emulator/Chain.hs
+++ b/plutus-contract/src/Wallet/Emulator/Chain.hs
@@ -95,15 +95,15 @@ getCurrentSlot = send GetCurrentSlot
 
 type ChainEffs = '[State ChainState, LogMsg ChainEvent]
 
-handleControlChain :: Members ChainEffs effs => ChainControlEffect ~> Eff effs
-handleControlChain = \case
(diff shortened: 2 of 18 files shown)


sjoerdvisscher:
@kk-hainq This fix looks good! Is it ready?

sjoerdvisscher (approved):

kk-hainq:
@sjoerdvisscher It is, thanks! The recent force pushes were only rebase(s) to make sure it wouldn't break the recent major changes.

sjoerdvisscher:
Thanks @kk-hainq!

PR #102, opened 2021-11-11
Plutus-use-cases QuickCheck models

MaximilianAlgehed:
This PR introduces some more `ContractModel` models for example contracts
and improves the error whitelist checking interface.

Pre-submit checklist:
- Branch
    - [X] Tests are provided (if possible)
    - [X] Relevant tickets are mentioned in commit messages
    - [X] Formatting, materialized Nix files, PNG optimization, etc. are updated
- PR
    - [X] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-contract/src/Plutus/Contract/Test/ContractModel.hs
+++ b/plutus-contract/src/Plutus/Contract/Test/ContractModel.hs
@@ -116,8 +116,10 @@ module Plutus.Contract.Test.ContractModel
     , NoLockedFundsProof(..)
     , checkNoLockedFundsProof
     -- $checkNoPartiality
-    , Whitelist(..)
+    , Whitelist
     , whitelistOk
+    , mkWhitelist
+    , errorPrefixes
     , defaultWhitelist
     , checkErrorWhitelist
     , checkErrorWhitelistWithOptions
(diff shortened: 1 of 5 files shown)


sjoerdvisscher:
By the looks of it these models are not yet used in any tests? Is that intentional?

MaximilianAlgehed:
Yes, that is intentional because the emulator is so slow that running a bunch of tests unfortunately gums up the works and makes everything slow.

sjoerdvisscher:
I just tested it:

```haskell
    testProperty "QuickCheck property" $
        withMaxSuccess 10 prop_Crowdfunding
```

and similar for `prop_Escrow` both take less than a second. 10 is maybe not enough to catch all potential bugs, but it will work as a sanity check in case we change the contract. So could you please add these?

MaximilianAlgehed:
Sure, but the problem with that is that `withMaxSuccess 10` will limit you to a few very small test cases that don't really mean anything. But in the spirit of getting something in there we might as well.

MaximilianAlgehed:
I don't understand the failures in `plutus-contract-test` - as far as I can tell there is nothing in this PR that should affect them.

MaximilianAlgehed:
Ok, they must have failed in one direction earlier and are now no longer failing - but the updated files are commited. That's odd and not what I expected but oh well.

sjoerdvisscher (approved):

MaximilianAlgehed:
@sjoerdvisscher I think this is ready to be merged :)

PR #103, opened 2021-11-11
Fix minting policy in tutorial and for state machine thread tokens

DavidEichmann:
Rather than open an issue I've started a PR with my proposed solution (untested). This grew out of a question I posed on slack as follows.

=== Question Start ===
the tutorial on minting has an example of a "one at a time" minting policy from some statically known token name `tname`:

```haskell
oneAtATimePolicy :: () -> ScriptContext -> Bool
oneAtATimePolicy _ ctx =
    -- 'ownCurrencySymbol' lets us get our own hash (= currency symbol)
    -- from the context
    let ownSymbol = ownCurrencySymbol ctx
        txinfo = scriptContextTxInfo ctx
        minted = txInfoMint txinfo
    -- Here we're looking at some specific token name, which we
    -- will assume we've got from elsewhere for now.
    in valueOf minted ownSymbol tname == 1
```

This seems odd to me! Doesn't this allow me to mint/burn any number of other tokens as long as they use a token name other than tname and as long as I mint a single token tname ? Surely we really want something like this:

```haskell
oneAtATimePolicy :: () -> ScriptContext -> Bool
oneAtATimePolicy _ ctx =
    ...
    in (getValue minted) Map.! ownSymbol == Map.singleton tname 1
```

Am I understanding correctly?
=== Question End ===

@brunjlar @michaelpj and Jean-Frederic Etienne agreed that this was a bug.

This has been moved from https://github.com/input-output-hk/plutus/pull/4105.

### DIFF
--- a/doc/plutus/tutorials/BasicPolicies.hs
+++ b/doc/plutus/tutorials/BasicPolicies.hs
@@ -30,7 +30,14 @@ oneAtATimePolicy _ ctx =
         minted = txInfoMint txinfo
     -- Here we're looking at some specific token name, which we
     -- will assume we've got from elsewhere for now.
-    in valueOf minted ownSymbol tname == 1
+    in case Map.lookup ownSymbol (getValue minted) of
+        Nothing -> n == 0
+        Just tokens -> all
+            (\(tname', n') -> if tname' == tname
+                then n' == 1
+                else n' == 0
+            )
+            (Map.toList tokens)
 
 -- We can use 'compile' to turn a minting policy into a compiled Plutus Core program,
 -- just as for validator scripts. We also provide a 'wrapMintingPolicy' function
--- a/plutus-contract/src/Plutus/Contract/StateMachine/ThreadToken.hs
+++ b/plutus-contract/src/Plutus/Contract/StateMachine/ThreadToken.hs
@@ -81,6 +81,31 @@ checkThreadTokenInner :: CurrencySymbol -> ValidatorHash -> Value -> Integer -> 
 checkThreadTokenInner currency (ValidatorHash vHash) vl i =
     Value.valueOf vl currency (TokenName vHash) == i
 
+-- | Check exactly `n` thread tokens and no other tokens with the given
+-- @CurrencySymbol@ are in the given @Value@.
+{-# INLINABLE checkThreadTokenInner #-}
+checkThreadTokenInner ::
+    -- | The currency symbol of the thread token.
+    CurrencySymbol ->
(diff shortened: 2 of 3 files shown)


plutus-contract/src/Plutus/Contract/StateMachine/ThreadToken.hs
```diff
@@ -81,6 +81,31 @@ checkThreadTokenInner :: CurrencySymbol -> ValidatorHash -> Value -> Integer ->
 checkThreadTokenInner currency (ValidatorHash vHash) vl i =
     Value.valueOf vl currency (TokenName vHash) == i
 
+-- | Check exactly `n` thread tokens and no other tokens with the given
+-- @CurrencySymbol@ are in the given @Value@.
```
DavidEichmann, line 85: @michaelpj You [suggested in the previous PR](https://github.com/input-output-hk/plutus/pull/4105#discussion_r729679862) that this might be too strict. Are you saying that the change is not necessary? My intention with this PR was specifically to disallow minting other tokens of the same CurrencySymbol. My reasoning was that the intention of the ThreadToken is for there to only be one unique token. The CurrencySymbol incorporates a `TxOutRef` used to anchor the NFT. In my mind we shouldn't allow creation of other TokenNames not intended by the StateMachine implementation.

doc/plutus/tutorials/BasicPolicies.hs
```diff
@@ -30,7 +30,14 @@ oneAtATimePolicy _ ctx =
...
     -- will assume we've got from elsewhere for now.
-    in valueOf minted ownSymbol tname == 1
+    in case Map.lookup ownSymbol (getValue minted) of
+        Nothing -> n == 0
+        Just tokens -> all
+            (\(tname', n') -> if tname' == tname
+                then n' == 1
+                else n' == 0
+            )
+            (Map.toList tokens)
```
DavidEichmann, line 40: Paerhaps I should take this suggestion https://github.com/input-output-hk/plutus/pull/4105#discussion_r729677649

DavidEichmann, line 40: If we merge https://github.com/input-output-hk/plutus/pull/4255 then we can replace this whole case 
```suggestion
in currencyValueOf ownSymbol minted == singleton ownSymbol tname 1
```

silky:
@DavidEichmann I'm closing this as stale; but feel free to re-open if you think it's relevant!

DavidEichmann:
Moved: https://github.com/input-output-hk/plutus-apps/pull/379

PR #126, opened 2021-11-19
SCP-2622: Document deployment details.

shlevy:


.github/workflows/deploy.yml
```diff
@@ -4,7 +4,7 @@ on:
     branches:
       - main # staging
     tags:
-      - plutus-pab/v[0-9]+.[0-9]+.[0-9]+ # production
+      - v[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9] # production
```
shlevy, line 7: Should last us a few thousand years

PR #154, opened 2021-11-30
Add mustPayWithDatumToPubKey (fix #146).

### LINKED ISSUE #146: Add Constraint to attach a Datum to a PubKey Output
## Area

[] *Plutus Foundation* Related to the GHC plugin, Haskell-to-Plutus compiler, on-chain code
[x] *Plutus Application Framework* Related to the Plutus application backend (PAB), emulator, Plutus libraries
[] *Marlowe* Related to Marlowe
[] *Other* Any other topic (Playgrounds, etc.)

## Describe the feature you'd like

The Cardano Ledger Specification allows attaching a Datum to PubKey outputs, not just to script outputs.
It is also possible to do this with the `Cardano API`, for example.
However, it is not possible using the Plutus `Contract`-monad, because there is no suitable constraint in `Legder.Constraints`.

I suggest changing the signature of `nustPayToPubKey` to

`mustPayToPubKey :: forall i o. PubKeyHash -> Maybe Datum -> Value -> TxConstraints i o`

to allow the (optional) attachment of a Datum to a public key output.

## Describe alternatives you've considered

Using a script that unlocks for a specific public key, one can replace public key outputs by outputs to this script, but this is obviously only a crude workaround.

ak3n:
Should fix #146.

---

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
+++ b/plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
@@ -48,6 +48,7 @@ module Ledger.Constraints.OffChain(
     ) where
 
 import Control.Lens (At (at), iforM_, makeLensesFor, over, use, view, (%=), (.=), (<>=))
+import Control.Monad (forM_)
 import Control.Monad.Except (MonadError (catchError, throwError), runExcept, unless)
 import Control.Monad.Reader (MonadReader (ask), ReaderT (runReaderT), asks)
 import Control.Monad.State (MonadState (get, put), execStateT, gets)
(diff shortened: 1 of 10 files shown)


plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
```diff
@@ -594,8 +594,9 @@ processConstraint = \case
         unbalancedTx . tx . Tx.mintScripts %= Set.insert mintingPolicyScript
         unbalancedTx . tx . Tx.mint <>= value i
         mintRedeemers . at mpsHash .= Just red
-    MustPayToPubKey pk vl -> do
-        unbalancedTx . tx . Tx.outputs %= (Tx.TxOut{txOutAddress=pubKeyHashAddress pk,txOutValue=vl,txOutDatumHash=Nothing} :)
+    MustPayToPubKey pk mdv vl -> do
+        let hash = datumHash <$> mdv
+        unbalancedTx . tx . Tx.outputs %= (Tx.TxOut{txOutAddress=pubKeyHashAddress pk,txOutValue=vl,txOutDatumHash=hash} :)
```
sjoerdvisscher, line 599: You should add the datum to `datumWitnesses` too, just like `MustPayToOtherScript` does.

plutus-ledger-constraints/src/Ledger/Constraints/OnChain.hs
```diff
@@ -85,7 +85,7 @@ checkTxConstraint ctx@ScriptContext{scriptContextTxInfo} = \case
     MustMintValue mps _ tn v ->
         traceIfFalse "L9" -- "Value minted not OK"
         $ Value.valueOf (txInfoMint scriptContextTxInfo) (Value.mpsSymbol mps) tn == v
-    MustPayToPubKey pk vl ->
+    MustPayToPubKey pk _ vl ->
```
sjoerdvisscher, line 88: I think this should do a datum check too, just like `MustPayToOtherScript`? Probably you should make `checkOutput` from there usable by both.

ak3n: Hmmm, I guess so. I've just wrote a contract test to ensure that datum is presented in `TxOut` and it seems to work. How do I test  `datumWitnesses` and `checkOutput` behaviour?

sjoerdvisscher: You only check for the hash, the actual datum probably isn't available. You can check that with `findDatum`.

`checkOutput` behaviour is a bit harder to test since the constraint mechanism should make sure a transaction satisfies it. I guess you should at least make sure the OnChain code is called by some test.

ak3n: I have tried to test the datum's presence but something isn't working. Test passes with any validator results. I guess I'm doing something wrong. But I have fixed the comments.

sjoerdvisscher (approved):
Looks good!

PR #156, opened 2021-12-01
Change path for chain-index.db

ganeshnithyanandam:
Currently, chain-index db is saved under /tmp. Since /tmp is periodically cleaned up by OS, chain-index will then start syncing again from genesis. 

This commit updates 'cicDbPath' to relocate 'chain-index.db' to "$PLUTUS/plutus-pab/test-node/testnet" directory, similar to that of db for cardano-node.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


silky:
Thanks @ganeshnithyanandam :)

I think we'll leave it as-is, given that you can just use your own config to get the behaviour you want.

PR #165, opened 2021-12-03
Update pab-config.yml to work with latest wallet config

Yasuke:
The current pab-config.yml uses an outdated wallet config that will throw an error, this can be especially confusing to users trying to do the integration test example. This PR will update the config to be compliant with the current wallet configuration.

### DIFF
--- a/plutus-pab/test-node/testnet/pab-config.yml
+++ b/plutus-pab/test-node/testnet/pab-config.yml
@@ -12,9 +12,9 @@ pabWebserverConfig:
   endpointTimeout: 5
 
 walletServerConfig:
-  baseUrl: http://localhost:8090
-  wallet:
-    getWallet: 1
+  tag: LocalWalletConfig
+  walletSettings:
+    baseUrl: http://localhost:8090
 
 nodeServerConfig:
   mscSocketPath: testnet/node.sock


plutus-pab/test-node/testnet/pab-config.yml
```diff
@@ -12,9 +12,9 @@ pabWebserverConfig:
   endpointTimeout: 5
 
 walletServerConfig:
-  baseUrl: http://localhost:8090
-  wallet:
-    getWallet: 1
+  tag: LocalWalletConfig
+  walletSettings:
+    baseUrl: http://localhost:9081
```
silky, line 17: just to verify; was this port change right? it's different on the left, so i imagine if something else was relying on it, that thing would also have to change?

silky (changes requested):
Looks good! Just one clarification re: the port change @Yasuke

plutus-pab/test-node/testnet/pab-config.yml
Yasuke: I am not really sure what the overarching changes may be, but what I do know is that the: defaultWalletConfig in Cardano.Wallet.Types uses 9081 as the port and that the rest of the pab configs point to that port.

Though the tx-inject config.yaml uses the older `getWallet: 1` and not t`ag: LocalWalletConfig`, but I am not certain if that should change as well.

bhoudebert: If I understood correctly, current cardano-wallet is still on 8090 even if you mean it will change (or has change but not in mainstream) it does not seem right to change it before here.

Am I missing something?

Yasuke: Well testing the above with a fresh clone of the repo only changing this config implies that the wallet is running at 9081.

The wallet config used by the CliSpec tests also uses the url which has  wallet on port 9081. 

But I may be the one missing something!

bhoudebert: Ok let me rephrase (to be sure I explained it right) in https://github.com/input-output-hk/cardano-wallet there is no mention at all of 9081, everything is 8090. So why should we have this number updated? That whats seems odd to me.

Not a big deal but it would mean that user following the doc would have to force their wallet server to another port (may not be possible outside docker which would be painful for me for example)

Yasuke: Thank you for rephrasing, what I was saying is I don't think you would have to change anything to follow the doc, I only changed this config and did not point my wallet anywhere else on a fresh clone of plutus-apps. 

What I must be missing is that this change doesn't effect the integration test but would effect other parts of the toolchain.

So what I am wondering now is why, without changing the wallet port , this config still worked for the entirety of the integration test [here](https://github.com/input-output-hk/plutus-apps/tree/main/plutus-pab/test-node), which I will investigate. There is a chance the port change was ignored and so I will try this again without the baseUrl change and I should have a more clear understanding of the machinations at play, and will update my PR accordingly.

bhoudebert: hmm could the integration test simply use the integrated mock wallet? If so the config would not matter at all (cannot dive right now but I could have a look later).

Yasuke: The config does matter, but it looks only the tag part is necessary. So when I am sure of that I will adjust things accordingly.

Yasuke:
I have updated this PR to not include the port change after confirming it was wrong.

silky (approved):
Thanks @Yasuke ! :)

PR #166, opened 2021-12-03
Fix mustSatisfyAnyOf logic (fix #101).

### LINKED ISSUE #101: `mustSatisfyAnyOf` fails the expected logic
## Summary

The function is admittedly completely undocumented, but from its type signature

    mustSatisfyAnyOf :: forall i o. [TxConstraints i o] -> TxConstraints i o

and the name, I would expect that it's meant to be equivalent to `any`, meaning if any of the items in the argument list is satisfied, the result will be satisfied as well. In a bit of pseudo-notation, 

    satisfied (mustSatisfyAnyOf xs) == any satisfied xs

As a reminder, the `TxConstraints` type is basically a conjunction of a list of `TxConstraint`, so it acts as an `all`; its `mempty` implementation is an empty list and is always trivially satisfied. So we can expand this to

    satisfied (mustSatisfyAnyOf xs) == any satisfied xs == any (all satisfied) xs

 That's not what happens. The actual function implementation is

    mustSatisfyAnyOf = singleton . MustSatisfyAnyOf . concatMap txConstraints

 The first thing the function does is to take a list of the conjunction lists flatten the list of `[[TxConstraint]]`, so we end up with

    satisfied (mustSatisfyAnyOf xs) == any (any satisfied) xs

## Steps to reproduce

    mustSatisfyAnyOf [mempty]

always fails.

## Expected behavior

    mustSatisfyAnyOf [mempty]

should be equal to `mempty`, i.e. it should be trivial to satisfy.

## System info (please complete the following information):

- OS: Ubuntu
- Version 21.10
- 58c093a49eb7a369865e361179d649264fc817a4

ak3n:
should fix #101

Pre-submit checklist:
- Branch
    - [x]  Relevant tickets are mentioned in commit messages
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-ledger-constraints/src/Ledger/Constraints/OnChain.hs
+++ b/plutus-ledger-constraints/src/Ledger/Constraints/OnChain.hs
@@ -113,7 +113,7 @@ checkTxConstraint ctx@ScriptContext{scriptContextTxInfo} = \case
         $ V.findDatum dvh scriptContextTxInfo == Just dv
     MustSatisfyAnyOf xs ->
         traceIfFalse "Ld" -- "MustSatisfyAnyOf"
-        $ any (checkTxConstraint ctx) xs
+        $ any (all (checkTxConstraint ctx)) xs
 
 {-# INLINABLE checkScriptContext #-}
 -- | Does the 'ScriptContext' satisfy the constraints?
--- a/plutus-contract/test/Spec/Contract.hs
+++ b/plutus-contract/test/Spec/Contract.hs
@@ -300,6 +300,13 @@ tests =
                 void $ Trace.assert "Always fails" $ const False
                 void $ Trace.waitNSlots 10
           in checkEmulatorFails "assert throws error" (defaultCheckOptions & minLogLevel .~ Debug) (waitingForSlot theContract tag 10) emTrace
+
+        , let c :: Contract () Schema ContractError () = do
+                let payment = Constraints.mustSatisfyAnyOf [mempty]
+                void $ submitTx payment
+          in run "mustSatisfyAnyOf [mempty] works"
+            ( assertDone c tag (const True) "should be done"
+            ) $ (void $ activateContract w1 c tag)
         ]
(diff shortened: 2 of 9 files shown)


plutus-ledger-constraints/src/Ledger/Constraints/OnChain.hs
```diff
@@ -107,7 +108,7 @@ checkTxConstraint ctx@ScriptContext{scriptContextTxInfo} = \case
         $ V.findDatum dvh scriptContextTxInfo == Just dv
     MustSatisfyAnyOf xs ->
         traceIfFalse "Ld" -- "MustSatisfyAnyOf"
-        $ any (checkTxConstraint ctx) xs
+        $ or $ map (all (checkTxConstraint ctx)) xs
```
sjoerdvisscher, line 111: That's just `any (all (checkTxConstraint ctx)) xs` isn't it?

sjoerdvisscher (approved):
👍 Would be nice to use `any` in the implementation of `MustSatisfyAnyOf` though 😄

plutus-ledger-constraints/src/Ledger/Constraints/OnChain.hs
ak3n: Ahah, yeah, missed that, thanks!

silky:
Seems fine @ak3n ; but I note that the PIR gets larger, and some of the golden outputs (amounts) change? Is that _because_ the PIR got larger? Is that expected?

ak3n:
@silky I think they get changed each time in `renderGuess.txt`.

At least it was true for https://github.com/input-output-hk/plutus-apps/commit/63123ffb527a37ba357e4cb38690625f945710b9, and for others commits in the history of that file.

silky (approved):

silky:
@ak3n I'm happy then if you're happy :)

PR #167, opened 2021-12-03
Fix missing datums for toCardanoTxBody

### LINKED ISSUE #139: Strange behavior of toCardanoTxBody (missing datums)
## Summary

I try to test my contracts with functions `toCardanoTxBody` and `evaulateTransactionExecutionUnits`.
I build Plutus `Tx` by hand and convert it to cardano `Tx` to use evaluation function. 

Something strange happens when I update datum from the input to output on the same address. 
It can not find datum by datum hash. The function that I use to create part of Tx that pays to script:

```haskell
-- | Pay to the script.
payToScript :: ToData (DatumType a) => TypedValidator a -> DatumType a -> Value -> Tx
payToScript tv dat val =
  mempty
    { txOutputs = [TxOut (validatorAddress tv) val (Just dh)]
    , txData = M.singleton dh datum
    }
  where
    dh = datumHash datum
    datum = Datum $ toBuiltinData dat
```

So it seems that I fill up `txData` and set up datum hash properly.
But when I use it in the code datum can only be found if it is the same as used in the inputs. If I change the datum validator can not find it in the outputs of TxInfo by hash although datum hash is present.
I tried to lookup how the function `toCardanoTxBody` works and I can see that it does not use `txData`.
Or maybe I miss something? 
Please can you explain how `txData` should be used in the Cardano `TxBody` properly so that I can continue with my testing?
or how it should be used in `toCardanoTxBody`?

As I understand Cardano Tx does not contain datum's and only hashes are used.
How can the datums be visible to `evaluateTransactionExecutionUnits`?

I've tried to include also output datums in the `UTxO` argument of `evaluateTransactionExecutionUnits` and it also can not find them :(

anton-k:
Fixes #139

### DIFF
--- a/plutus-ledger/src/Ledger/Tx/CardanoAPI.hs
+++ b/plutus-ledger/src/Ledger/Tx/CardanoAPI.hs
@@ -59,6 +59,7 @@ import Cardano.Api.Byron qualified as C
 import Cardano.Api.Shelley qualified as C
 import Cardano.BM.Data.Tracer (ToObject)
 import Cardano.Chain.Common (addrToBase58)
+import Cardano.Ledger.Alonzo.Data qualified as Alonzo
 import Cardano.Ledger.Alonzo.Scripts qualified as Alonzo
 import Cardano.Ledger.Alonzo.TxWitness qualified as C
 import Cardano.Ledger.Core qualified as Ledger
@@ -303,6 +304,7 @@ fromAlonzoLedgerScript (Alonzo.PlutusScript _ bs) =
              $ SBS.fromShort bs
    in either (const Nothing) Just script
 
+
 toCardanoTxBody ::
     [Api.PubKeyHash] -- ^ Required signers of the transaction
     -> Maybe C.ProtocolParameters -- ^ Protocol parameters to use. Building Plutus transactions will fail if this is 'Nothing'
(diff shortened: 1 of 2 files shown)


sjoerdvisscher:
@anton-k The change has already been applied after we copied it! https://github.com/input-output-hk/cardano-node/blob/master/cardano-api/src/Cardano/Api/TxBody.hs#L2600

It looks like your `nub` is not necessary.

plutus-ledger/src/Ledger/Tx/CardanoAPI.hs
```diff
@@ -419,11 +421,19 @@ fromCardanoTxOut (C.TxOut addr value datumHash) =
...
             <*> toCardanoTxOutValue value
-            <*> toCardanoTxOutDatumHash datumHash
+            <*> cardanoDatumHash
+  where
+    cardanoDatumHash =
+      case flip Map.lookup datums =<< datumHash of
+        Just datum -> pure $ C.TxOutDatum C.ScriptDataInAlonzoEra (toScriptData $ P.getDatum datum)
+        Nothing    -> toCardanoTxOutDatumHash datumHash
+
+toScriptData :: P.BuiltinData -> C.ScriptData
```
sjoerdvisscher, line 435: Isn't this the same as `toCardanoScriptData`?

anton-k:
> @anton-k The change has already been applied after we copied it! https://github.com/input-output-hk/cardano-node/blob/master/cardano-api/src/Cardano/Api/TxBody.hs#L2600
> 
> It looks like your `nub` is not necessary.

We need nub in case if we have several inputs with the same script, otherwise witness is going to be duplicated.
Also outputs can contain several datums which are the same.

plutus-ledger/src/Ledger/Tx/CardanoAPI.hs
anton-k: Indeed, I've fixed that

sjoerdvisscher:
Thanks!

anton-k:
Thanks for merging! 
It will be also cool to add `nub` cardano-api version of it. In our project we had to create our own `makeTransactionBody` just for that reason

sjoerdvisscher:
@anton-k I agree, could you open an issue (or PR) against `cardano-node`?

PR #178, opened 2021-12-08
Write arbitrary instance for data that uses sized

### LINKED ISSUE #172: `Arbitrary Datum` produces infinitely* large data structures
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

`arbitrary` method of `Arbitrary Datum` instance generates infinitely* large `Datum`s with very high probability (more than `0.3` i guess), that require corresponding amount of memory.

(*) - possibly, didn't have infinite amount of time to ensure

1. Run following code:
```haskell
import Test.QuickCheck qualified as QC
import Ledger qualified
import Plutus.PAB.Arbitrary ()

main = print =<< QC.generate (QC.arbitrary :: QC.Gen Ledger.Datum)
```

### Actual Result

Infinitely large Datum printing to the console.

### Expected Result

Some finite, likely small Datum data structure prints out.

### What have you tried to resolve this issue?

After my property-based tests were hang because of `arbitrary @Datum` call, I tried out to look what it generates by printing it out. Printing process took more than 10 minutes and didn't stop.

### Checked our docs?

- [X] Yes, I have looked in the the readme, plutus docs, and technical report for help on this issue.

### Checked Stack Exchange?

- [X] Yes, I have searched Stack Exchange for this issue and it doesn't exist.
- [X] I have considered raising this issue in Stack Exchange.

### System info

OS: Ubuntu 
Version: 20.04
Plutus: 3f089ccf0ca746b399c99afe51e063b0640af547
plutus-apps: 404af7ac3e27ebcb218c05f79d9a70ca966407c9

### Screenshots and attachments

_No response_

Yasuke:
By using sized we can have an arbitrary instance that actually
terminates and is configurable via standard QuickCheck
methods (e.g. generate, resize). This will allow testing with Datums and
will prevent users unintentionally stalling their tests when generating
Datums, BuiltinData, and Data. Fixes #172.

### DIFF
--- a/plutus-pab/src/Plutus/PAB/Arbitrary.hs
+++ b/plutus-pab/src/Plutus/PAB/Arbitrary.hs
@@ -8,6 +8,7 @@
 -- across to the test suite.
 module Plutus.PAB.Arbitrary where
 
+import Control.Monad (replicateM)
 import Data.Aeson (Value)
 import Data.Aeson qualified as Aeson
 import Data.ByteString (ByteString)
@@ -29,7 +30,7 @@ import Plutus.Contract.StateMachine (ThreadToken)
 import PlutusTx qualified
 import PlutusTx.AssocMap qualified as AssocMap
 import PlutusTx.Prelude qualified as PlutusTx
-import Test.QuickCheck (Gen, oneof)
+import Test.QuickCheck (Gen, Positive (..), oneof, sized)
 import Test.QuickCheck.Arbitrary.Generic (Arbitrary, arbitrary, genericArbitrary, genericShrink, shrink)
 import Test.QuickCheck.Instances ()
 import Wallet (WalletAPIError)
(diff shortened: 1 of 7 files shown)


plutus-pab/src/Plutus/PAB/Arbitrary.hs
```diff
@@ -156,8 +157,53 @@ instance Arbitrary ThreadToken where
     shrink = genericShrink
 
 instance Arbitrary PlutusTx.Data where
-    arbitrary = genericArbitrary
-    shrink = genericShrink
```
sjoerdvisscher, line 160: `genericShrink` is actually ok, it's from `QuickCheck` not `generic-arbitrary`.

plutus-pab/src/Plutus/PAB/Arbitrary.hs
```diff
@@ -156,8 +157,53 @@ instance Arbitrary ThreadToken where
...
+
+        arbitraryMap 0 = pure $ PlutusTx.Map []
+        arbitraryMap n = do
+          (n', m) <- segmentRange n
+          kvs <- replicateM m (arbitraryPair n')
+          pure $ PlutusTx.Map kvs
+
+        arbitraryPair n = do
+          (n', _) <- segmentRange n
+          (,) <$> arbitraryData n' <*> arbitraryData n'
```
sjoerdvisscher, line 188:
```suggestion
          (,) <$> arbitraryData n' <*> arbitraryData (n - n')
```
perhaps?

sjoerdvisscher (approved):

silky:
@Yasuke Looks good; I think it would also make sense to have a few tests that confirm this is working how you'd expect?

plutus-pab/src/Plutus/PAB/Arbitrary.hs
kozross: I definitely recommend using `div` for size reductions.

plutus-pab/src/Plutus/PAB/Arbitrary.hs
Yasuke: I added back the genericShrink, thanks for the heads up!

plutus-pab/src/Plutus/PAB/Arbitrary.hs
Yasuke: I ended up using div in this way, thanks @MaximilianAlgehed and @kozross

Yasuke:
Test-suite added and comments addressed.

silky (approved):
Very nice! :) Thanks @Yasuke

PR #179, opened 2021-12-08
add test check utxo function

kk-anlm:
Add a function to check if datum and value matching while testing. Function `dataAtValue` only check with a list of datums, while we want to check if a datum is matching with the right value in testing. For example:
`checkOutputsWithDatumAtAddress @OracleDatum (oracleAddress oracleParams) (== [(Price AssocMap.empty, unitValue oracleToken)])`

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-contract/src/Plutus/Contract/Test.hs
+++ b/plutus-contract/src/Plutus/Contract/Test.hs
@@ -51,6 +51,7 @@ module Plutus.Contract.Test(
     , waitingForSlot
     , valueAtAddress
     , dataAtAddress
+    , outputsAtAddress
     , reasonable
     , reasonable'
     -- * Checking predicates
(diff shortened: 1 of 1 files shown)


kk-hainq (changes requested):
Nice utility but still need better code and documentation. Both `plutus-contract` and `plutus-use-cases` import and use `Plutus.Contract.Test` a lot. This PR would be way more convincing if you could use these functions somewhere there to show its usage, and that it is an upgrade from an old way of writing things, etc. Those diffs would be very helpful.

silky:
@kk-anlm - Thanks for the work so far on this; are you planning to continue on? Otherwise I'll close due to inactivity.

kk-anlm:
> @kk-anlm - Thanks for the work so far on this; are you planning to continue on? Otherwise I'll close due to inactivity.

I'm still working on it. Hope to have comments to finish.

kk-anlm:
We don't follow this direction anymore.

PR #181, opened 2021-12-09
SCP-3130: Tx constraint to specify stake stake key hash with payment

koslambrou:
Modified the tx creation constraints (`plutus-ledger-constraints` repo) to allow to specify the stake public key hash of a transaction output address.

This isn't strictly necessary, since you can make payments using only the payment key. However, it is useful when you want to derive a specific Cardano address (bech32 format) which needs the payment key + stake key.

While doing that, I also took the time to add a additionnal newtype wrappers to `PubKeyHash` such as `PaymentPubKeyHash`, `PaymentPubKey`, `StakePubKeyHash` and `StakePubKey` in order to clearly understand to which `PubKeyHash` we are refering to. This implied a lot of code change. These datatypes where defined in `Ledger.Address`.

Related to that, I changed the name of `WalletEffect.ownPubKeyHash` to `WalletEffect.ownPaymentPubKeyHash`. Maybe it also makes sense to also add the effect `WalletEffect.ownStakePubKeyHash`, but this should be in a separate PR.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-ledger/src/Ledger/Address.hs
```diff
@@ -1,20 +1,87 @@
...
+
+{-# INLINABLE pubKeyHashAddress #-}
+-- | The address that should be targeted by a transaction output locked by the
+-- given public payment key (with it's public stake key).
+--
+-- TODO: This should be moved to Plutus.V1(or V2).Ledger.Address with the newtypes.
+pubKeyHashAddress :: PaymentPubKeyHash -> Maybe StakePubKeyHash -> Address
+pubKeyHashAddress (PaymentPubKeyHash pkh) skh =
+    Address (PubKeyCredential pkh)
+            (fmap (StakingHash . PubKeyCredential . unStakePubKeyHash) skh)
```
koslambrou, line 75: @michaelpj This was copied from `plutus` because it's own `pubKeyHashAddress` would set the stake credential to `Nothing`. Is it okay if modify that function on the `plutus` side? V1 or V2?

michaelpj: Yeah, fine to modify it, it's not used by the key bit.

silky (approved):
Looks good! I think it's a net win for readability as well :)

PR #191, opened 2021-12-13
Add the rollback history command line option.

raduom:
Provide a `--rollback-history INT` and a configuration setting for the PAB Webserver (named `rollbackHistory`) that controls how much history we retain for rollbacks.

If you don't need to test rollbacks, then setting the rollback-history to a lower value (I tested with `500`) will greatly limit the amount of RAM that the PAB will consume. A value of `0` turns off this feature.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


raduom:
I just thought of moving the option into a new `Development` section as this and the one in the next PR should probably not be used in production.

plutus-pab/src/Plutus/PAB/Core/ContractInstance/BlockchainEnv.hs
```diff
@@ -97,24 +96,19 @@ blockAndSlot BlockchainEnv{beCurrentBlock, beCurrentSlot} =
...
-      else pure ()
-  pure result
+  case event of
+    Resume _ -> Right <$> blockAndSlot blockchainEnv
+    RollForward (BlockInMode (C.Block header transactions) era) _ ->
+      case era of
+        -- Unfortunately, we need to pattern match again all eras because
+        -- 'processBlock' has the constraints 'C.IsCardanoEra era', but not
+        -- 'C.BlockInMode'.
+        C.ByronEraInCardanoMode   -> processBlock header blockchainEnv transactions era
```
sjoerdvisscher, line 106: I know this is older code, but isn't the simple solution here to let `processBlock` take the whole `BlockInMode` value?

plutus-chain-index-core/src/Plutus/ChainIndex/UtxoState.hs
```diff
@@ -104,9 +105,10 @@ trimIndex ::
   => Integer
   -> UtxoIndex a
   -> UtxoIndex a
+trimIndex 0          ix = ix
 trimIndex kParameter ix =
     let (lb, rb) = bounds ix
-    in  if (rb - lb) > kParameter * 2
+    in  if (rb - lb) > kParameter * 10
```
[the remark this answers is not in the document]
sjoerdvisscher: Let me understand this: the difference between `trimIndex` and `reduceBlockCount` is that `trimIndex` is not semantics-preserving, i.e. it throws away older utxos?

plutus-pab/src/Plutus/PAB/Core/ContractInstance/STM.hs
```diff
@@ -149,16 +149,17 @@ data OpenTxOutProducedRequest =
 --   may be interested in.
 data BlockchainEnv =
     BlockchainEnv
-        { beCurrentSlot  :: TVar Slot -- ^ Current slot
-        , beTxChanges    :: TVar (UtxoIndex TxIdState) -- ^ Map holding metadata which determines the status of transactions.
-        , beTxOutChanges :: TVar (UtxoIndex TxOutBalance) -- ^ Map holding metadata which determines the status of transaction outputs.
-        , beCurrentBlock :: TVar BlockNumber -- ^ Current block
+        { beRollbackHistory :: Integer -- ^ How much history do we retain in the environment. Zero signifies no trimming is done.
```
sjoerdvisscher, line 152: `Maybe Integer` instead of special casing `0` seems to me more the Haskelly thing to do here?

sjoerdvisscher (changes requested):

plutus-pab/src/Plutus/PAB/Core/ContractInstance/STM.hs
raduom: Yes. I was considering doing that.

plutus-chain-index-core/src/Plutus/ChainIndex/UtxoState.hs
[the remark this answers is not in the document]
raduom: Yes. But since I want it to be semantics preserving I switched to using `reduceBlockCount` instead.

plutus-pab/src/Plutus/PAB/Core/ContractInstance/BlockchainEnv.hs
raduom: I think this is written this way because you need a 'proof' that `C.IsCardanoEra era` (required by a constraint on `processBlock`) and that is provided by pattern matching on the era part. I may not understand what you are asking though.

silky (approved):

PR #210, opened 2021-12-18
Raduom/pab resume

raduom:
Adds a command-line option (`--resume-from`) and a configuration section which allows for the PAB to resume from a given block and slot id.

Notes: 
1. The resume will happen from the block immediately after the one specified on the command line (or the configuration).
2. This will be rolled back when we will implement persistence for our internal state, but provides a way to "synchronise" very fast in cases where the history of the blockchain is not needed for contract operation.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


silky (approved):

silky:
Nice; maybe could do with a hint about how to get the block information when you want to start from a certain point?

PR #213, opened 2021-12-21
Refine executable dependencies

kk-hainq:
Complicating our Cabal setup to build executable dependencies is not nice. This PR pulls `cardano-cli`, `cardano-node`, and `cardano-wallet` to the Nix shell with a few words here and there directing people to the original repositories to find their best installation option. When people do want to build the executables from the source, they should do it in their respective clones, not changing commit hashes and doing it here.

The net effects of this PR are:
- `cardano-cli`, `cardano-node`, `cardano-wallet` available in `nix-shell` by default.
- `cardano-node`, `cardano-cli`, `cardano-config`, and an `optparse-applicative` fork removed from our `cabal.project`.
- A thinner development environment (+95 -2194 yay!).

The second commit is only `updateMaterialized`, so reviewing the others should be sufficient.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-pab/test-node/testnet/pab-config.yml
```diff
@@ -46,6 +46,9 @@ signingProcessConfig:
 metadataServerConfig:
   mdBaseUrl: http://localhost:9085
 
+developmentOptions:
```
kk-hainq, line 49: This is a random bug fix I needed to test `plutus-pab/test-node`. Should be moved to a separate PR if this one isn't wanted.

raduom:
I tried to have nix provide the `cardano-node` and `cardano-cli` at some point. The problem was that inside the plutus `nix-shell` the version of the plutus (for example) dependency was not the same as the one from the `cardano-node` repository. So you get a pretty unpredictable build of `cardano-node` where some of its dependencies are actually not tested. This is why the current advice is to build or install the node separately.

I think that removing them from cabal is a very good idea, but adding them to nix is going to create problems for us down the road (which in fact, it already has).

kk-hainq:
@raduom I indeed built on your previous work and saw the removal commit! That is why I was still very vague with the `cardano-node` instructions. I don't know what the correct way is...

Overall, I don't think adding `cardano-node` to the Nix shell is the problem. People can choose not to use them, or easily build even less stable executables themselves. We should either document a stable way or just automate that process. I wonder if we could open an issue to track that, then merge this earlier? I'd be happy to remove `cardano-node` and `cardano-cli` from `nix-shell` for now if required.

For the issue, do you have notes on previously found problems? If `plutus-apps` should follow `cardano-node` then we should enforce that with tests and simply use prebuilt nodes?

raduom:
I've tested this PR and the node builds without linking to the `plutus-apps` dependencies.

plutus-pab/test-node/testnet/pab-config.yml
```diff
@@ -46,6 +46,9 @@ signingProcessConfig:
...
+  pabRollbackHistory: null
+
```
raduom, line 51: It should be fine. However we need to add some more things.
```suggestion
developmentOptions:
  pabRollbackHistory: null
  pabResumeFrom:
    { "tag" : "PointAtGenesis" }
    # { "tag" : "Point"
    # , "pointBlockId" : "613d23b8df670e3692f4e4ab59b179778519de15999ee1a51077fcc02a48abaa"
    # , "pointSlot" : { "getSlot" : 45656608 }
    # }

```

raduom (approved):
Nice work!

kk-hainq:
Thank you @raduom for your insights and help on Slack!

PR #216, opened 2021-12-23
Move psgenerator out of plutus-pab

### LINKED ISSUE #212: Remove mandatory playground and PureScript dependencies from PAB
### Describe the feature you'd like

Many (if not most?) PAB users do not care about the playground nor use PureScript in their projects (also see #19). However, the current setup forces them to build and track `playground-common`, the `purescript-bridge` & `servant-purescript` forks, and more related dependencies by default. This is unnecessarily resource-consuming and scary. It would be way better if `plutus-pab` doesn't depend on any playground and PureScript machinery by default.

From a quick experiment, it looks like `plutus-pab` only needs `FunctionSchema`, `FormSchema`, and perhaps `EndpointToSchema` from `playground-common`, which may be ported to `plutus-schema` or something? For PureScript we can just port `Plutus.PAB.Run.PSGenerator` to `plutus-purescript`? 

We would love to work on this if accepted.

### Describe alternatives you've considered

We can progress just fine with these dependencies in place. However, we do believe PAB would be a lot more user-friendly without them. This will also ease off dependency management a bit for dApp developers, which at the moment is quite stressful and unfriendly, see #180.

ak3n:
an attempt to fix #212

- moved all executables from `plutus-pab` to `plutus-pab-executables`
- removed psgenerator from `plutus-pab`, now lives in the `plutus-pab-psgenerator` sublibrary in `plutus-pab-executables`

turned out this is necessary for `playground-common` in #223 removal as well

---

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


silky (approved):
Nice! Thanks; I'm happy pending the discussed changes to bring some minimal tests back to the `plutus-pab` package :D

silky (approved):

raduom (approved):
LGTM

plutus-pab/src/Plutus/PAB/Run/Command.hs
```diff
@@ -49,10 +46,7 @@ allServices =
 
 
 data NoConfigCommand =
-    PSGenerator -- ^ Generate purescript bridge code
-          { psGenOutputDir :: !FilePath -- ^ Path to write generated code to
-          }
-    | WriteDefaultConfig -- ^ Write default logging configuration
+    WriteDefaultConfig -- ^ Write default logging configuration
```
koslambrou, line 49: Is this still useful in the `plutus-pab`? It already exists in plutus-pab-executables.

ak3n: Hmm. It depends on if we want that any PAB could generate a default config? Or only `plutus-pab-setup` binary?

koslambrou: Well, `WriteDefaultConfig` needs to be removed either from `plutus-pab` or `plutus-pab-executables`. There's some duplication.

ak3n: Will remove it from `plutus-pab` then. I think it will ideally fit `plutus-pab-setup` executable.

koslambrou (approved):

PR #219, opened 2021-12-28
Experiment: remove traces & inline trivial lambdas in usecase tests

kk-hainq:
This is just an experiment (not to be merged) on changes at the PIR level when we remove traces and inline trivial lambdas in the relevant usecase tests.

The overall diff is not interesting as it includes a huge diff from applying `plutus`'s master HEAD. The second commit portraits the actual experiment from applying https://github.com/HachiSecurity/plutus/commit/c8accb12d87de646544fd4200fa240835046a995.

Not much is inlined so improvements are very minimal, but it is also a data point showing little risk from inlining trivial lambdas (there are not that many in real scripts?).

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -241,7 +241,7 @@ source-repository-package
 source-repository-package
  type: git
  location: https://github.com/input-output-hk/plutus
- tag: 2721c59fd2302b75c4138456c29fd5b509e8340a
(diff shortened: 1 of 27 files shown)


kk-hainq:
https://github.com/input-output-hk/plutus/pull/4311 closes this now!

Issue #129, opened 2021-11-21
Improve Chain Index documentation


kk-hainq:
## Area
[x] *Plutus Application Framework* Related to the Plutus application backend (PAB), emulator, Plutus libraries

## Describe the feature you'd like

We should write documentation for chain index setup, API, use cases, customization, and more. 

A good start is to add `openapi3` support to the RESTful endpoints, similar to https://github.com/input-output-hk/plutus/pull/3807. We can then write more textual documentation, from high-level guidelines to the details like tuning `cicStoreFrom` from `2877844` on the testnet, and `6236060` on the mainnet for Alonzo-only transactions, etc.

## Describe alternatives you've considered

We mostly infer things from the source code at the moment, which is not always convenient for developers in general.

## Additional context / screenshots

This issue is part of the #4 mega-thread.

Yasuke:
@kk-hainq Thank you for the suggestion, we are making chain-index into a library and we will be documenting the features while doing so.

kk-hainq:
@Yasuke Awesome, thank you! Let us know if we could help in any way.

Yasuke:
Hey @kk-hainq It looks like #186 has laid the groundwork for separating chain-index out as a library, if you see anything obviously missing or aren't satisfied with the added documentation, we would definitely appreciate your input into the state of things.

kk-hainq:
Documentation is hard and dangerous (my left wrist still screams carpal tunnel after our last documentation sprint) so I think we should take steps. We can help add more haddock to `plutus-chain-index`, `plutus-chain-index-core` and decorate their README(s) more, what do you think?

Yasuke:
We would be happy to have your contribution @kk-hainq!

kk-hainq:
Hang tight my right hand is already typing!

kk-hainq:
This issue is outdated now. We'll open more documentation issues when it's more appropriate (architecture is finalized, code is refactored well enough, etc.).

PR #222, opened 2021-12-29
More chain index configs

### LINKED ISSUE #73: Chain Index: Allow more user configurations
## Area
[x] *Plutus Application Framework* Related to the Plutus application backend (PAB), emulator, Plutus libraries

## Describe the feature you'd like

We should extend the configuration file to let users decide what they want to track and store. For example, we only care for Alonzo's data for our current security work. The ability to store data by era would likely be helpful. For dApp work, we mainly care for transactions that interact with our protocol. The ability to not store irrelevant data would be excellent.

For example, a small addition to `AppendBlock` like in #72 allows applications to configure the choice to store a batch of transactions to the DB or not. Given that tip and UTXO processing is unaffected, most of these configurations are pretty safe to the functioning of the chain index?

## Describe alternatives you've considered

We have been customizing the chain index to our needs, but it is prone to upstream changes. I guess the more people can share and have in upstream, the better as well.

## Additional context / screenshots

This issue is part of the #4 mega-thread.

kk-hainq:
This should resolve #73.

Done:
- Add `cicStoreNonScriptTxs` to allow not storing transactions that do not reference a script.

TODO:
- Add `cicWhitelistedAddresses` to allow only storing transactions that interact with one of the pre-defined addresses.
- Write tests.
- Write more documentation. However, most documentation efforts should still be done separately for #129 once this one is merged.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-chain-index/src/Plutus/ChainIndex/App.hs
+++ b/plutus-chain-index/src/Plutus/ChainIndex/App.hs
@@ -26,8 +26,8 @@ import Cardano.BM.Configuration.Model qualified as CM
 import Plutus.ChainIndex.CommandLine (AppConfig (..), Command (..), applyOverrides, cmdWithHelpParser)
 import Plutus.ChainIndex.Compatibility (fromCardanoBlockNo)
 import Plutus.ChainIndex.Config qualified as Config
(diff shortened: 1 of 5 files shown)


kk-hainq:
We have now realized that this is not a good direction in the near future. One would need a very fast chain index first :(.

PR #223, opened 2021-12-31
Move playground schema to plutus-pab-schema

### LINKED ISSUE #212: Remove mandatory playground and PureScript dependencies from PAB

ak3n:
an attempt to fix #212 (second half)

TODO:

- allow to integrate schema's `apiHandler`
- client

---

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


koslambrou:
Is there something blocking this PR?

ak3n:
I got stuck on something related to Servant API and server type mismatch.

I'm not sure I like the solution that I got in this PR. It makes the PAB more modular but also more complex. Too modular to my taste. We don't have a need in such complexity yet.

It raises the question what's the point of `/schema` endpoint? I haven't found any usage in the PAB itself. I suspect that the only client of it might be Marlowe. Then maybe we should move that functionality there? I don't know if usual PAB users need contract schemas. Would love to hear your opinion.

kk-hainq:
> It raises the question what's the point of /schema endpoint? I haven't found any usage in the PAB itself. I suspect that the only client of it might be Marlowe. Then maybe we should move that functionality there? I don't know if usual PAB users need contract schemas. Would love to hear your opinion.

True...

@koslambrou @ak3n Can we nuke it here, and re-add to Marlowe later on if such a demand arises?

ak3n:
@kk-hainq we have discussed it internally and I will try again to take a look and if it won't be possible to re-organize the dependencies then we will remove the `/schema` endpoint completely.

PR #238, opened 2022-01-11
PureScript Updates

jhbertra:
Original:

This was an error made when writing the command for buildPursPackage. It should be reading from `$src/**/*.purs`, not `src/**/*.purs`, as this gets set when calling this function.

Now:

I've taken the liberty of updating several things regarding the PureScript:

1. I've fixed the problem mentioned above with `buildPursPackage`
2. I've un-gitignored the generated purescript code. This will make it so that Haskell changes that result in generated PureScript changes are always visible in pull requests.
3. I've also added a CI test to make sure that the generated purescript is up-to-date (a consequence of checking it in is that it needs to be updated when the original Haskell types change).
4. I've removed purs-loader. It offered no benefit and slowed down / complicated the webpack build significantly. Now it simply bundles the javascript emitted by `purs` in `output`.
5. I've improved a couple things in the nami-demo config.


sjoerdvisscher:
How does this suddenly generate this much errors? Do you need help fixing them?

jhbertra:
@sjoerdvisscher it's complicated, but I fixed it in `marlowe-cardano`, so I know what needs to be done.

koslambrou (approved):

sjoerdvisscher (approved):
Wow, I had no idea we generated such a large amount of code!

PR #243, opened 2022-01-12
Fix protocol version error messages happening during synchronisation.

raduom:
This fixes a couple of issues:

A. When starting all servers the mocked chain index and wallet are also started. However, they can only use the mocked versions of the client which requires a mocked server running. When these clients try to connect to the Alonzo node the versions don't match causing errors.

* Solution: I removed the mocked services from the list of commands to run if we are connecting to an alonzo node and added a list of services started to the command's output (to point out that we are not even trying to start those services in those circumstances).

B. When starting the PAB server the mocked chain is always used to synchronise the slot number, which causes the same connection errors when trying to connect to the alonzo node.

* Solution: I wrote a function that runs chain sync with the proper client according to the way the PAB is configured.
* Note: This *does not* support resuming right now. It's not clear what a solution to this problem is (and indeed, if it even is a problem). 

C. The `StartNode` command, when using the Alonzo node, exists after logging some information. The semantics of commands is that if any command exits, they all do, which caused the PAB to always exit early. This bug was not found, due to bug (E) but became apparent once I stoped running the chain index.

* Solution: Make the `StartNode` command wait indefinitely when using the Alonzo node.

D. The `TxSender` thread is always started even though it is only used when the mock node is running. Even though nobody sent any transactions if the real node was configured the thread would still try to connect to the node and fail due to wrong versioning (again, mocked vs real versioning missmatch).

* Solution: Make it optional and only start it when in mocked mode. Note that while the `TxSender` is optional in the code running the executable, it should not be in the mocked node server and wallet where it is required. This is why i had to inject some `Just txHandler`s here and there.

** At this point we are not getting any more connection errors, and everything runs smoothly. However... **

E. The PAB commands infrastructure uses a `MVar` based semaphore that ensures that two commands are not ran at the same time by using the `ccaAvailability` field which is filled when a command has finished processing and is cleared by the `ForkCommand` (thus waiting for completion of the previous command). However, the ChainIndex command never places that `MVar` back blocking the execution of the rest of the commands. As long as the ChainIndex is the last command ran this is not apparent (as it was supposed to block anyway), but when I skipped this command the PAB would exit early (due to (C)).

* Solution: Make ChainIndex set `ccaAvailability` when it finished initializing.

Some refactoring:
* Moved ChainSyncHandle from Cardano.Node.Client to Cardano.Node.Types

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-pab-executables/tx-inject/config.yaml
+++ b/plutus-pab-executables/tx-inject/config.yaml
@@ -13,18 +13,18 @@ walletServerConfig:
     baseUrl: http://localhost:9081
 
 nodeServerConfig:
-  mscBaseUrl: http://localhost:9082
-  mscSocketPath: /tmp/node-server.sock
-  mscKeptBlocks: 100000
-  mscNetworkId: "1097911063"
-  mscSlotConfig:
+  pscBaseUrl: http://localhost:9082
+  pscSocketPath: /tmp/node-server.sock
+  pscKeptBlocks: 100000
+  pscNetworkId: "1097911063"
+  pscSlotConfig:
     scSlotZeroTime: 1596059091000 # Wednesday, July 29, 2020 21:44:51 - shelley launch time in milliseconds
     scSlotLength: 1000 # In milliseconds
-  mscFeeConfig:
+  pscFeeConfig:
     fcConstantFee:
       getLovelace: 10 # Constant fee per transaction in lovelace
     fcScriptsFeeFactor: 1.0 # Factor by which to multiply size-dependent scripts fee in lovelace
-  mscInitialTxWallets:
+  pscInitialTxWallets:
     - getWallet: 1
     - getWallet: 2
     - getWallet: 3
@@ -35,7 +35,7 @@ nodeServerConfig:
     - getWallet: 8
     - getWallet: 9
     - getWallet: 10
-  mscNodeMode: AlonzoNode
+  pscNodeMode: AlonzoNode
 
 chainIndexConfig:
   ciBaseUrl: http://localhost:9083
--- a/plutus-pab-executables/tx-inject/config.yaml
+++ b/plutus-pab-executables/tx-inject/config.yaml
@@ -16,7 +16,8 @@ nodeServerConfig:
   pscBaseUrl: http://localhost:9082
   pscSocketPath: /tmp/node-server.sock
   pscKeptBlocks: 100000
-  pscNetworkId: "1097911063"
+  # pscNetworkId: "1097911063"
+  pscNetworkId: "3"
   pscSlotConfig:
     scSlotZeroTime: 1596059091000 # Wednesday, July 29, 2020 21:44:51 - shelley launch time in milliseconds
     scSlotLength: 1000 # In milliseconds
--- a/plutus-pab/src/Cardano/Node/Client.hs
+++ b/plutus-pab/src/Cardano/Node/Client.hs
@@ -38,17 +38,24 @@ handleNodeClientClient ::
     forall m effs.
     ( LastMember m effs
     , MonadIO m
-    , Member (Reader MockClient.TxSendHandle) effs
+    , Member (Reader (Maybe MockClient.TxSendHandle)) effs
     , Member (Reader ChainSyncHandle) effs
     )
     => SlotConfig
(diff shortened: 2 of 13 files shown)


plutus-pab/src/Cardano/Node/Client.hs
```diff
@@ -37,19 +38,44 @@ handleNodeClientClient ::
...
     chainSyncHandle <- ask @ChainSyncHandle
     case e of
-        PublishTx tx  -> liftIO $ MockClient.queueTx txSendHandle tx
+        PublishTx tx  ->
+            case txSendHandle of
+              Nothing ->
+                  -- If the PAB is started with the real node working transactions
+                  -- need to be sent via the wallet, not the mocked server node
+                  -- (which is not actually running).
+                  liftIO $ putStrLn "Cannot send a transaction when connected to the real node."
```
koslambrou, line 57: Should we make this throw an error if this case arises?

plutus-pab/src/Cardano/Node/Mock.hs
```diff
@@ -51,22 +51,25 @@ consumeEventHistory stateVar =
...
     , MonadIO m
     , LastMember m effs
     )
  => Tx -> Eff effs NoContent
 addTx tx = do
     logInfo $ BlockOperation $ NewTransaction tx
     clientHandler <- Eff.ask
-    liftIO $ Client.queueTx clientHandler tx
+    case clientHandler of
+      Nothing      -> logError TxSendCalledWithoutMock
```
koslambrou, line 63: Throw error here?

raduom: There is no error throwing effect on this function. I figured it's not really supposed to throw errors. Do you think we should change that?

raduom: Also, would we make that an effect error, or a Control.Exception one?

raduom: I will add an error effect and see what happens. I think it's reasonable to throw an error if this api is used when it shouldn't be.

raduom: Meta idea: Maybe we should make this distinction at type level so we cannot write code that triggers this behaviour (but this would be another PR).

plutus-pab/src/Cardano/Node/Client.hs
raduom: I will try to add an error effect.

plutus-pab-executables/tx-inject/config.yaml
```diff
@@ -35,7 +36,7 @@ nodeServerConfig:
     - getWallet: 8
     - getWallet: 9
     - getWallet: 10
-  mscNodeMode: AlonzoNode
+  pscNodeMode: AlonzoNode
```
koslambrou, line 39: Since you modified this config file, you could do the same with `plutus-pab-executables/plutus-pab.yaml.sample`?

plutus-pab/src/Cardano/Node/Mock.hs
koslambrou: Yeah makes sense. Create a ticket for that or add a TODO comment so we don't forget?

raduom: This code is only used in the mocked server so I think logging an error is good enough since this is only used for testing.

plutus-pab/src/Cardano/Node/Client.hs
raduom: Done.

plutus-pab-executables/tx-inject/config.yaml
raduom: Done.

plutus-pab/src/Cardano/Node/Mock.hs
raduom: TODO: https://github.com/input-output-hk/plutus-apps/issues/264

silky (approved):

silky:
Thanks for doing all this @raduom ! :)

PR #246, opened 2022-01-13
Fix min Ada bug.

### LINKED ISSUE #237: Emulator transaction balancing doesn't work after fix
### Summary

 The balancer in the emulator does not compute the correct Ada required to cover the minimum Ada
  per output when minting tokens. Our best guess is that this is because it generates separate outputs for
  minted tokens and left-over tokens from consumed outputs, instead of merging them into a single
  output, but computes the Ada requirement as if they were merged.

This happens even after the fixes in PR#229.

The following test fails in `Spec.Contract` in `plutus-contract-tests` (with a small change to the imports):
```Haskell
import Ledger.Scripts (datumHash, mintingPolicyHash, unitDatum, unitRedeemer)
import Ledger.Typed.Scripts.MonetaryPolicies qualified as MPS

balanceTxnMinAda' :: TestTree
balanceTxnMinAda' =
    let vA n = Value.singleton "ee" "A" n
        vB n = Value.singleton "ff" "B" n
        mps  = MPS.mkForwardingMintingPolicy vHash
        vL n = Value.singleton (Value.mpsSymbol $ mintingPolicyHash mps) "L" n
        options = defaultCheckOptions
            & changeInitialWalletValue w1 (<> vA 1 <> vB 2)
        vHash = validatorHash someValidator
        payToWallet w = Constraints.mustPayToPubKey (EM.mockWalletPaymentPubKeyHash w)
        mkTx lookups constraints = Constraints.adjustUnbalancedTx . either (error . show) id $ Constraints.mkTx @Void lookups constraints

        setupContract :: Contract () EmptySchema ContractError ()
        setupContract = do
            -- Make sure there is a utxo with 1 A, 1 B, and 4 ada at w2
            submitTxConfirmed $ mkTx mempty (payToWallet w2 (vA 1 <> vB 1 <> Value.scale 2 (Ada.toValue Ledger.minAdaTxOut)))
            -- Make sure there is a UTxO with 1 B and datum () at the script
            submitTxConfirmed $ mkTx mempty (Constraints.mustPayToOtherScript vHash unitDatum (vB 1))
            -- utxo0 @ wallet2 = 1 A, 1 B, 4 Ada
            -- utxo1 @ script  = 1 B, 2 Ada

        wallet2Contract :: Contract () EmptySchema ContractError ()
        wallet2Contract = do
            utxos <- utxosAt someAddress
            let txOutRef = head (Map.keys utxos)
                lookups = Constraints.unspentOutputs utxos
                        <> Constraints.otherScript someValidator
                        <> Constraints.mintingPolicy mps
                constraints = Constraints.mustSpendScriptOutput txOutRef unitRedeemer                                        -- spend utxo1
                            <> Constraints.mustPayToOtherScript vHash unitDatum (vB 1)                                       -- 2 ada and 1 B to script
                            <> Constraints.mustPayToOtherScript vHash (Datum $ PlutusTx.toBuiltinData (0 :: Integer)) (vB 1) -- 2 ada and 1 B to script (different datum)
                            <> Constraints.mustMintValue (vL 1) -- 1 L and 2 ada to wallet2
                            -- the leftover A doesn't get any ada here
                            -- the right solution would be to merge the 1 L and 2 Ada with the leftover A
                            --
                            -- Note: this bug is flaky - if you change the amount of A at the wallet to 899
                            -- this test passes, if you change it to 900 it fails, and at 901 it passes again.
                            -- This is most likely because the order in which UTxOs appear in the UTxO map depend
                            -- on their hash - so when UTxOs are pulled in to balance the UTxOs
                            -- containing Bs one of two things happen:
                            --  1. Either the precise amount of Ada required to balance existing utxos is found in the UTxO containing A (utxo0)
                            --  but we produce an extra UTxO that contains only A (that never gets any Ada) or
                            --  2. A utxo with a lot of ada is taken in first and is then squashed with the output utxo for As.
            submitTxConfirmed $ mkTx lookups constraints

        trace = do
            Trace.activateContractWallet w1 setupContract
            Trace.waitNSlots 10
            Trace.activateContractWallet w2 wallet2Contract
            Trace.waitNSlots 10

    in checkPredicateOptions options "balancing doesn't create outputs with no Ada (2)" assertNoFailedTransactions (void trace)
```

### Actual Result

The test fails with a `ValueContainsLessThanMinAda` validaton error.

### Expected Result

We expect the test to succeed.

_No response_

### System info

OS: WSL 2 (this really shouldn't matter...)
Plutus: `7f53f18df`

sjoerdvisscher:
Fixes #237.

Very similar to the previous min Ada bug, just in a different place.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-contract/src/Wallet/Emulator/Wallet.hs
+++ b/plutus-contract/src/Wallet/Emulator/Wallet.hs
@@ -46,8 +46,8 @@ import Ledger (Address (addressCredential), CardanoTx, ChainIndexTxOut,
                PaymentPrivateKey (PaymentPrivateKey, unPaymentPrivateKey),
                PaymentPubKey (PaymentPubKey, unPaymentPubKey),
                PaymentPubKeyHash (PaymentPubKeyHash, unPaymentPubKeyHash), PubKeyHash,
-               ScriptValidationEvent (sveScript), StakePubKey, Tx (txFee, txMint), TxIn (TxIn, txInRef), TxOut,
-               TxOutRef, UtxoIndex (UtxoIndex, getIndex), ValidationCtx (ValidationCtx), ValidatorHash, Value)
+               ScriptValidationEvent (sveScript), StakePubKey, Tx (txFee, txMint), TxIn (TxIn, txInRef), TxOutRef,
+               UtxoIndex (UtxoIndex, getIndex), ValidationCtx (ValidationCtx), ValidatorHash, Value)
 import Ledger qualified
 import Ledger.Ada qualified as Ada
 import Ledger.CardanoWallet (MockWallet, WalletNumber)
@@ -443,7 +443,7 @@ adjustBalanceWithMissingLovelace utxo ownPaymentPubKey unBalancedTx (neg, pos) =
         -- minted, and if the positive balance is > 0 and < 'Ledger.minAdaTxOut',
         -- then we adjust it to the minimum Ada.
         missingLovelaceFromPosValue =
-          if Value.isZero posWithMintAda || Ada.fromValue posWithMintAda >= Ledger.minAdaTxOut
+          if valueIsZeroOrHasMinAda posWithMintAda
             then 0
             else max 0 (Ledger.minAdaTxOut - Ada.fromValue posWithMintAda)
         -- We calculate the final negative and positive balances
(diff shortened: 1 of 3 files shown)


koslambrou (approved):

ak3n:
@sjoerdvisscher Thanks!

PR #250, opened 2022-01-14
Add github test to check for uncommitted updates

shlevy:


jhbertra (approved):
Fancy!

jhbertra:
So fancy the new tests broke the build :cry:

PR #275, opened 2022-01-24
SCP-3305 SCP-3263 fixed Ledger.Constraints.Offchain.updateUtxoIndex

bwbush:
Prior to this commit, `Ledger.Constraints.Offchain.updateUtxoIndex` would discard lookups for inputs that aren't script outputs. However, sometimes it is necessary to include particular public-key `TxIn`s in a transaction, via `Plutus.Contract.Wallet.ExportTx.lookups`: a good example of this is when a `Plutus.Contract.Currency.OneShotCurrency` needs to consume a specified `UTxO` in its minting policy. Even though `Plutus.Constraints.Tx.mustSpendPubKeyOutput` adds a public-key input to the lookups, `updateUtxoIndex` discards that, with the result that it is only by chance that `cardano-wallet` would select that necessary input for the balanced transaction and provide it to Plutus validators. (Note that `cardano-wallet` does not automatically provide the `TxIn` in the partially constructed transaction to validators: it only provides `Cardano.Wallet.PartialTx.inputs` and the inputs it has semi-randomly chosen.) The consequence of all of this is that scripts that require a particular input will randomly fail during balancing in wallets with more than one UTxO.

This fix simply adds public-key inputs to the type `Ledger.Constraints.OffChain.ScriptOutput` so that it can also hold public-key inputs. (There might be a more descriptive name than `ScriptOutput` now.) It makes minor adjustments to the chain of functions that pass this into an unbalanced transaction before it is sent to `cardano-wallet`.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/playground-common/src/PSGenerator/Common.hs
+++ b/playground-common/src/PSGenerator/Common.hs
@@ -22,7 +22,7 @@ import Ledger (Address, BlockId, ChainIndexTxOut, DatumHash, MintingPolicy, OnCh
                PubKey, PubKeyHash, RedeemerPtr, ScriptTag, Signature, StakePubKey, StakePubKeyHash, StakeValidator, Tx,
                TxId, TxIn, TxInType, TxOut, TxOutRef, TxOutTx, UtxoIndex, ValidationPhase, Validator)
 import Ledger.Ada (Ada)
-import Ledger.Constraints.OffChain (MkTxError, ScriptOutput, UnbalancedTx)
+import Ledger.Constraints.OffChain (MkTxError, UnbalancedTx)
 import Ledger.Credential (Credential, StakingCredential)
 import Ledger.DCert (DCert)
 import Ledger.Index (ExCPU, ExMemory, ScriptType, ScriptValidationEvent, ValidationError)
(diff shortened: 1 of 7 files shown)


raduom (approved):
This looks good to me, but I have not worked in this area recently, so I asked @sjoerdvisscher for a review also.

sjoerdvisscher:
The reason we removed public-key inputs is that `cardano-wallet` doesn't support it. See #57.
So it might be that we were wrongly informed, but did you check that this works with `cardano-wallet`?

sjoerdvisscher (changes requested):
If we do this, the way to go would be to revert #57.

bwbush:
> The reason we removed public-key inputs is that `cardano-wallet` doesn't support it. See #57. So it might be that we were wrongly informed, but did you check that this works with `cardano-wallet`?

Yes, I've tested this will wallet and actual Marlowe contract endpoints. The `MustSpendPubKeyOutput` could have been named `MustSpendWalletOutput` because it just provides a `UTxO` that could reside at any of the wallet's addresses.

I stepped through each phase of wallet's processing of the request from the PAB to see how it balances the transaction: the wallet makes provision for spending particular UTxOs in the [`PartialTx.inputs`](https://github.com/input-output-hk/cardano-wallet/blob/5ad348d26313363bacec061ee963c89fd9d467b8/lib/core/src/Cardano/Wallet.hs#L1396), which is poplulated from `ExportTx`. The wallet only provides UTxOs to scripts if they are specified in `PartialTx.inputs` or if the wallet happens to select the UTxO when it balances; unfortunately, the wallet does not pass the `TxIn`s contained in `PartialTx.partialTx :: Tx AlonzoEra` to scripts involved in the transaction.

bwbush:
> If we do this, the way to go would be to revert #57.

It looks like reverting #57 would work, but I'd like to test that.

bwbush:
@sjoerdvisscher, I reverted #57 in d23de941d7b851764d77d05f846bea513f565e3a. The Marlowe PAB integration test passes (running the full Marlowe contract on our private testnet) using this revision. Would you please re-review? Thanks.

sjoerdvisscher (approved):
I like all this red :)

PR #292, opened 2022-02-02
Use slotConfig from the config for awaitTime

### LINKED ISSUE #284: Use of default SlotConfig in awaitTime of PAB ContractInstance requestHandler
### Summary

The use of the default `SlotConfig` in `awaitTime` function defined in `Plutus.PAB.Core.ContractInstance.STM` may impact script validation when running PAB with a private testnet such that, the beginning of slot zero and the slot length may not correspond to the default public testnet configuration. Indeed, `pscSlotConfig` specified in the `nodeServerConfig` must be considered instead.

1. Define an endpoint calling `awaitTime 0` and perform a `tell` on the result
2. Invoke the endpoint with a PAB context running a fresh instance of a local testnet
2. Check the observable state of the corresponding contract instance.

### Actual Result

The produced POSIXTime will differ from the local testnet slotconfig.

### Expected Result

The produced POSIXTime should be in conformity with the local testnet slotconfig.

Any BlockchainEvent request to a contract instance must be parameterized with the node configuration.

A current workaround is to explicitly pass the slotConfig to the invoked endpoint and replicate what awaitTime does.  

### System info

plutus commit : 51da951ee2a7b8f05bde624bc9da4ce9a1f81ea1

ak3n:
fix #284

---

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-pab/src/Plutus/PAB/Core/ContractInstance/BlockchainEnv.hs
+++ b/plutus-pab/src/Plutus/PAB/Core/ContractInstance/BlockchainEnv.hs
@@ -56,7 +56,7 @@ startNodeClient ::
   -> InstancesState -- ^ In-memory state of running contract instances
   -> IO BlockchainEnv
 startNodeClient socket mode rollbackHistory slotConfig networkId resumePoint instancesState = do
-    env <- STM.atomically $ emptyBlockchainEnv rollbackHistory
+    env <- STM.atomically $ emptyBlockchainEnv rollbackHistory (Just slotConfig)
     case mode of
       MockNode -> do
         void $ MockClient.runChainSync socket slotConfig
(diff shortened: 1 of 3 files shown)


raduom (approved):
This works, but I was thinking a better solution would be to eventually parse the node configuration and get the slotting configuration from that.

ak3n:
>This works, but I was thinking a better solution would be to eventually parse the node configuration and get the slotting configuration from that.

Then we should remove these settings from the PAB configuration?

raduom:
> > This works, but I was thinking a better solution would be to eventually parse the node configuration and get the slotting configuration from that.
> 
> Then we should remove these settings from the PAB configuration?

I think that was the idea. But that would be a pretty big change.

PR #302, opened 2022-02-07
Switch `UtxoSetAtAddress` and `TxoSetAtAddress` from using `Credential` to `Address`

sorki:
Mostly cosmetic fix with no functionality changes. This allows handler to work with full address if staking credential is provided which is required when querying utxos from e.g. db-sync which uses `bech32` encoded address with staking credential. Current beam handler just grabs `Credential` from `addressCredential` which is now an implementation detail.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


sorki:
Looks like it's not really needed since dbsync stores `TxOut PaymentCred` as well as bech32 address. Docs for the field seems to match requirements - _The payment credential part of the Shelley address. (NULL for Byron addresses). For a script-locked address, this is the script hash._

koslambrou:
@sorki Then I'll close this

PR #305, opened 2022-02-07
SCP-3343 fixed intermittent rollback failures in PAB

bwbush:
The cause of the rollback failure was that `Plutus.ChainIndex.UtxoState.rollbackWith` computes the current tip of the index as the slot of the last transaction, ignoring the fact that there may be empty blocks after that. When a rollback with a target falls within such a sequence of empty blocks, the `rollbackWith` function misinterprets this as a `TipMismatch`.

This change simply checks to see if the rollback is unnecessary because it is
through a sequence of empty blocks ending at the former tip.

This was manually tested on the private Marlowe testnet, where we verified that rollback failures no long occur when they just rollback through empty blocks.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


sjoerdvisscher (approved):
LGTM. @raduom can you verify?

raduom (approved):

PR #309, opened 2022-02-09
SCP-3431: Better syncing logging in the chain-index.

koslambrou:
Summary of changes:

* Syncing message now logs every 30s

* Prints syncing percentage, number of applied blocks, number of rollbacks and current tip

* Logging uses the Trace for more consistent logging with PAB and cardano-node.

* Added warning flag for missing import list

Example of new chain-index logging (the number shown are not the actual values):

```
[chain-index:Warning:30] [2022-02-09 13:59:51.71 UTC] No synced blocks in the last 30s.
[chain-index:Info:30] [2022-02-09 13:59:54.71 UTC] Syncing (7.38%). Applied 7 blocks, 1 rollbacks in the last 30s. Current tip is Point(Slot 3898564, BlockId 93e58766e4ab743ea07d5f2a070cb9c3e7b95db160d04203c4764fb1050d422d)
[chain-index:Info:30] [2022-02-09 13:59:57.72 UTC] Syncing (7.38%). Applied 67 blocks, 0 rollbacks in the last 30s. Current tip is Point(Slot 3898572, BlockId 1df9f2d8967be312bdadef8649e0738101c5660096343c17d4ca1a34b66f7f59)
[chain-index:Warning:30] [2022-02-09 14:01:24.80 UTC] No synced blocks in the last 30s.
[chain-index:Error:34] [2022-02-09 14:01:47.22 UTC] ChainIndexError: Error during Beam operation: SqlError (via Beam): SQLite3 returned ErrorMisuse while attempting to perform prepare "ROLLBACK TRANSACTION": bad parameter or other API misuse
```

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-chain-index/src/Plutus/ChainIndex/Lib.hs
+++ b/plutus-chain-index/src/Plutus/ChainIndex/Lib.hs
@@ -1,8 +1,10 @@
-{-# LANGUAGE DataKinds         #-}
-{-# LANGUAGE LambdaCase        #-}
-{-# LANGUAGE NamedFieldPuns    #-}
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE TupleSections     #-}
+{-# LANGUAGE DataKinds          #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleContexts   #-}
+{-# LANGUAGE GADTs              #-}
+{-# LANGUAGE LambdaCase         #-}
+{-# LANGUAGE OverloadedStrings  #-}
+{-# LANGUAGE TupleSections      #-}
 {-| Using the chain index as a library.
 
 A minimal example that just syncs the chain index:
@@ -10,7 +12,7 @@ A minimal example that just syncs the chain index:
 @
 `withDefaultRunRequirements` $ \runReq -> do
 
-    syncHandler <- `showingProgress` $ `defaultChainSyncHandler` runReq
+    syncHandler <- `defaultChainSyncHandler` runReq
 
     `syncChainIndex` `defaultConfig` runReq syncHandler
 
@@ -35,7 +37,7 @@ module Plutus.ChainIndex.Lib (
     , defaultChainSyncHandler
     , storeFromBlockNo
     , filterTxs
-    , showingProgress
+    , writeChainSyncEventToChan
     -- * Utils
     , getTipSlot
 ) where
@@ -46,7 +48,6 @@ import Control.Monad.Freer.Extras.Beam (BeamEffect, BeamLog (SqlLog))
 import Control.Monad.Freer.Extras.Log qualified as Log
 import Data.Default (def)
 import Data.Functor (void)
-import Data.IORef (IORef, newIORef, readIORef, writeIORef)
 import Database.Beam.Migrate.Simple (autoMigrate)
 import Database.Beam.Sqlite qualified as Sqlite
 import Database.Beam.Sqlite.Migrate qualified as Sqlite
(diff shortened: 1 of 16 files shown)


plutus-chain-index/src/Plutus/ChainIndex/SyncStats.hs
```diff
@@ -0,0 +1,102 @@
...
+    => TChan ChainSyncEvent
+    -> Eff effs ()
+logProgress broadcastChan = do
+    chan <- liftIO $ atomically $ dupTChan broadcastChan
+    forever $ do
+        liftIO $ threadDelay 30000000 -- 30s
+        (SyncStats a b chainSyncPoint nodePoint) <- liftIO $ generateSummary chan
+        case fmap fst $ preview CI._Point nodePoint of
+            Nothing -> do
+                logWarn NoNewBlocksSyncLog
```
sjoerdvisscher, line 82: I guess you'd only want to see this while syncing, so only while `chainSyncPoint` has not reached `nodePoint`.

plutus-chain-index/src/Plutus/ChainIndex/Lib.hs
```diff
@@ -170,26 +171,16 @@ getTipSlot config = do
...
+    -> ChainSyncHandler
+    -> IO ChainSyncHandler
+writeChainSyncEventToChan tchan handler = do
     pure $ \case
-        (RollForward block@(CI.Block (CI.Tip blockSlot _ _) _) tip@(CI.Tip tipSlot _ _)) -> do
-            showProgress lastProgressRef tipSlot blockSlot
-            handler $ RollForward block tip
-        evt -> handler evt
+        evt -> do
+            atomically $ writeTChan tchan evt
```
sjoerdvisscher, line 182: These events can be pretty big I think? Dumping everything in a channel for 30s doesn't sound like a good idea. I guess you'd want a function like

```haskell
writeToChan
     :: (ChainSyncEvent -> a)
     -> TChan a
     -> ChainSyncHandler
     -> IO ChainSyncHandler
```

so you can only keep in memory what you need.

sjoerdvisscher (changes requested):

plutus-chain-index/src/Plutus/ChainIndex/SyncStats.hs
koslambrou: Mmm.. good point. However, I think we would still want to see how many blocks were added even when `chainSyncPoint` reached `nodePoint`. When that happens, we could log every 300s instead of 30s? That's what `cardano-wallet` does.

plutus-chain-index/src/Plutus/ChainIndex/Lib.hs
koslambrou: Right! :)

plutus-chain-index/src/Plutus/ChainIndex/SyncStats.hs
sjoerdvisscher: Sorry, I meant specifically the "no new blocks" message.

plutus-chain-index/src/Plutus/ChainIndex/SyncStats.hs
```diff
@@ -0,0 +1,141 @@
...
+--
+-- TODO: The syncing percentage is valid when the node is already fully synced.
+-- But when the node and chain-index are started at the same time, the syncing
+-- percentage is not a valid number considering the actual tip of the node.
+-- Find a better way to calculate this percentage.
+getSyncStateFromStats :: SyncStats -> SyncState
+getSyncStateFromStats (SyncStats _ _ chainSyncPoint nodePoint) =
+    case (chainSyncPoint, nodePoint) of
+        (_, PointAtGenesis) -> NotSyncing
+        (CI.PointAtGenesis, CI.Point _ _) -> NotSyncing
```
sjoerdvisscher, line 118: Does this mean that on the first run of the chain index you'll only see the first update after 5 minutes?
I think we can assume syncing will start soon in this case so `Syncing 0`?

plutus-chain-index/src/Plutus/ChainIndex/SyncStats.hs
```diff
@@ -0,0 +1,141 @@
...
+        (p1, p2) | p1 == p2 -> Synced
```
sjoerdvisscher, line 119: Since new transactions keep coming in, I wonder if this state is ever reached. In my old code this was fixed implicitly because rounding would show `100%` at some point and this would then never be shown again. I believe the cardano wallet has defined a margin of error somewhere that says: if we're this close to the node point, we call syncing done.

plutus-chain-index/src/Plutus/ChainIndex/SyncStats.hs
koslambrou: You're right! I'll change it.

plutus-chain-index/src/Plutus/ChainIndex/SyncStats.hs
koslambrou: Mmm.. Then I can do something similar: `(p1, p2) | p2 - p1 < delta -> Synced`

koslambrou: @sjoerdvisscher I don't know the best number, so I did:

```
        (CI.Point (Slot chainSyncSlot) _, CI.Point (Slot nodeSlot) _) | nodeSlot - chainSyncSlot < 100 -> Synced
```

Should be reasonable for now.

sjoerdvisscher (approved):
💯

PR #314, opened 2022-02-14
ContractModel: Add wiggle room options to NoLockedFunds

MaximilianAlgehed:
Some contracts, like the Uniswap example contract, have predictable overhead and rounding errors that prevent the strict NoLockedFunds property from passing. This PR introduces a little bit of wiggle room to allow for such contracts.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
+++ b/plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
@@ -136,10 +136,9 @@ module Plutus.Contract.Test.ContractModel.Internal
     --
     -- $noLockedFunds
     , NoLockedFundsProof(..)
+    , defaultNLFP
     , checkNoLockedFundsProof
     , checkNoLockedFundsProofFast
-    , checkNoLockedFundsProofWithWiggleRoom
-    , checkNoLockedFundsProofWithWiggleRoomFast
     -- $checkNoPartiality
     , Whitelist
     , whitelistOk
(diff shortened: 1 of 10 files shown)


MaximilianAlgehed:
@sjoerdvisscher Do you mind looking at this when you have a gap in your schedule?

plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
```diff
@@ -1578,63 +1591,56 @@ checkNoLockedFundsProof
...
-        prop s =
+        prop s0 s =
           let lockedVal = lockedValue s
-          in (counterexample ("Locked funds should be zero, but they are\n  " ++ show lockedVal) $ symIsZero lockedVal)
+          in (counterexample ("Locked funds should be at most " ++ show (overhead s0) ++ ", but they are\n  " ++ show lockedVal)
+            $ symLeq lockedVal (overhead s0))
 
-        walletProp as w bal = DL.forAllUniqueDL (nextVarIdx as) (stateAfter $ Actions as) (DL.action (Unilateral w) >> walletStrat w) $ \ acts ->
+        walletProp s0 as w bal =
+          let wig = wiggle s0
```
sjoerdvisscher, line 1629: Is this code so slow it needs to sit outside of the `DL.forAllUniqueDL` loop? 
(It looks a bit like a hack in its current form.)

sjoerdvisscher (approved):

plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
MaximilianAlgehed: The problem is that it's user code - so we don't know how fast or slow it is. Therefore we try to limit the number of footguns. On the other hand - it's only used twice when reporting errors so I don't think performance here is a super big deal.

MaximilianAlgehed:
Done :)

PR #315, opened 2022-02-16
SCP-3501: Remove txs from chain index

ak3n:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-chain-index-core/src/Plutus/ChainIndex/Api.hs
+++ b/plutus-chain-index-core/src/Plutus/ChainIndex/Api.hs
@@ -25,10 +25,9 @@ import Data.OpenApi qualified as OpenApi
 import Data.Proxy (Proxy (..))
 import GHC.Generics (Generic)
 import Ledger (AssetClass, Datum, DatumHash, MintingPolicy, MintingPolicyHash, Redeemer, RedeemerHash, StakeValidator,
-               StakeValidatorHash, TxId, Validator, ValidatorHash)
+               StakeValidatorHash, Validator, ValidatorHash)
 import Ledger.Credential (Credential)
-import Ledger.Tx (ChainIndexTxOut, TxOutRef)
-import Plutus.ChainIndex.Tx (ChainIndexTx)
+import Ledger.Tx (TxOutRef)
 import Plutus.ChainIndex.Types (Diagnostics, Tip)
 import Servant qualified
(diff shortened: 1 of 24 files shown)


plutus-chain-index-core/src/Plutus/ChainIndex/Handlers.hs
```diff
@@ -82,7 +82,6 @@ handleQuery = \case
     MintingPolicyFromHash hash  -> getScriptFromHash hash
     RedeemerFromHash hash       -> getRedeemerFromHash hash
     StakeValidatorFromHash hash -> getScriptFromHash hash
-    TxFromTxId txId             -> getTxFromTxId txId
     TxOutFromRef tor            -> getTxOutFromRef tor
```
koslambrou, line 85: I would change this to `UnspentTxOutFromRef` and `getUtxoutFromRef`

koslambrou (changes requested):
The implementation looks good. However, I would also add a property test to ensure this works as expected. Here's an idea.

In HandlerSpec, you can add a property test which:
- calls `appendBlock`
- for all possibles addresses (or at least one address), call `utxoSetAtAddress` 
- Then call `getUtxoutFromRef` for each utxo
- Verify the result is not Nothing

plutus-chain-index-core/src/Plutus/ChainIndex/Api.hs
```diff
@@ -147,11 +146,9 @@ type API
     = "healthcheck" :> Description "Is the server alive?" :> Get '[JSON] NoContent
     :<|> "from-hash" :> FromHashAPI
     :<|> "tx-out" :> Description "Get a transaction output from its reference." :> ReqBody '[JSON] TxOutRef :> Post '[JSON] ChainIndexTxOut
```
koslambrou, line 148: Change this as well to refer to unspent txouts only.

plutus-chain-index-core/src/Plutus/ChainIndex/Handlers.hs
ak3n: Should we rename everything in plutus-contract then?

koslambrou: Yep!

koslambrou (approved):

PR #319, opened 2022-02-18
SCP-3502: Add AppendBlocks effect

ak3n:
This should increase the speed of the chain index.

The PR also adds two new config parameters:
- `appendPeriod` — How often the appending of blocks should happen (in microseconds)
- `appendBatchSize` — The number of blocks to collect before writing to the database 

---

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-chain-index/src/Plutus/ChainIndex/Events.hs
+++ b/plutus-chain-index/src/Plutus/ChainIndex/Events.hs
@@ -22,6 +22,8 @@ import Plutus.ChainIndex.Lib (ChainSyncEvent (Resume, RollBackward, RollForward)
 import Plutus.ChainIndex.SyncStats (SyncLog, logProgress)
 import Plutus.Monitoring.Util (PrettyObject (PrettyObject), convertLog, runLogEffects)
 
+-- | 'processEventsQueue' reads events from 'TBQueue', collects enough 'RollForward's to
+-- append blocks at once.
 processEventsQueue :: Trace IO (PrettyObject SyncLog) -> RunRequirements -> EventsQueue -> Int -> IO ()
 processEventsQueue trace runReq eventsQueue period = void $ do
   putStrLn "Starting processing of events"
--- a/plutus-chain-index/src/Plutus/ChainIndex/Events.hs
+++ b/plutus-chain-index/src/Plutus/ChainIndex/Events.hs
@@ -23,10 +23,14 @@ import Plutus.ChainIndex.SyncStats (SyncLog, logProgress)
 import Plutus.Monitoring.Util (PrettyObject (PrettyObject), convertLog, runLogEffects)
 import System.Clock (Clock (Monotonic), diffTimeSpec, getTime)
 
+-- | How often do we check the queue
+period :: Int
+period = 5_000_000 -- 5s
+
 -- | 'processEventsQueue' reads events from 'TBQueue', collects enough 'RollForward's to
 -- append blocks at once.
-processEventsQueue :: Trace IO (PrettyObject SyncLog) -> RunRequirements -> EventsQueue -> Int -> IO ()
-processEventsQueue trace runReq eventsQueue period = void $ do
+processEventsQueue :: Trace IO (PrettyObject SyncLog) -> RunRequirements -> EventsQueue -> IO ()
+processEventsQueue trace runReq eventsQueue = void $ do
   putStrLn "Starting processing of events"
   go []
   where
--- a/plutus-chain-index/src/Plutus/ChainIndex/Events.hs
+++ b/plutus-chain-index/src/Plutus/ChainIndex/Events.hs
@@ -13,7 +13,7 @@ module Plutus.ChainIndex.Events where
 import Cardano.BM.Trace (Trace)
 import Control.Concurrent (threadDelay)
 import Control.Concurrent.STM (atomically, flushTBQueue, isFullTBQueue)
-import Control.Monad (void)
+import Control.Monad (void, forever)
 import Data.Functor ((<&>))
 import Data.Maybe (catMaybes)
 import Plutus.ChainIndex qualified as CI
@@ -21,7 +21,7 @@ import Plutus.ChainIndex.Lib (ChainSyncEvent (Resume, RollBackward, RollForward)
                               runChainIndexDuringSync)
 import Plutus.ChainIndex.SyncStats (SyncLog, logProgress)
 import Plutus.Monitoring.Util (PrettyObject (PrettyObject), convertLog, runLogEffects)
-import System.Clock (Clock (Monotonic), diffTimeSpec, getTime)
+import System.Clock (Clock (Monotonic), diffTimeSpec, getTime, TimeSpec)
 
 -- | How often do we check the queue
 period :: Int
@@ -30,42 +30,29 @@ period = 5_000_000 -- 5s
 -- | 'processEventsQueue' reads events from 'TBQueue', collects enough 'RollForward's to
 -- append blocks at once.
 processEventsQueue :: Trace IO (PrettyObject SyncLog) -> RunRequirements -> EventsQueue -> IO ()
-processEventsQueue trace runReq eventsQueue = go []
+processEventsQueue trace runReq eventsQueue = forever $ do
+  start <- getTime Monotonic
+  eventsToProcess <- do
+    let
+      waitUntilEvents = do
+        isFull <- atomically $ isFullTBQueue eventsQueue
+        if isFull then atomically $ flushTBQueue eventsQueue
+        else threadDelay period >> waitUntilEvents
+    waitUntilEvents
+  processEvents start eventsToProcess
   where
-    go unprocessedEvents = do
-      start <- getTime Monotonic
-      eventsToProcess <-
-        if null unprocessedEvents then do
+    processEvents :: TimeSpec -> [ChainSyncEvent] -> IO ()
+    processEvents start events = case events of
+      (Resume resumePoint) : (RollBackward backwardPoint _) : restEvents -> do
+        void $ runChainIndexDuringSync runReq $ do
+          CI.rollback backwardPoint
+          CI.resumeSync resumePoint
+        processEvents start restEvents
+      rollForwardEvents -> do
           let
-            waitUntilEvents = do
-              isFull <- atomically $ isFullTBQueue eventsQueue
-              if isFull then atomically $ flushTBQueue eventsQueue
-              else threadDelay period >> waitUntilEvents
-          waitUntilEvents
-        else return unprocessedEvents
-      case eventsToProcess of
-        firstEvent : restEvents -> do
-          -- rollback or resume
-          void $ runChainIndexDuringSync runReq $ case firstEvent of
-            (RollBackward point _) -> CI.rollback point
-            (Resume point)         -> CI.resumeSync point
-            _                      -> pure () -- ignore forward block
-          -- append blocks
-          let
-            -- if the first event is 'RollForward' then process all events, otherwise only the tail
-            eventsToProcess' = case firstEvent of
-              (RollForward _ _) -> eventsToProcess
-              _                 -> restEvents
-            isRollForward = \case { (RollForward _ _) -> True; _ -> False }
-            (rollForwardEvents, restUnprocessedEvents) = span isRollForward eventsToProcess'
             blocks = catMaybes $ rollForwardEvents <&> \case
               (RollForward block _) -> Just block
               _                     -> Nothing
           void $ runChainIndexDuringSync runReq $ CI.appendBlocks blocks
-          -- we should include 'firstEvent' to processed events if it was a rollback or a resume
-          let processedEvents = (if isRollForward firstEvent then [] else [firstEvent]) ++ rollForwardEvents
           end <- getTime Monotonic
-          void $ runLogEffects (convertLog PrettyObject trace) $ logProgress processedEvents (diffTimeSpec end start)
-          go restUnprocessedEvents
-        [] -> error "The queue can't be empty"
-      go []
+          void $ runLogEffects (convertLog PrettyObject trace) $ logProgress events (diffTimeSpec end start)
--- a/plutus-chain-index/src/Plutus/ChainIndex/Events.hs
+++ b/plutus-chain-index/src/Plutus/ChainIndex/Events.hs
@@ -30,9 +30,7 @@ period = 5_000_000 -- 5s
 -- | 'processEventsQueue' reads events from 'TBQueue', collects enough 'RollForward's to
 -- append blocks at once.
 processEventsQueue :: Trace IO (PrettyObject SyncLog) -> RunRequirements -> EventsQueue -> IO ()
-processEventsQueue trace runReq eventsQueue = void $ do
-  putStrLn "Starting processing of events"
-  go []
+processEventsQueue trace runReq eventsQueue = go []
   where
     go unprocessedEvents = do
       start <- getTime Monotonic
@@ -69,5 +67,5 @@ processEventsQueue trace runReq eventsQueue = void $ do
           end <- getTime Monotonic
           void $ runLogEffects (convertLog PrettyObject trace) $ logProgress processedEvents (diffTimeSpec end start)
           go restUnprocessedEvents
-        [] -> putStrLn "The queue can't be empty"
+        [] -> error "The queue can't be empty"
       go []
--- a/plutus-chain-index/src/Plutus/ChainIndex/Events.hs
+++ b/plutus-chain-index/src/Plutus/ChainIndex/Events.hs
@@ -21,6 +21,7 @@ import Plutus.ChainIndex.Lib (ChainSyncEvent (Resume, RollBackward, RollForward)
                               runChainIndexDuringSync)
 import Plutus.ChainIndex.SyncStats (SyncLog, logProgress)
 import Plutus.Monitoring.Util (PrettyObject (PrettyObject), convertLog, runLogEffects)
+import System.Clock (Clock (Monotonic), diffTimeSpec, getTime)
 
 -- | 'processEventsQueue' reads events from 'TBQueue', collects enough 'RollForward's to
 -- append blocks at once.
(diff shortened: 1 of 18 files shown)


plutus-chain-index/src/Plutus/ChainIndex/Lib.hs
```diff
@@ -131,14 +137,21 @@ toCardanoChainSyncHandler runReq handler = \case
...
-    (RollForward block _)  -> appendBlock block
+    (RollForward block _)  -> appendBlocks [block]
     (RollBackward point _) -> rollback point
     (Resume point)         -> resumeSync point
 
+storeChainSyncHandler :: BlocksChan -> RunRequirements -> ChainSyncHandler
+storeChainSyncHandler blocksChan runReq evt = case evt of
+    (RollForward block _)  -> atomically $ writeTChan blocksChan block
+    (RollBackward point _) -> void $ runChainIndexDuringSync runReq $ rollback point
+    (Resume point)         -> void $ runChainIndexDuringSync runReq $ resumeSync point
```
koslambrou, line 153: Doing it this way, I think you have a concurrency problem.

My understanding of your code is that it writes blocks to the TChan and the appendBlocks is handled by a separate thread.

However, what might happen here is that the `RollBackward` event is called before the thread processing the `RollForward` blocks is actually finished. So, the `RollBackward` is called, it updates the db to a different point, and the thread processes blocks that should have been rollbacked.

I think the solution is to use the `ChainSyncEvent` type in the TChan, batch the `RollForward` events until a `RollBackward` is met and use `appendBlocks`. Something like that...

ak3n: Hmmm, yeah, I agree, IO operations with one resource should happen in one thread. I haven't thought much writing this, was trying to make the `appendBlocks` work. Will fix it. Thanks!

ak3n: >I think the solution is to use the ChainSyncEvent type in the TChan, batch the RollForward events until a RollBackward is met and use appendBlocks. Something like that...

Just reading `RollForward` events until `RollBackward` looks fragile as we keep a lot of `RollForward`s in memory and it's very easy to loose the processed state.

koslambrou:
Got this when syncing in the beginning of testnet:

```
[chain-index:Error:118] [2022-03-07 18:39:24.21 UTC] ChainIndexError: Error during Beam operation: SqlError (via Beam): SQLite3 returned ErrorError while attempting to perform prepare "INSERT INTO \"tips\"(\"row_slot\", \"row_block_id\", \"row_block_number\") VALUES (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?) ON CONFLICT  DO NOTHING": too many SQL variables
```

koslambrou (changes requested):
- The new configs should also be added in CommandLine.hs
- The Resume doesn't seem to work anymore. When I stop the syncing and restart, it always restart from zero.
- The default 15k of batch size uses too much memory. Tried it on testnet and on about 20% syncing, more than 20GB was used. Maybe 1k? Or 5k?

We should strive to make this work on testnet at least, at optimize later for mainnet.

ak3n:
@koslambrou 

>The new configs should also be added in CommandLine.hs

At first I added them as CLI params but then thought a bit and decided that that params are better to be config params as they shouldn't be updated that often (very similar to `cicSecurityParam` and `cicStoreFrom`). Will add them as cli params as well.

>The Resume doesn't seem to work anymore. When I stop the syncing and restart, it always restart from zero.

Hmmm. This is strange, it works for me — https://gist.github.com/ak3n/702b62fd3d86b1f8aaf54db5d195bfdf

Could you please describe how to reproduce it? Maybe there is a specific scenario.

>The default 15k of batch size uses too much memory. Tried it on testnet and on about 20% syncing, more than 20GB was used. Maybe 1k? Or 5k?

Ok, this is what I was afraid of. It depends on the system. Every user should find the optimal params. 15k works for me with 8gb of RAM on macos. I suppose it heavily uses the swapping to make it work.

koslambrou:
Mmm... there's definitely something wrong with resuming. 

The command I used was: `cabal run plutus-chain-index -- start-index --network-id 1097911063 --socket-path /tmp/cardano-node/testnet/node.sock --db-path /tmp/cardano-node/testnet/chain-index.db`

Here's the last log I got:

```
...
[chain-index:Info:29] [2022-03-08 18:56:22.06 UTC] Syncing (85.42%). Applied 2864 blocks, 0 rollbacks in the last 30s. Current tip is Point(Slot 44758433, BlockId 7ff89d67303ecca9540d3a7e284f47a8eb44b29faf9f10918c3d2275db55e8c2)
```

After simply restarting the chain-index, here's the first log I get:

```
[chain-index:Info:29] [2022-03-09 12:50:12.22 UTC] Syncing (0.40%). Applied 36284 blocks, 0 rollbacks in the last 30s. Current tip is Point(Slot 210875, BlockId 5546a784b4ab9a935afb553e29b52ebaba581869f4a5a602eafe2939ba6ba36f)
```

koslambrou:
Just tried it on `main`, and everything works as expected. `main` manages to resume correctly, and no sql errors.

locallycompact:
I'm running a modification of this branch with pab-cli and hardcoding the append-batch-size option to 1000.

```
@@ -335,6 +335,7 @@ startChainIndex networkId
                <> " --db-path " <> getChainIndexDbPath (createChainIndexDbPath nodeOptsOutputDir)
                <> " --port " <> show (unChainIndexPort chainIndexOptsPort)
                <> " --network-id " <> show (C.unNetworkMagic $ C.toNetworkMagic networkId)
+               <> " --append-batch-size " <> (show 1000)
```

Running for a few hours now and the pab application itself is using over 80% of 64GB RAM system. CPU is a threadripper 1950x (16C/32T).

```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
 185920 lc        20   0 1030.4g   3.6g  24920 S  1262   5.8   2342:49 plutus-chain-in
 185784 lc        20   0 1025.4g   4.3g  29824 S 199.0   6.8 393:23.21 cardano-node
 185976 lc        20   0 1024.3g  49.8g  34116 R   9.0  79.4  34:16.97 multi-pab
 185924 lc        20   0 1024.7g 116304  34584 S   1.0   0.2  10:11.52 cardano-wallet
 ```

**Comment by ak3n** [comment]
@locallycompact thanks for reporting! That approves that `chain-index` is in a better form now.

Have no idea what might be wrong with PAB yet.

**Comment by locallycompact** [comment]
I'm also experiencing the reset from zero on restart. Any idea what could be causing this?

**Comment by ak3n** [comment]
I suspect it's because of a missing transaction. We delete tips and other stuff in `reduceOldUtxoDb` before inserting new tips. By stopping `chain-index` somewhere between we break the db state.

Working on a fix at the moment.

File: plutus-chain-index/src/Plutus/ChainIndex/Events.hs [hunk 2]
```diff
@@ -0,0 +1,73 @@
...
+            blocks = catMaybes $ rollForwardEvents <&> \case
+              (RollForward block _) -> Just block
+              _                     -> Nothing
+          void $ runChainIndexDuringSync runReq $ CI.appendBlocks blocks
+          -- we should include 'firstEvent' to processed events if it was a rollback or a resume
+          let processedEvents = (if isRollForward firstEvent then [] else [firstEvent]) ++ rollForwardEvents
+          end <- getTime Monotonic
+          void $ runLogEffects (convertLog PrettyObject trace) $ logProgress processedEvents (diffTimeSpec end start)
+          go restUnprocessedEvents
+        [] -> putStrLn "The queue can't be empty"
```
koslambrou, line 72: Use `logWarn` here.

plutus-chain-index/src/Plutus/ChainIndex/Events.hs
```diff
@@ -0,0 +1,73 @@
...
+
+-- | How often do we check the queue
+period :: Int
+period = 5_000_000 -- 5s
+
+-- | 'processEventsQueue' reads events from 'TBQueue', collects enough 'RollForward's to
+-- append blocks at once.
+processEventsQueue :: Trace IO (PrettyObject SyncLog) -> RunRequirements -> EventsQueue -> IO ()
+processEventsQueue trace runReq eventsQueue = void $ do
+  putStrLn "Starting processing of events"
```
koslambrou, line 34: Remove or replace with `logInfo`

koslambrou (changes requested):

plutus-chain-index/src/Plutus/ChainIndex/SyncStats.hs
```diff
@@ -79,31 +76,15 @@ instance Pretty SyncState where
...
-              logInfo $ SyncLog syncState syncStats (fromMicroseconds delay)
-              go chan 30_000_000 -- 30s
+-- | Log syncing summary.
+logProgress :: forall effs. (Member (LogMsg SyncLog) effs) => [ChainSyncEvent] -> TimeSpec -> Eff effs ()
+logProgress events period = do
+    let syncStats = foldl (<>) mempty $ map convertEventToSyncStats events
+    let syncState = getSyncStateFromStats syncStats
+    let syncLog = SyncLog syncState syncStats (fromMicroseconds $ toNanoSecs period `div` 1000)
+    case syncState of
+      NotSyncing -> pure () -- logWarn syncLog
```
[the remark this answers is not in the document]
ak3n: To turn off `Not syncing. Applied 0 blocks, 0 rollbacks.` message at the beginning.

plutus-chain-index/src/Plutus/ChainIndex/Events.hs
ak3n: Hmm, changed it to `error` as it's an impossible situation.

plutus-chain-index/src/Plutus/ChainIndex/SyncStats.hs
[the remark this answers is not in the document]
koslambrou: Keep this log message. However, you need to change the implementation of `processEventsQueue`. The reason it applied 0 blocks in the beginning is because you suppose that the first event is either Resume or RollBackwards, and the rest are RollForward. However, the two first events are actually "Resume" AND "RollBackwards", so there are 2 events to skip before the RollForward. Therefore, you need to make the implementation a bit more general.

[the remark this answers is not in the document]
ak3n: Yeah, I know, I turned it off to avoid changing stuff as it gets expensive to change logic just to support one logging line.

Will fix it.

koslambrou (approved):

locallycompact:
This is still using tons of memory if you leave it for long enough

```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
2724484 root      20   0 1032.7g  56.9g  19972 S  3036  90.7  10586:22 plutus-chain-in
```

ak3n:
@locallycompact yeah, this is because blocks become much bigger later but the queue size is constant. I'm working at the moment to improve it and make the memory usage constant during the whole chain.

locallycompact:
I'm also still getting the restart behaviour, I'm not sure if that fix was included here.

PR #349, opened 2022-03-08
Add deriving Data to SymValue, PaymentPubKey and PaymentPubKeyHash

locallycompact:


locallycompact:
I don't know how to update the sha256map in `nix/pkgs/haskell/haskell.nix` for this.

kk-hainq:
I would suggest finding and replacing all `4127e9cd6e889824d724c30eae55033cb50cbf3e` with `aa6c7bcb7b79357f4449c5bf62e8c0afec23b9d2` in the code base.

You can get the new hash with:
```
$ nix-shell -p nix-prefetch-git
$ nix-prefetch-git https://github.com/input-output-hk/plutus aa6c7bcb7b79357f4449c5bf62e8c0afec23b9d2
...
"sha256": "0903xqr1ql9h80kjwrwkjd6s1lgjbfal0gapq4iicv2gyhj265b5"
...
```

To then update `sha256map` in `nix/pkgs/haskell/haskell.nix`. After that, you would need another `updateMaterialized` as well:
```
$ nix-shell
$ updateMaterialized
```

We really should make it easier to encourage more contributors :smiling_face_with_tear:.

On another note, may I ask what motivated this addition?

locallycompact:
Data is needed for the new changes to the ContractModel to allow Action shrinking to work properly with Symbolic references, and I use PaymentPubKeyHash in my rest interface so I need it there as well to generate Actions.

locallycompact:
Not sure what this next CI failure is

koslambrou:
Given the way we currently handle dependencies, we only update `plutus` when we upgrade the `cardano-node` dependency. We don't want to use a `plutus` version different from the one `cardano-node` is using.

locallycompact:
These Data instances depend on the Plutus ones

koslambrou:
Then I'll close this. Once the new cardano-node version comes up, you can reopen.

PR #360, opened 2022-03-15
SCP-3331 Fixed error handling for transaction balancing in PAB.

bwbush:
The `Cardano.Wallet.LocalClient.runClient'` function erroneously rethrows errors it catches instead of returning them via `pure . Left`, as required by its type signature. The errors are not handled and end up killing the thread that handles the contract instance. This change successfully passes the Marlowe test https://github.com/input-output-hk/marlowe-cardano/blob/sprint-52/marlowe-cli/test/test-wallet-failure.yaml.

(In general, I recommend reviewing the PAB codebase for similar re-throwing problems and for `forkIO`s that aren't guarded against its thread dying because of an uncaught exception.)

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


sjoerdvisscher (approved):
Thanks, nice catch!

PR #382, opened 2022-03-30
[chain-index]: start a web server with forkProcess instead of withAsync

ak3n:
This PR replaces `withAsync` with `forkProcess` to start a web server in a different process to avoid thread blocking because `processEventsQueue` and `syncChainIndex` take too much resources.

Not sure if it a good solution but at the moment it seems the only one that works.

---

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-chain-index/plutus-chain-index.cabal
+++ b/plutus-chain-index/plutus-chain-index.cabal
@@ -64,6 +64,7 @@ library
         stm -any,
         time-units -any,
         yaml -any,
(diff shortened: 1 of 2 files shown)


sjoerdvisscher (approved):
Just needs an `updateMaterialized`.

ak3n:
The proper fix is #386.

PR #401, opened 2022-04-06
Allow chain-index to print silently.

zmrocze:
### description
Pr adds a new definition `Plutus.ChainIndex.App (runMainWithLog)` allowing to run chain-index with different function to do startup logs and refactors `runMain` to use this definition. This is a tiny commit.

### explanation
In [plutip](https://github.com/mlabs-haskell/plutip/issues/48) - a tool to spin up a testnet and run contracts - we call **chain-index**'s main function (from code with `runMain`) as one of the components. It does logging at startup to stdout and these logs get mixed up with the rest of the logs. Running it in a new process is troublesome, so I am asking if I could just add this small change to chain-index itself.

### DIFF
--- a/plutus-chain-index/src/Plutus/ChainIndex/App.hs
+++ b/plutus-chain-index/src/Plutus/ChainIndex/App.hs
@@ -72,7 +72,7 @@ main = do
 runMain :: CM.Configuration -> Config.ChainIndexConfig -> IO ()
 runMain = runMainWithLog putStrLn
 
--- Run main with provided function to log startup logs. 
+-- Run main with provided function to log startup logs.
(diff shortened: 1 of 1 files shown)


koslambrou:
Only the logs when starting the chain-index bother you and not the syncing logs?

zmrocze:
I missed the syncing logs. They bother me as well, but these are logged using proper logging machinery so I hope I can silence them with different `logConfig`. These originate from `processEventsQueue`. Failed to define right logConfig today.

zmrocze:
I fixed the formatting. Could we run workflows and possibly merge afterwards?

koslambrou (approved):

PR #403, opened 2022-04-07
Export full PABEnvironment

agocorona:
Export all the content of `PABEnvironment` so that other alternative monads  may orchestrate the execution of DApp applications and smart contracts and yet invoke PAB primitives for managing smart contracts etc. 

Example of usage in https://github.com/agocorona/DAppFlow/blob/main/pabtest.hs#L227-L238

For that purpose, the alternative monad need to store and set the entire PABEnvironment before invoking such primitives. For that purpose all the PABEnvironment should be reachable.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-pab/src/Plutus/PAB/Core.hs
+++ b/plutus-pab/src/Plutus/PAB/Core.hs
@@ -39,7 +39,7 @@ module Plutus.PAB.Core
     , EffectHandlers(..)
     , runPAB
     , runPAB'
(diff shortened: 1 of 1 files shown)


agocorona (commented):
Some field of PABEnvironment is exported twice and this is an error. I i will remove the duplicated export and resubmit

PR #419, opened 2022-04-14
Add AdjustUnbalancedTx effect

### LINKED ISSUE #414: Contract Model failing to validate with UtxoFailure OutputTooSmallUTxO
### Summary

Hi, I have the following transaction failing to validate in the contract model

```
[INFO] Slot 55: W[1]: Balancing an unbalanced transaction:
                        Tx:
                          Tx a5bf70436dc5f91bf8243c43c9b0d77498fc4253606664a91b5a580bc3bdfc6d:
                            {inputs:
                            collateral inputs:
                            outputs:
                              - Value (Map [(,Map [("",2000000)]),(5cecdeade418c8ed11ece4606e38e7c2639d0d60ebcfa259e4648038,Map [("hvmjukaikcbo",1)])]) addressed to
                                ScriptCredential: ea6610877e6636e1275c3ec7a677a14e8717e18ad5530a22735db27e (no staking credential)
                            mint: Value (Map [])
                            fee: Value (Map [])
                            mps:
                            signatures:
                            validity range: Interval {ivFrom = LowerBound NegInf True, ivTo = UpperBound PosInf True}
                            data:
                              <"">}
                        Requires signatures:
                        Utxo index:
                        Validity range:
                          (-∞ , +∞)
[INFO] Slot 55: W[1]: Finished balancing:
                        Tx 5bb168cfaa938e90412a41e931e77eca84b0f6b7ef8c9b532549f6c42674a53d:
                          {inputs:
                             - 8559a95e9b295035dfa107bdc914ded2924691b32fa8f00073ff17a156893dbc!1

                             - ba588cd0323242dcb13535132cc09d971e6db00f7259eb1faa4376e32187069e!0

                             - ba588cd0323242dcb13535132cc09d971e6db00f7259eb1faa4376e32187069e!1

                          collateral inputs:
                            - ba588cd0323242dcb13535132cc09d971e6db00f7259eb1faa4376e32187069e!0

                          outputs:
                            - Value (Map [(,Map [("",77663957)])]) addressed to
                              PubKeyCredential: a2c20c77887ace1cd986193e4e75babd8993cfd56995cd5cfce609c2 (no staking credential)
                            - Value (Map [(,Map [("",2000000)]),(5cecdeade418c8ed11ece4606e38e7c2639d0d60ebcfa259e4648038,Map [("",1),("an",1),("ootfzncmn",1),("rkuxine",1)]),(97e1179c05b16a76da408c686fc93dbe2c7c6a122d5a04f3745aa022,Map [("A",1000000),("B",1000000),("C",1000000),("D",1000000)]),(ee340377feb9820e7f5e79cf5421db776bce744a72a31266222eee4e,Map [("",1),(0x2c88,1),(0x95,1)])]) addressed to
                              PubKeyCredential: a2c20c77887ace1cd986193e4e75babd8993cfd56995cd5cfce609c2 (no staking credential)
                            - Value (Map [(,Map [("",2000000)]),(5cecdeade418c8ed11ece4606e38e7c2639d0d60ebcfa259e4648038,Map [("hvmjukaikcbo",1)])]) addressed to
                              ScriptCredential: ea6610877e6636e1275c3ec7a677a14e8717e18ad5530a22735db27e (no staking credential)
                          mint: Value (Map [])
                          fee: Value (Map [(,Map [("",958)])])
                          mps:
                          signatures:
                          validity range: Interval {ivFrom = LowerBound NegInf True, ivTo = UpperBound PosInf True}
                          data:
                            <"">}
[INFO] Slot 55: W[1]: Signing tx: 5bb168cfaa938e90412a41e931e77eca84b0f6b7ef8c9b532549f6c42674a53d
[INFO] Slot 55: W[1]: Submitting tx: 5bb168cfaa938e90412a41e931e77eca84b0f6b7ef8c9b532549f6c42674a53d
[INFO] Slot 55: W[1]: TxSubmit: 5bb168cfaa938e90412a41e931e77eca84b0f6b7ef8c9b532549f6c42674a53d
[WARNING] Slot 55: TxnValidationFail Phase1 5bb168cfaa938e90412a41e931e77eca84b0f6b7ef8c9b532549f6c42674a53d: CardanoLedgerValidationError "ApplyTxError [UtxowFailure (WrappedShelleyEraFailure (UtxoFailure (OutputTooSmallUTxO [(Addr Testnet (KeyHashObj (KeyHash \"a2c20c77887ace1cd986193e4e75babd8993cfd56995cd5cfce609c2\")) StakeRefNull,Value 2000000 (fromList [(PolicyID {policyID = ScriptHash \"5cecdeade418c8ed11ece4606e38e7c2639d0d60ebcfa259e4648038\"},fromList [(\"\",1),(\"an\",1),(\"ootfzncmn\",1),(\"rkuxine\",1)]),(PolicyID {policyID = ScriptHash \"97e1179c05b16a76da408c686fc93dbe2c7c6a122d5a04f3745aa022\"},fromList [(\"A\",1000000),(\"B\",1000000),(\"C\",1000000),(\"D\",1000000)]),(PolicyID {policyID = ScriptHash \"ee340377feb9820e7f5e79cf5421db776bce744a72a31266222eee4e\"},fromList [(\"\",1),(\",\\136\",1),(\"\\149\",1)])]),SNothing)])))]"
```

Am I right in assuming there's too much locked value from the amount of different tokens in this UTxO that 2_000_000 lovelace isn't enough ADA for that UTxO? Other transactions validate fine.

Sorry not public code but I can put together a reproduction if needed.

### Actual Result

See above

### Expected Result

See above

_No response_

### System info

NixOS

ak3n:
Adds `AdjustUnbalancedTx` `PABReq` effect that uses protocol parameters.

Should fix #414.

---

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-contract/src/Wallet/API.hs
```diff
@@ -91,7 +91,7 @@ payToPaymentPublicKeyHash range v pk = do
     utx <- either (throwError . PaymentMkTxError)
                   pure
                   (Constraints.mkTx @Void mempty constraints)
-    let adjustedUtx = Constraints.adjustUnbalancedTx utx
+    let adjustedUtx = Constraints.adjustUnbalancedTx minAdaTxOut utx
```
ak3n, line 94: The only place where `adjustUnbalancedTx` from `Constraints` being used. It's hard to import `adjustUnbalancedTx` from `Contract`, there are circular dependency modules.

plutus-pab/src/Plutus/PAB/Core/ContractInstance/BlockchainEnv.hs
```diff
@@ -48,23 +54,23 @@ import System.Random
...
   -> IO BlockchainEnv
-startNodeClient socket mode rollbackHistory slotConfig networkId resumePoint instancesState = do
-    env <- STM.atomically $ emptyBlockchainEnv rollbackHistory slotConfig
-    case mode of
+startNodeClient config params instancesState = do
+    let Config { nodeServerConfig = PABServerConfig{pscSocketPath, pscSlotConfig, pscNodeMode, pscNetworkId = NetworkIdWrapper networkId}
+               , developmentOptions = DevelopmentOptions{pabRollbackHistory, pabResumeFrom} } = config
+    let ProtocolParameters{protocolParamUTxOCostPerWord = mCostPerWord } = params
+    let coinsPerUTxOWord = fromMaybe (fromJust $ protocolParamUTxOCostPerWord def) mCostPerWord
+    env <- STM.atomically $ emptyBlockchainEnv pabRollbackHistory pscSlotConfig coinsPerUTxOWord
```
ak3n, line 66: I decided to pass `Config` and `ProtocolParameters` directly. `startNodeClient` depends on too many params to use arguments here to my taste.

plutus-contract/src/Wallet/API.hs
sjoerdvisscher: Makes sense 👍

sjoerdvisscher:
`coinsPerUTxOWord` is not a drop-in replacement for `minAdaTxOut`. See https://github.com/input-output-hk/cardano-ledger/blob/master/doc/explanations/min-utxo-alonzo.rst for details. I'm still not sure what the best way is to fix #414.

ak3n:
@sjoerdvisscher hmm, well, there are


- [`evaluateMinLovelaceOutput`](https://github.com/input-output-hk/cardano-ledger/blob/9c3b4737b13b30f71529e76c5330f403165e28a6/eras/shelley-ma/impl/src/Cardano/Ledger/Mary.hs#L87)

- [`utxoEntrySize`](https://github.com/input-output-hk/cardano-ledger/blob/70dbfc9f12283666e29f8492aff5da9b1c8c5a7f/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxo.hs#L106) (from documentation)

- [`calculateMinimumUTxO`](https://github.com/input-output-hk/cardano-node/blob/9d6cbe47997f6aaf6e80ceb5663e0297474aad7e/cardano-api/src/Cardano/Api/Fees.hs#L1022a)

that uses `utxoEntrySize`, and is used by `checkMinUTxOValue` in [`makeTransactionBodyAutoBalance`](https://github.com/input-output-hk/cardano-node/blob/9d6cbe47997f6aaf6e80ceb5663e0297474aad7e/cardano-api/src/Cardano/Api/Fees.hs#L851).

I suppose we should use it but I suspect we don't already do it because we need to convert the transaction from our representation to the cardano-node's types?

sjoerdvisscher:
We even have `Ledger.Validation.evaluateMinLovelaceOutput` and `Ledger.Validation.fromPlutusTxOut` which together can easily calculate the correct amount. But we suspect it will be really hard to fix all the testcases. But maybe you want to try it out?

plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
```diff
@@ -392,14 +392,12 @@ mkTx lookups txc = mkSomeTx [SomeLookupsAndConstraints lookups txc]
 
 -- | Each transaction output should contain a minimum amount of Ada (this is a
 -- restriction on the real Cardano network).
---
--- TODO: In the future, the minimum Ada value should be configurable.
-adjustUnbalancedTx :: UnbalancedTx -> UnbalancedTx
-adjustUnbalancedTx = over (tx . Tx.outputs . traverse) adjustTxOut
+adjustUnbalancedTx :: Ledger.Ada -> UnbalancedTx -> UnbalancedTx
+adjustUnbalancedTx minAdaTxOut = over (tx . Tx.outputs . traverse) adjustTxOut
```
koslambrou, line 396: The parameter should be `coinsPerUtxoWord` instead of `minAdaTxOut`, no? That's what you're passing in through the handlers. In that case, this implementation should be different.

koslambrou (changes requested):

plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
```diff
@@ -396,14 +398,13 @@ mkTx lookups txc = mkSomeTx [SomeLookupsAndConstraints lookups txc]
...
--- TODO: In the future, the minimum Ada value should be configurable.
-adjustUnbalancedTx :: UnbalancedTx -> UnbalancedTx
-adjustUnbalancedTx = over (tx . Tx.outputs . traverse) adjustTxOut
+adjustUnbalancedTx :: ProtocolParameters -> UnbalancedTx -> UnbalancedTx
+adjustUnbalancedTx pparams = over (tx . Tx.outputs . traverse) adjustTxOut
   where
     adjustTxOut :: TxOut -> TxOut
     adjustTxOut txOut =
-      let missingLovelace = max 0 (Ledger.minAdaTxOut - Ada.fromValue (txOutValue txOut))
+      let minAdaTxOut' = either (const minAdaTxOut) (Ada.fromValue . evaluateMinLovelaceOutput pparams) $ fromPlutusTxOut' txOut in
```
koslambrou, line 406: Not sure here, but feels like we shouldn't default to `minAdaTxOut` because of a conversion error between Plutus TxOut and Cardano TxOut. Someone who calls `adjustUnbalancedTx` would expect it to always work because of the signature. Maybe return `Either ... UnbalancedTx`?

plutus-contract/src/Wallet/API.hs
```diff
@@ -91,12 +95,20 @@ payToPaymentPublicKeyHash range v pk = do
...
     balancedTx <- balanceTx adjustedUtx
     either throwError signTxAndSubmit balancedTx
+    where
+        adjustUnbalancedTx :: Constraints.UnbalancedTx -> Constraints.UnbalancedTx
+        adjustUnbalancedTx = over (tx . Tx.outputs . traverse) adjustTxOut
+          where
+            adjustTxOut :: Tx.TxOut -> Tx.TxOut
+            adjustTxOut txOut =
+              let missingLovelace = max 0 (minAdaTxOut - Ada.fromValue (Tx.txOutValue txOut))
+               in txOut { Tx.txOutValue = Tx.txOutValue txOut <> Ada.toValue missingLovelace }
```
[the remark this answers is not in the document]
ak3n: Good question.

In general `minAda` is not enough but it seems that `payToPaymentPublicKeyHash` is being used only in tests, use cases, and on PAB's start.

We can bring protocol params as dependency here but it will require us to fix all tests to pass it. Or we can use the default protocol parameters as in the emulator handler.

plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
ak3n: Agree, I choose `minAdaTxOut` as a default behaviour. Returning `Either` sounds better.

plutus-contract/src/Wallet/API.hs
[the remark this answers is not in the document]
koslambrou: Since the PAB is actually using it through `payToPaymentPublicKey` in `Core`, I would bring the protocol params as dependency and update the tests. Maybe in another PR? Up to you though.

sjoerdvisscher:
@ak3n What's the status of this?

plutus-contract/src/Wallet/Emulator/Folds.hs
```diff
@@ -120,6 +121,10 @@ scriptEvents = preMapMaybe (preview (eteEvent . chainEvent) >=> getEvent) (conca
 walletTxBalanceEvents :: EmulatorEventFold [UnbalancedTx]
 walletTxBalanceEvents = preMapMaybe (preview (eteEvent . walletEvent' . _2 . _TxBalanceLog . _BalancingUnbalancedTx)) L.list
 
+-- | Min lovelace of 'txOut's from adjusted unbalanced transactions
+walletAdjustedTxEvents :: EmulatorEventFold [(Wallet, [Value])]
```
koslambrou, line 125: I'm guessing this includes ALL transaction outputs of transaction's for which `adjustUnbalancedTx` was called on? Even the outputs which didn't need adjusting are listed?

plutus-contract/src/Plutus/Contract/Test/ContractModel/MissingLovelace.hs
```diff
@@ -0,0 +1,30 @@
...
+  ( calculateDelta
+  ) where
+
+import Data.List as List
+import Ledger.Ada qualified as Ada
+import Ledger.Value (Value, noAdaValue)
+import PlutusTx.Prelude qualified as P
+import Wallet.Emulator (Wallet)
+
+calculateDelta :: Value -> Value -> Value -> Wallet -> [(Wallet, [Value])] -> Value
```
koslambrou, line 11: Missing haddock comments on the function and on the parameters. What's the difference between each of those `Value`s?

plutus-contract/src/Plutus/Contract/Test/ContractModel/MissingLovelace.hs
```diff
@@ -0,0 +1,30 @@
+module Plutus.Contract.Test.ContractModel.MissingLovelace
```
koslambrou, line 1: Why put it as a submodule of `ContractModel`?

plutus-use-cases/test/Spec/Prism.hs
```diff
@@ -165,11 +165,11 @@ instance ContractModel PrismModel where
         case cmd of
             Revoke w  -> do
               issued <- use (isIssued w)
-              when (issued == Issued) $ deposit mirror minAdaTxOut
+              when (issued == Issued) $ deposit mirror (Ada.lovelaceOf 0)
               isIssued w %= doRevoke
             Issue w   -> do
               wait 1
-              withdraw mirror minAdaTxOut
+              withdrawMinted mirror $ someTokenValue "kyc" 1
```
koslambrou, line 172: In curious here, what is this supposed to do?

plutus-contract/src/Plutus/Contract/Test/ContractModel/MissingLovelace.hs
```diff
@@ -0,0 +1,30 @@
...
+  ) where
+
+import Data.List as List
+import Ledger.Ada qualified as Ada
+import Ledger.Value (Value, noAdaValue)
+import PlutusTx.Prelude qualified as P
+import Wallet.Emulator (Wallet)
+
+calculateDelta :: Value -> Value -> Value -> Wallet -> [(Wallet, [Value])] -> Value
+calculateDelta initialDelta initialValue finalValue w txOutCosts =
```
koslambrou, line 12: Add some comments on the different step in order to better understand what is going on?

plutus-pab/src/Plutus/PAB/Run/Cli.hs
```diff
@@ -110,12 +109,14 @@ runConfigCommand _ ConfigCommandArgs{ccaTrace, ccaPABConfig=Config{dbConfig}} Mi
     App.migrate (toPABMsg ccaTrace) dbConfig
 
 -- Run mock wallet service
-runConfigCommand _ ConfigCommandArgs{ccaTrace, ccaPABConfig = Config {nodeServerConfig, chainIndexConfig, walletServerConfig = LocalWalletConfig ws},ccaAvailability} MockWallet =
+runConfigCommand _ ConfigCommandArgs{ccaTrace, ccaPABConfig = Config {nodeServerConfig, chainIndexConfig, walletServerConfig = LocalWalletConfig ws},ccaAvailability} MockWallet = do
+    params <- liftIO $ Params.fromPABServerConfig nodeServerConfig
+    -- = Params { pSlotConfig = pscSlotConfig nodeServerConfig, pProtocolParams = protocolParameters, pNetworkId = networkId }
```
koslambrou, line 114: Remove comment?

plutus-contract/src/Wallet/API.hs
```diff
@@ -104,9 +109,10 @@ payToPaymentPublicKeyHash_ ::
     ( Member WalletEffect effs
     , Member (Error WalletAPIError) effs
     , Member (LogMsg Text) effs
+    , Member (LogMsg RequestHandlerLogMsg) effs
     )
     => SlotRange -> Value -> PaymentPubKeyHash -> Eff effs ()
-payToPaymentPublicKeyHash_ r v = void . payToPaymentPublicKeyHash r v
+payToPaymentPublicKeyHash_ r v = void . payToPaymentPublicKeyHash def r v
```
koslambrou, line 115: Mmm... not sure here about using `def`. This is not an emulator function, this can actually be used in the `Contract` monad, no?

plutus-ledger-constraints/plutus-ledger-constraints.cabal
```diff
@@ -44,6 +44,7 @@ library
     build-depends:
         base >=4.9 && <5,
         aeson -any,
+        cardano-api -any,
```
koslambrou, line 47: We don't actually need `cardano-api`, do we?

plutus-use-cases/test/Spec/GameStateMachine.hs
```diff
@@ -280,7 +278,7 @@ noLockedFunds = do
     w      <- forAllQ $ elementsQ wallets
     secret <- viewContractState currentSecret
     val    <- viewContractState gameValue
-    when (val > Ada.getLovelace Ledger.minAdaTxOut) $ do
+    when (val >= Ada.getLovelace Ledger.minAdaTxOut) $ do
```
koslambrou, line 281: I'm guessing we can't remove `minAdaTxOut` from here?

plutus-use-cases/test/Spec/GameStateMachine.hs
```diff
@@ -318,27 +316,27 @@ tests :: TestTree
...
-        .&&. walletFundsChange w3 (Ada.toValue Ledger.minAdaTxOut <> Ada.adaValueOf 5 <> guessTokenVal))
+        .&&. walletFundsChange w1 (Ada.adaValueOf (-8))
+        .&&. walletFundsChange w3 (Ada.adaValueOf 5 <> guessTokenVal))
         successTrace2
 
     , checkPredicateOptions options "run a successful game trace where we try to leave 1 Ada in the script address"
-        (walletFundsChange w1 (Ada.toValue (-Ledger.minAdaTxOut) <> guessTokenVal)
-        .&&. valueAtAddress (Scripts.validatorAddress $ G.typedValidator gameParam) (Ada.toValue Ledger.minAdaTxOut ==))
+        (walletFundsChange w1 (guessTokenVal)
+        .&&. valueAtAddress (Scripts.validatorAddress $ G.typedValidator gameParam) (Ada.toValue 1344798 ==))
```
koslambrou, line 333: What's this magic number?

plutus-ledger/src/Ledger/Validation.hs
```diff
@@ -317,10 +323,26 @@ fromPlutusTxId = fmap toShelleyTxId . P.toCardanoTxId
...
+
 fromPaymentPrivateKey :: P.PrivateKey -> TxBody EmulatorEra -> C.Api.KeyWitness C.Api.AlonzoEra
 fromPaymentPrivateKey xprv txBody
   = C.Api.makeShelleyKeyWitness
       (C.Api.ShelleyTxBody C.Api.ShelleyBasedEraAlonzo txBody notUsed notUsed notUsed notUsed)
       (C.Api.WitnessPaymentExtendedKey (C.Api.PaymentExtendedSigningKey xprv))
   where
     notUsed = undefined -- hack so we can reuse code from cardano-api
+
+getCardanoTxOutputsMissingCosts :: P.Params -> P.CardanoTx -> [P.Value]
```
koslambrou, line 339: Haddock? What does this do?

plutus-ledger/src/Ledger/Index.hs
```diff
@@ -397,11 +350,15 @@ minAdaTxOut = Ada.lovelaceOf 2_000_000
 -- slightly more expensive than 1 Ada. So, to be on the safe side, we set the
 -- minimum Ada of each transaction output to 2 Ada.
 checkMinAdaInTxOutputs :: ValidationMonad m => Tx -> m ()
```
koslambrou, line 352: Rework the old haddock comment.

plutus-ledger/src/Ledger/Index.hs
```diff
@@ -397,11 +350,15 @@ minAdaTxOut = Ada.lovelaceOf 2_000_000
...
-checkMinAdaInTxOutputs t@Tx { txOutputs } =
-    for_ txOutputs $ \txOut ->
-        if Ada.fromValue (txOutValue txOut) >= minAdaTxOut
+checkMinAdaInTxOutputs t@Tx { txOutputs } = do
+    params <- vctxParams <$> ask
+    for_ txOutputs $ \txOut -> do
+        let
+            minAdaTxOut' = either (const minAdaTxOut) id $
```
koslambrou, line 357: Mmm... I'm wondering whether we should keep this `minAdaTxOut`. What do you think @sjoerdvisscher ?

koslambrou:
@ak3n Very good work by the way. Didn't seem easy to update all of those plutus-use-cases :D

plutus-contract/src/Wallet/API.hs
ak3n: The only user of it in plutus-apps is `Spec.Emulator` module so I thought that it's internal. But yeah, we can introduce an argument here and pass `def` at usage points. Should I?

plutus-use-cases/test/Spec/GameStateMachine.hs
ak3n: It's a calculated cost by `evaluateMinLovelaceOutput` that is necessary for a `txOut` with 1 token.

As long as we don't use a test framework (`ContractModel`) for this test where we can change wallets balances we have to check values.

We can introduce a variable with a name and a comment that will cover that.

plutus-ledger-constraints/plutus-ledger-constraints.cabal
ak3n: Yep, removed, thanks!

plutus-ledger/src/Ledger/Index.hs
ak3n: I have started in that direction, to remove it everywhere, but it seems to be a next step as it breaks more tests and introduces more changes in the source tree (the PR is big already).

But yes, I would like to hear if it's a valid direction. I assume that we don't need it anymore.

plutus-use-cases/test/Spec/Prism.hs
ak3n: As `ContractModel` layer introduces it's own state above contracts we have to make them converge.

In `ContractModel` it's possible to set the initial distribution of wallets and add tokens for example. But in this case, in the underlying contract, we issue a token and then withdraw it at the same time. To trigger that in `ContractModel` I added `withdrawMinted` that deposits that token and then immediately withdraws.

plutus-contract/src/Wallet/Emulator/Folds.hs
ak3n: Yep, that's correct. At the beginning I even tried to use it — to understand what transactions were adjusted to calculate wallets funds. But turned out it's too hard and requires more info.

plutus-ledger/src/Ledger/Index.hs
```diff
@@ -397,11 +350,15 @@ minAdaTxOut = Ada.lovelaceOf 2_000_000
...
+                fromPlutusTxOutUnsafe params txOut <&> \txOut' -> Ada.fromValue $ evaluateMinLovelaceOutput params txOut'
```
sjoerdvisscher, line 358: It seems we call `Ada.fromValue` after every `evaluateMinLovelaceOutput` call. Better fix `evaluateMinLovelaceOutput` to return `Ada`.

plutus-ledger/src/Ledger/Index.hs
sjoerdvisscher: Yes, `minAdaTxOut` should no longer exist. It's fine to do that in a next PR, as long as you are confident we will be able to fix all tests.

plutus-pab/src/Cardano/Node/Params.hs
```diff
@@ -0,0 +1,30 @@
+{-# LANGUAGE NamedFieldPuns #-}
+module Cardano.Node.Params where
```
sjoerdvisscher, line 2: Nice, this is much better!

Did you need this change, or is it just something you noticed?

plutus-pab/src/Plutus/PAB/Simulator.hs
```diff
@@ -751,7 +752,7 @@ addWalletWith funds = do
         STM.writeTVar _agentStates newWallets
     _ <- handleAgentThread (knownWallet 2) Nothing
             $ Modify.wrapError WalletError
-            $ MockWallet.distributeNewWalletFunds funds (CW.paymentPubKeyHash mockWallet)
+            $ MockWallet.distributeNewWalletFunds def funds (CW.paymentPubKeyHash mockWallet)
```
sjoerdvisscher, line 755: Since you added Params to BlockchainEnv, you can get it from the PABEnvironment.

plutus-pab/src/Plutus/PAB/Simulator.hs
```diff
@@ -782,8 +783,8 @@ payToWallet :: forall t. Wallet -> Wallet -> Value -> Simulation t CardanoTx
 payToWallet source target = payToPaymentPublicKeyHash source (Emulator.mockWalletPaymentPubKeyHash target)
 
 -- | Make a payment from one wallet to a public key address
-payToPaymentPublicKeyHash :: forall t. Wallet -> PaymentPubKeyHash -> Value -> Simulation t CardanoTx
+payToPaymentPublicKeyHash :: forall t.  Wallet -> PaymentPubKeyHash -> Value -> Simulation t CardanoTx
 payToPaymentPublicKeyHash source target amount =
     handleAgentThread source Nothing
         $ flip (handleError @WAPI.WalletAPIError) (throwError . WalletError)
-        $ WAPI.payToPaymentPublicKeyHash WAPI.defaultSlotRange amount target
+        $ WAPI.payToPaymentPublicKeyHash def WAPI.defaultSlotRange amount target
```
sjoerdvisscher, line 790: Same here.

plutus-use-cases/test/Spec/GameStateMachine.hs
sjoerdvisscher: Or we change the testcase to try to leave 2 Ada, so this can just be `2_000_000`?

sjoerdvisscher (commented):
This looks really good!

Some first pass comments. I want to go through it once more.

MaximilianAlgehed:
Can I chime in and kindly ask that before you merge this you run the `ContractModel` tests a few hundred times instead of the 20 or so that's in CI to make sure nothing derps out there? :)

plutus-contract/src/Plutus/Contract/Test/ContractModel/MissingLovelace.hs
sjoerdvisscher: Indeed, this is really hard to follow.

sjoerdvisscher (changes requested):

plutus-contract/src/Plutus/Contract/Test.hs
```diff
@@ -572,11 +573,12 @@ walletFundsExactChange :: Wallet -> Value -> TracePredicate
 walletFundsExactChange = walletFundsChangeImpl True
 
 walletFundsChangeImpl :: Bool -> Wallet -> Value -> TracePredicate
-walletFundsChangeImpl exact w dlt = TracePredicate $
-    flip postMapM (L.generalize $ (,) <$> Folds.walletFunds w <*> Folds.walletFees w) $ \(finalValue', fees) -> do
+walletFundsChangeImpl exact w dlt' = TracePredicate $
+    flip postMapM (L.generalize $ (,,) <$> Folds.walletFunds w <*> Folds.walletFees w <*> Folds.walletAdjustedTxEvents) $ \(finalValue', fees, txOutCosts) -> do
```
sjoerdvisscher, line 577: For symmetry it would be better to let `Folds.walletAdjustedTxEvents` also only return the events for a given wallet. Then you don't have to filter in `calculateDelta`.

plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
```diff
@@ -401,14 +401,20 @@ mkTx lookups txc = mkSomeTx [SomeLookupsAndConstraints lookups txc]
...
+adjustUnbalancedTx params utx =
+    let (acc, res) = mapAccumLOf (tx . Tx.outputs . traverse) step (Right []) utx
+    in (flip (,) res) <$> acc
   where
-    adjustTxOut :: TxOut -> Either Tx.ToCardanoError TxOut
+    step acc txOut = case (acc, adjustTxOut txOut) of
+        (Left _, _)                       -> (acc, txOut) -- acc is an error, do nothing
+        (Right _, Left e)                 -> (Left e, txOut) -- failed somewhere, return error as acc
+        (Right acc', Right (ada, txOut')) -> (Right $ ada:acc', txOut')
+    adjustTxOut :: TxOut -> Either Tx.ToCardanoError (Ada.Ada, TxOut)
```
sjoerdvisscher, line 413: If you let this return `Either Tx.ToCardanoError ([Ada.Ada], TxOut)` you have 2 nested applicative effects (`Either Tx.ToCardanoError` and `(,) [Ada.Ada]`), which you could use in a traversal using `Compose`. Then you won't need `step`.

plutus-contract/src/Plutus/Contract/Test/ContractModel/MissingLovelace.hs
```diff
@@ -1,21 +1,20 @@
...
+calculateDelta :: Value -> Value -> Value -> Wallet -> [(Wallet, [Ada.Ada])] -> Value
+calculateDelta initialDelta initialValue finalValue w (Set.toList . Set.fromList -> txOutCosts) =
   let
     deltas =
-      let txOutCosts' = case List.uncons $ reverse $ filter ((== w) . fst) txOutCosts of
-            Just ((_, vs), _) -> map Ada.fromValue vs
-            _                 -> []
-          otherWalletsTxOutCosts = concatMap (map Ada.fromValue . snd) $ txOutCosts
+      let txOutCosts' = concatMap snd $ filter ((== w) . fst) txOutCosts
+          otherWalletsTxOutCosts = concatMap snd txOutCosts
```
sjoerdvisscher, line 17: Isn't this `allWalletsTxOutCosts`?

plutus-pab/src/Plutus/PAB/Simulator.hs
ak3n: Nice, thanks! I fixed `def` in some places but missed it here. I think we should be more careful with `def`.

plutus-contract/src/Plutus/Contract/Test/ContractModel/MissingLovelace.hs
ak3n: I decided to move this function to its own module because `ContractModel.Internal` is already almost 2k lines. Also because I wanted to write some long comment to it and a separate module might emphasize that this function is special in some sense. But I don't mind can move it to `Internal`.

plutus-use-cases/test/Spec/GameStateMachine.hs
ak3n: I think to do it later as a next step that removes `minAdaTxOut`.

plutus-contract/src/Plutus/Contract/Test.hs
ak3n: As we found out `walletEvent` and `walletEvent'` return a bit different results. I renamed `walletAdjustedTxEvents` to `walletsAdjustedTxEvents` and added `walletAdjustedTxEvents :: Wallet -> ...`.

plutus-contract/src/Plutus/Contract/Test/ContractModel/MissingLovelace.hs
ak3n: Yeah, indeed, thanks for pointing this out. I will refactor and write a comment a bit later.

plutus-pab/src/Cardano/Node/Params.hs
ak3n: I noticed several `readPP` functions and decided to make a proper one.

plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
```diff
@@ -692,6 +695,14 @@ deposit w val = modState (balanceChangesL . at w) (Just . maybe (toSymValue val)
 withdraw :: SymValueLike v => Wallet -> v -> Spec state ()
 withdraw w val = deposit w (inv . toSymValue $ val)
 
+-- | Withdraw minted tokens from a wallet. This works for situations when we mint tokens and withdraw them
+-- in one step. We have to deposit them at first to the wallet and then withdraw
+-- to make sure that our adjusted ada will get into 'delta' to check balances properly.
+withdrawMinted :: SymValueLike v => Wallet -> v -> Spec state ()
```
MaximilianAlgehed, line 701: Isn't this function just `withdrawMinted = \ w _ -> deposit w mempty`?

plutus-contract/src/Plutus/Contract/Test.hs
sjoerdvisscher: I don't think this is the right conclusion to get from this. `walletEvent` and `walletEvent'` differ only in filtering out events for the given wallet. (look at the implementation, it's really simple) The issue lay in how you used it.

When I suggested the above I didn't realise you also needed the events from all the wallets. So instead of having 2 functions it should also be possible to fix the implementation of `walletsAdjustedTxEvents`.

ak3n: >The issue lay in how you used it. 

I agree, my problem is here I don't understand yet what's the issue. `walletAdjustedTxEvents` and `walletsAdjustedTxEvents` looks very similar and I don't get why they have a different behaviour.

plutus-use-cases/test/Spec/SealedBidAuction.hs
```diff
@@ -249,5 +247,5 @@ tests :: TestTree
 tests =
   testGroup "sealed bid auction"
     [ testProperty "packInteger is injective" $ \x y -> x /= y ==> packInteger x /= packInteger y
-    , testProperty "prop_AuctionModelCorrect" $ withMaxSuccess 20 prop_Auction
+    , testProperty "prop_AuctionModelCorrect" $ withMaxSuccess 100 prop_Auction
```
sjoerdvisscher, line 250: This should have been a temporary change I think?

plutus-use-cases/test/Spec/Prism.hs
```diff
@@ -165,11 +164,10 @@ instance ContractModel PrismModel where
         case cmd of
             Revoke w  -> do
               issued <- use (isIssued w)
-              when (issued == Issued) $ deposit mirror minAdaTxOut
+              when (issued == Issued) $ deposit mirror (Ada.lovelaceOf 0)
```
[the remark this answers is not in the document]
ak3n: Yep, checked and deleted it, thanks!

plutus-use-cases/test/Spec/SealedBidAuction.hs
ak3n: Got it back, sorry :D

plutus-contract/src/Plutus/Contract/Test/ContractModel/MissingLovelace.hs
koslambrou: Sorry, I was unclear. What I meant was: Why as a submodule of `ContractModel` specifically? `calculateDelta` can be used outside of ContractModel, no? Maybe put it `Plutus.Contract.Test.MissingLovlace`? I could be wrong...

plutus-contract/src/Plutus/Contract/Test/ContractModel/MissingLovelace.hs
```diff
@@ -0,0 +1,63 @@
...
+  ( calculateDelta
+  ) where
+
+import Data.Set qualified as Set
+import Ledger.Ada qualified as Ada
+import Ledger.Value (Value, noAdaValue)
+import PlutusTx.Prelude qualified as P
+import Wallet.Emulator (Wallet)
+
+-- | Returns the calculated delta between initial and final values. Might be false positive.
```
koslambrou, line 12: What do you mean by "false positive"?

plutus-contract/src/Plutus/Contract/Test/ContractModel/MissingLovelace.hs
```diff
@@ -0,0 +1,63 @@
...
+--
+-- The tests check if a wallet's funds are equal to some expected value at the end.
+-- Unfortunately because of the adjustion of transactions, the outputs' costs change
```
koslambrou, line 15: Change to: "Unfortunately, because of transaction adjusting, the [...]"

plutus-contract/src/Plutus/Contract/Test/ContractModel/MissingLovelace.hs
ak3n: I think there is a risk when expected delta has only ada part and `expected delta /= realDelta` and `realDelta` is divisible by some delta from deltas, then we will return `realDelta`'s ada. Which means that the test will pass but without strong confidence in wallets' funds consistency.

For example, we expected `-n`, but there is `n` among deltas and `realDelta` is `n`, it is divisible by `n` => our test will pass.

plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
sjoerdvisscher: This worked even better than I thought, trying it out in vscode I got:
```haskell
adjustUnbalancedTx params = alaf Compose (tx . Tx.outputs . traverse) adjustTxOut
```

ak3n: Oh, indeed, looks cool and works fine! Thanks!

But lacks some readability as non-trivial lens code as usual.

sjoerdvisscher:
Looking at what the problem could be with

```haskell
walletsAdjustedTxEvents = preMapMaybe (preview (eteEvent . walletEvent' . to (\x -> (x ^. _1, x ^. _2 . _RequestHandlerLog . _AdjustingUnbalancedTx)))) L.list
```

One thing I notice is `_2` is a lens and `_RequestHandlerLog . _AdjustingUnbalancedTx)` is a prism, so composed they make a traversal. When applied to `^.` a traversal needs a monoid instance on the result, which is `[Ada.ada]` here. So one difference with `walletAdjustedTxEvents w` is that for each wallet event that does't match `_RequestHandlerLog . _AdjustingUnbalancedTx` you get a result with an empty list. I still don't see how this could cause missing data though.

I do think that with the current implementation you don't need the wallet in the return value, since you throw them away in `allWalletsTxOutCosts`, so a safer implementation could be 

```haskell
walletsAdjustedTxEvents = preMapMaybe (preview (eteEvent . walletEvent' . _2 . _RequestHandlerLog . _AdjustingUnbalancedTx)) L.list
```

plutus-contract/src/Plutus/Contract/Test/ContractModel/MissingLovelace.hs
koslambrou: I think this explanation should be added as a comment on the function.

ak3n: Added to the comment. 👍

sjoerdvisscher (approved):

koslambrou (approved):

koslambrou:
@ak3n You can merge whenever you think it's done :)

PR #421, opened 2022-04-15
Export full PABEnvironment

agocorona:
Export all the content of PABEnvironment so that other alternative monads may orchestrate the execution of DApp applications and smart contracts and yet make use of the native monad to invoke PAB primitives for managing smart contracts etc.

Example of usage in https://github.com/agocorona/DAppFlow/blob/main/pabtest.hs#L227-L238

For that purpose, the alternative monad need to store and set the entire PABEnvironment before invoking such primitives. so all the PABEnvironment should be reachable externally to get/set his value

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-pab/src/Plutus/PAB/Core.hs
+++ b/plutus-pab/src/Plutus/PAB/Core.hs
@@ -39,7 +39,7 @@ module Plutus.PAB.Core
     , EffectHandlers(..)
     , runPAB
     , runPAB'
-    , PABEnvironment(appEnv)
+    , PABEnvironment(..)
     -- * Contracts and instances
     , reportContractState
     , activateContract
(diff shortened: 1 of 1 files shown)


ak3n (approved):

plutus-pab/src/Plutus/PAB/Core.hs
```diff
@@ -67,7 +67,6 @@ module Plutus.PAB.Core
     , activeContracts
     , finalResult
     , waitUntilFinished
-    , blockchainEnv
```
koslambrou, line 70: Why remove this one?

ak3n: It's `PABEnvironment`'s field that is being exported as `PABEnvironment(..)`.

sjoerdvisscher (approved):

ak3n:
Thanks @agocorona!

PR #437, opened 2022-04-28
Have plutus-ledger not depend on cardano-wallet-core

ch1bo:
Avoid depending on cardano-wallet-core in `plutus-ledger` to make downstream consumption of only that package easier.

This would help at least us over at https://github.com/input-output-hk/hydra-poc as we are not using anything else from `plutus-apps` right now.

Maybe this does not make sense as you will be adding more wallet-related things to `plutus-ledger`.

We may be able to refactor / vendor out the relevant usage of `plutus-ledger` alltogether.


ch1bo:
See this commit for an example of what dependencies we can avoid downstream: https://github.com/input-output-hk/hydra-poc/commit/9bb2c5249f957d16d4d9c4800f4ceba3cb74ce38

koslambrou:
LGTM, thanks @ch1bo :)

PR #455, opened 2022-05-11
SCP-3631: Make protocolParameters configurable

sjoerdvisscher:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-contract/src/Plutus/Contract/Test.hs
```diff
@@ -184,6 +186,13 @@ defaultCheckOptions =
 changeInitialWalletValue :: Wallet -> (Value -> Value) -> CheckOptions -> CheckOptions
 changeInitialWalletValue wallet = over (emulatorConfig . initialChainState . _Left . ix wallet)
 
+allowBigTransactions :: CheckOptions -> CheckOptions
```
koslambrou, line 189: Could add a comment explaining in which scenario the user might think of using this option? Like, what's the specific error one might get if the transaction is too big.

plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
```diff
@@ -1497,6 +1497,7 @@ propRunActions_ actions =
 defaultCheckOptionsContractModel :: CheckOptions
 defaultCheckOptionsContractModel =
   defaultCheckOptions & emulatorConfig . initialChainState .~ (Left . Map.fromList $ zip knownWallets (repeat (Ada.lovelaceValueOf 100_000_000_000_000_000)))
+                      & allowBigTransactions
```
koslambrou, line 1500: Should we specify this somewhere (haddock comment or a note) that the tests using the Contract Model might be accurate when submitting to real Cardano network because of big transactions? Not sure where though.. What do you think?

plutus-ledger/src/Ledger/Params.hs
```diff
@@ -0,0 +1,65 @@
...
+makeLensesFor
+  [ ("pSlotConfig", "slotConfigL")
+  , ("pProtocolParams", "protocolParamsL")
+  , ("pNetworkId", "networkIdL") ]
+  ''Params
+
+instance Default Params where
+  def = Params def def (Testnet $ NetworkMagic 1)
+
+instance Default ProtocolParameters where
```
koslambrou, line 37: Specify where these default values come from? Is this Alonzo era protocol parameters?

plutus-ledger/src/Ledger/Params.hs
```diff
@@ -0,0 +1,65 @@
...
+    def = ProtocolParameters
+        { protocolParamProtocolVersion = (6,0)
```
koslambrou, line 39: mmm... I thought Alonzo era protocol version was 5..

plutus-ledger/src/Ledger/TimeSlot.hs
```diff
@@ -138,8 +142,24 @@ currentSlot :: SlotConfig -> IO Slot
...
+  . Haskell.truncate
+  . (Haskell.* 1000) -- Convert to ms
+  . Time.nominalDiffTimeToSeconds
+
+posixTimeToNominalDiffTime :: POSIXTime -> Time.NominalDiffTime
+posixTimeToNominalDiffTime
+  = Time.secondsToNominalDiffTime
+  . (Haskell./ 1000)
+  . Haskell.fromInteger
+  . getPOSIXTime
```
koslambrou, line 165: I think we should add property tests for testing the reverse property (e.g. `\time -> utcTimeToPOSIXTime time == (posixTimeToUTCTime $ utcTimeToPOSIXTime time)`. What do you think?

koslambrou (changes requested):
Most the tests have activated the `allowBigTransactions` option. Is that a good thing? Does it mean that the examples will not work on the current Cardano network.

sjoerdvisscher:
> Does it mean that the examples will not work on the current Cardano network.

It does. Basically anything using the state machine library is too big. And some of the other bigger contracts too.

plutus-ledger/src/Ledger/Params.hs
sjoerdvisscher: It was updated to 6 with the time conversion issues.

plutus-use-cases/test/Spec/Governance.hs
```diff
@@ -32,17 +32,17 @@ import Test.Tasty.HUnit qualified as HUnit
...
 tests =
     testGroup "governance tests"
-    [ checkPredicate "vote all in favor, 2 rounds - SUCCESS"
+    [ checkPredicateOptions (defaultCheckOptions & allowBigTransactions) "vote all in favor, 2 rounds - SUCCESS"
         (assertNoFailedTransactions
         .&&. dataAtAddress (Scripts.validatorAddress $ Gov.typedValidator params) (maybe False ((== lawv3) . Gov.law) . listToMaybe))
         (doVoting 10 0 2)
 
-    , checkPredicate "vote 60/40, accepted - SUCCESS"
+    , checkPredicateOptions (defaultCheckOptions & allowBigTransactions) "vote 60/40, accepted - SUCCESS"
```
ak3n, line 40: Should `allowBigTransactions` be a default option?

ak3n (approved):
It seems like a half of my #419 because I had to pass protocol params in the same places.

Resolving conflicts will be fun.

plutus-use-cases/test/Spec/Governance.hs
sjoerdvisscher: No, because that could lead to somebody concluding their code works on Mainnet while it doesn't.

plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
sjoerdvisscher: Thanks, after a second look this turned out not to be necessary.

koslambrou (approved):

PR #476, opened 2022-05-24
Add plutus-streaming package

andreabedini:
Hi there,

this brings into main the plutus-streaming "library" I had written. I am using quotes because it's effectively one function :D

~It still has one bug where the user gets a STM exception if the chain-sync client cannot connect. I need to find a way to handle this.~ solved.

Pre-submit checklist:
- Branch
    - [ ] Tests are provided (if possible)
    - [ ] Commit sequence broadly makes sense
    - [ ] Key commits have useful messages
    - [ ] Relevant tickets are mentioned in commit messages
    - [ ] Formatting, materialized Nix files, PNG optimization, etc. are updated
- PR
    - [ ] Self-reviewed the diff
    - [ ] Useful pull request description
    - [ ] Reviewer requested

PR #480, opened 2022-05-25
Initial DatumHash -> Datum index implementation.

raduom:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-chain-index/app/Indexer.hs
+++ b/plutus-chain-index/app/Indexer.hs
@@ -5,7 +5,7 @@
 module Main where
 
 import Cardano.Api (Block (Block), BlockHeader (BlockHeader), BlockInMode (BlockInMode), CardanoMode,
-                    NetworkId (Mainnet))
+                    NetworkId (Mainnet), SlotNo)
 import Cardano.Api qualified as C
 import Cardano.BM.Trace (nullTracer)
 import Cardano.Index.Datum (DatumIndex)
@@ -15,7 +15,6 @@ import Control.Concurrent (threadDelay)
 import Control.Monad (forever)
 import Data.IORef (IORef, newIORef, readIORef, writeIORef)
 import Data.Map (assocs)
-import Data.Maybe (catMaybes)
 import Ledger.TimeSlot (SlotConfig (..))
 import Plutus.ChainIndex.Tx (ChainIndexTx (..))
 import Plutus.Contract.CardanoAPI (fromCardanoTx)
@@ -34,8 +33,8 @@ slotConfig =
 networkId :: NetworkId
 networkId = Mainnet
 
-getDatums :: BlockInMode CardanoMode -> [(DatumHash, Datum)]
-getDatums (BlockInMode (Block _ txs) era) =
+getDatums :: BlockInMode CardanoMode -> [(SlotNo, (DatumHash, Datum))]
+getDatums (BlockInMode (Block (BlockHeader slotNo _ _) txs) era) =
   case era of
     C.ByronEraInCardanoMode   -> concatMap (go era) txs
     C.ShelleyEraInCardanoMode -> concatMap (go era) txs
@@ -46,10 +45,10 @@ getDatums (BlockInMode (Block _ txs) era) =
     go :: C.IsCardanoEra era
        => C.EraInMode era C.CardanoMode
        -> C.Tx era
-       -> [(DatumHash, Datum)]
+       -> [(SlotNo, (DatumHash, Datum))]
     go era' tx =
-      concat $ assocs . _citxData <$> catMaybes [either (const Nothing) Just (fromCardanoTx era' tx)]
-
+      let hashes = either (const []) (assocs . _citxData) $ fromCardanoTx era' tx
+      in  map (\h -> (slotNo, h)) hashes
 
 processBlock :: IORef DatumIndex -> ChainSyncEvent -> IO ()
 processBlock ixref = \case
--- a/plutus-chain-index/app/Indexer.hs
+++ b/plutus-chain-index/app/Indexer.hs
@@ -6,7 +6,8 @@
 module Main where
 
 import Cardano.Api (Block (Block), BlockHeader (BlockHeader), BlockInMode (BlockInMode), BlockNo (BlockNo), CardanoMode,
-                    ChainPoint, NetworkId (Mainnet), SlotNo, chainPointToSlotNo)
+                    ChainPoint (ChainPoint), Hash, NetworkId (Mainnet), SlotNo (SlotNo), chainPointToSlotNo,
+                    deserialiseFromRawBytesHex, proxyToAsType)
 import Cardano.Api qualified as C
 import Cardano.BM.Trace (nullTracer)
 import Cardano.Index.Datum (DatumIndex)
@@ -18,14 +19,15 @@ import Control.Monad (forever, when)
 import Data.IORef (IORef, newIORef, readIORef, writeIORef)
 import Data.List (findIndex)
 import Data.Map (assocs)
-import Data.Maybe (fromMaybe)
+import Data.Maybe (fromJust, fromMaybe)
+import Data.Proxy (Proxy (Proxy))
+import Data.Text (pack)
+import Data.Text.Encoding (encodeUtf8)
 import Index.VSplit qualified as Ix
 import Ledger.TimeSlot (SlotConfig (..))
 import Plutus.ChainIndex.Tx (ChainIndexTx (..))
 import Plutus.Contract.CardanoAPI (fromCardanoTx)
 import Plutus.Script.Utils.V1.Scripts (Datum, DatumHash)
-
--- Connecting to cardano node.
 
 -- We only care about the mainnet
 slotConfig :: SlotConfig
@@ -37,6 +39,16 @@ slotConfig =
 
 networkId :: NetworkId
 networkId = Mainnet
+
+closeToGoguen :: ChainPoint
+closeToGoguen =
+  ChainPoint
+    (SlotNo 39795032)
+    (fromJust $ parseHash "3e6f6450f85962d651654ee66091980b2332166f5505fd10b97b0520c9efac90")
+
+parseHash :: String -> Maybe (Hash BlockHeader)
+parseHash hash =
+  deserialiseFromRawBytesHex (proxyToAsType Proxy) (encodeUtf8 $ pack hash)
 
 getDatums :: BlockInMode CardanoMode -> [(SlotNo, (DatumHash, Datum))]
 getDatums (BlockInMode (Block (BlockHeader slotNo _ _) txs) era) =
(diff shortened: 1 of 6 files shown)


plutus-chain-index/app/Indexer.hs
```diff
@@ -0,0 +1,107 @@
...
+
+closeToGoguen :: ChainPoint
+closeToGoguen =
+  ChainPoint
+    (SlotNo 39795032)
+    (fromJust $ parseHash "3e6f6450f85962d651654ee66091980b2332166f5505fd10b97b0520c9efac90")
+
+parseHash :: String -> Maybe (Hash BlockHeader)
+parseHash hash =
+  deserialiseFromRawBytesHex (proxyToAsType Proxy) (encodeUtf8 $ pack hash)
```
andreabedini, line 51: :-/ I had made this simpler in https://github.com/input-output-hk/cardano-node/pull/3619 but we don't have it yet.

plutus-chain-index/src/Cardano/Index/Datum.hs
```diff
@@ -0,0 +1,95 @@
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE OverloadedStrings  #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+module Cardano.Index.Datum
```
andreabedini, line 5: Why does this start with `Cardano.` and not with `Plutus` like all our packages?

Can this be `Plutus.Index.Datum`? even the executable is `plutus-indexer` (also `index`? `indexer`? pick one :-))

plutus-chain-index/app/Indexer.hs
```diff
@@ -0,0 +1,107 @@
...
+       -> C.Tx era
+       -> [(SlotNo, (DatumHash, Datum))]
+    go era' tx =
+      let hashes = either (const []) (assocs . _citxData) $ fromCardanoTx era' tx
+      in  map (slotNo,) hashes
+
+processBlock :: IORef DatumIndex -> ChainSyncEvent -> IO ()
+processBlock ixref = \case
+  -- Not really supported
+  Resume point         -> putStrLn ("resume " <> show point) >> pure ()
```
andreabedini, line 73: is this ever going to happen using `runChainSync` ?

plutus-chain-index/app/Indexer.hs
```diff
@@ -0,0 +1,107 @@
...
+    rollbackOffset :: [Ix.Event] -> DatumIndex -> Maybe DatumIndex
+    rollbackOffset events ix = do
+      slot   <- chainPointToSlotNo point
+      offset <- findIndex (any (\(s, _) -> s < slot)) events
+      Ix.rewind offset ix
+
+main :: IO ()
+main = do
+  tix <- Ix.open "datum.sqlite" 2160 >>= newIORef
+  _ <- runChainSync "/tmp/node.sock"
```
andreabedini, line 101: We need to be able to pass a different location of the node socket.

plutus-chain-index/app/Indexer.hs
```diff
@@ -0,0 +1,107 @@
...
+processBlock :: IORef DatumIndex -> ChainSyncEvent -> IO ()
+processBlock ixref = \case
+  -- Not really supported
+  Resume point         -> putStrLn ("resume " <> show point) >> pure ()
+  RollForward blk@(BlockInMode (Block (BlockHeader slotNo _ blockNo@(BlockNo b)) _txs) _era) _tip -> do
+    when (b `rem` 1000 == 0) $
+      putStrLn $ show slotNo <> " / " <> show blockNo
+    ix  <- readIORef ixref
+    ix' <- Ix.insert (getDatums blk) ix
+    writeIORef ixref ix'
```
andreabedini, line 79: Can we use `modifyIORef`? (or the strict one `modifyIORef'`)

```
modifyIORef' ixref (Ix.insert . getDatums blk)
```

andreabedini (changes requested):
I assume this is for https://input-output.atlassian.net/browse/PLT-89. Good work.

This PR still uses the chain sync client rather than the streaming one. Hopefully I will get #476 in soon.

Regarding the `hysterical-screams` dependency, we are waiting on a new repo, then it will be merged in.

I left some comments and I have few requests:

1) We need to be able to pass the node socket from the command line, we cannot hard code it
2) I think we should align the module name with the repo and executable, so `Plutus.` not `Cardano.`
3) We need a README
4) How do we query this? Directly from SQLite? can you add a query example in the README?

raduom:
Here are some clarifications regarding `Indexer.hs`. It's not here to stay. The only purpose of this executable is to gather some performance informations, make sure we can sync with the mainnet fully and under what conditions. It will be replaced by the streaming library implementation, once it gets merged.

plutus-chain-index/app/Indexer.hs
```diff
@@ -0,0 +1,107 @@
...
+import Data.Proxy (Proxy (Proxy))
+import Data.Text (pack)
+import Data.Text.Encoding (encodeUtf8)
+import Index.VSplit qualified as Ix
+import Ledger.TimeSlot (SlotConfig (..))
+import Plutus.ChainIndex.Tx (ChainIndexTx (..))
+import Plutus.Contract.CardanoAPI (fromCardanoTx)
+import Plutus.Script.Utils.V1.Scripts (Datum, DatumHash)
+
+-- We only care about the mainnet
```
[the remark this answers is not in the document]
raduom: Yes. This binary is only here to be able to test that we can sync with the main net.

plutus-chain-index/app/Indexer.hs
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+networkId :: NetworkId
+networkId = Mainnet
+
+closeToGoguen :: ChainPoint
+closeToGoguen =
+  ChainPoint
+    (SlotNo 39795032)
+    (fromJust $ parseHash "3e6f6450f85962d651654ee66091980b2332166f5505fd10b97b0520c9efac90")
```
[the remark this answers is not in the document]
raduom: Yes, that is the reason. I could also remove it as it takes only about 15 minutes to get to this point, but I will add a comment.

plutus-chain-index/app/Indexer.hs
raduom: Yeah. This is just temporary (tm) :)

plutus-chain-index/src/Cardano/Index/Datum.hs
raduom: My reasoning is that Indexers are generally more connected to `Cardano`, since that is what we are indexing, the blockchain. For example, `UTXOs` don't have any connection with `Plutus`.

plutus-chain-index/app/Indexer.hs
raduom: Probably not, but this is just a temporary way of checking that the code can sync with the main net.

plutus-chain-index/app/Indexer.hs
raduom: Do you still think we need to do this, if we just use the client until `plutus-streaming` gets merged?

plutus-chain-index/app/Indexer.hs
raduom: I don't think we can, since `Ix.insert` is an effect producing function, but we don't need to worry about concurrency either since nothing is concurrent right now.

andreabedini: Oh right, :facepalm:

andreabedini:
> Here are some clarifications regarding `Indexer.hs`. It's not here to stay.

/shrug The things I pointed out are needed for [PLT-89](https://input-output.atlassian.net/browse/PLT-89) to be done. I don't know why you keep saying everything is temporary.

> It will be replaced by the streaming library implementation, once it gets merged.

I just managed to merge it (finally).

raduom:
> 4. How do we query this? Directly from SQLite? can you add a query example in the README?

We don't right now. I am not sure how we could expose this data in a consistent way, but for now we could just query the sqlite db.

plutus-chain-index/app/Indexer.hs
andreabedini: I am not expecting the user to having to change their node configuration to use this indexer. Yes, I don't think hardcoding the socket path is a good idea. (PS, I just merged #476)

andreabedini:
> We don't right now. I am not sure how we could expose this data in a consistent way, but for now we could just query the sqlite db.

Can you add some documentation where you show an example of how to query this from sqlite?

raduom:
> Can you add some documentation where you show an example of how to query this from sqlite?

Added a comment at the top of the file.

plutus-chain-index/src/Cardano/Index/Datum.hs
koslambrou: I agree with @raduom . However, the module name should reflect the package's name. For example, if we call the package `Marconi`, then the module should be `Marconi.Index.Datum`.

plutus-chain-index/src/Cardano/Index/Datum.hs
```diff
@@ -0,0 +1,95 @@
...
+
+instance FromField SlotNo where
+  fromField f = SlotNo <$> fromField f
+
+instance ToField SlotNo where
+  toField (SlotNo s) = SQLInteger $ fromIntegral s
+
+open
+  :: FilePath
+  -> Int
```
koslambrou, line 62: Use `SecurityParam` from Cardano.Api.

plutus-chain-index/src/Cardano/Index/Datum.hs
```diff
@@ -0,0 +1,95 @@
...
+  -> IO DatumIndex
+open dbPath k = do
+  ix <- fromJust <$> Ix.newBoxed query store onInsert k ((k + 1) * 2) dbPath
+  let c = ix ^. Ix.handle
+  SQL.execute_ c "CREATE TABLE IF NOT EXISTS kv_datumhsh_datum (datumHash TEXT PRIMARY KEY, datum BLOB, slotNo INT)"
+  pure ix
+
+query :: DatumIndex -> Query -> [Event] -> IO Result
```
koslambrou, line 70: Add haddock comments on the parameters.
Especially for `[Event]`, why do we need this to query the database?

plutus-chain-index/app/Indexer.hs
```diff
@@ -0,0 +1,145 @@
...
+    { scSlotZeroTime = 1596059091000
+    , scSlotLength   = 1000
+    }
+
+networkId :: NetworkId
+networkId = Mainnet
+
+-- We don't generally need to sync blocks earlier than the Goguen era (other than
+-- testing for memory leaks) so we may want to start synchronising from a slot that
+-- is closer to Goguen era.
```
koslambrou, line 74: For datums, yes. But some indexers will want to index from start (like the utxosByAddress indexer)

koslambrou (changes requested):

plutus-chain-index/app/Indexer.hs
raduom: I will remove it when I will add the other indexer.

koslambrou:
> > Here are some clarifications regarding `Indexer.hs`. It's not here to stay.
> 
> /shrug The things I pointed out are needed for [PLT-89](https://input-output.atlassian.net/browse/PLT-89) to be done. I don't know why you keep saying everything is temporary.
> 
> > It will be replaced by the streaming library implementation, once it gets merged.
> 
> I just managed to merge it (finally).

Since this PR works by itself, how about we do the integration with `plutus-streaming` in a separate PR?

plutus-chain-index/src/Cardano/Index/Datum.hs
raduom: We need something to query the database plus the information that is currently in memory.

koslambrou (approved):

plutus-chain-index/src/Marconi/Index/Datum.hs
```diff
@@ -72,6 +72,12 @@ open dbPath (Depth k) = do
 
 -- | This function is used to query the data stored in the indexer as a whole:
 --   data that can still change (through rollbacks), buffered data and stored data.
+--
+--   Here is a query that takes into account all received events:
+--   > getEvents (ix ^. storage) >>= query ix <hash>
+--
+--   Here is a query that only takes into account stored events:
+--   > query ix <hash> []
```
koslambrou, line 80: I think the examples got mixed up. `query ix <hash> []` should be for *all* received events, no?

raduom: Nope. It's correct. You may want to drop some of the events <K according to risks that you may want to take. So without any events you get a version of the database that has been settled.

The first example adds all unsettled events (<K) and then runs the query.

andreabedini (approved):

PR #482, opened 2022-05-25
Create v2 versions of v1 scripts

newhoggy:
Pre-submit checklist:
- Branch
    - [ ] Tests are provided (if possible)
    - [ ] Commit sequence broadly makes sense
    - [ ] Key commits have useful messages
    - [ ] Relevant tickets are mentioned in commit messages
    - [ ] Formatting, materialized Nix files, PNG optimization, etc. are updated
- PR
    - [ ] Self-reviewed the diff
    - [ ] Useful pull request description
    - [ ] Reviewer requested

PR #486, opened 2022-05-27
Create separate directories for v1 and v2 plutus scripts

Jimbo4350:
Refactor plutus-example to make it easier to add future versions of plutus
Add an example plutus v2

### DIFF
--- a/plutus-example/plutus-example.cabal
+++ b/plutus-example/plutus-example.cabal
@@ -58,14 +58,19 @@ library
 
   hs-source-dirs:       src
 
-  exposed-modules:      PlutusExample.AlwaysFails
-                        PlutusExample.AlwaysSucceeds
-                        PlutusExample.CustomDatumRedeemerGuess
-                        PlutusExample.DatumRedeemerGuess
-                        PlutusExample.MintingScript
+  exposed-modules:      PlutusExample.PlutusVersion1.AlwaysFails
+                        PlutusExample.PlutusVersion1.AlwaysSucceeds
+                        PlutusExample.PlutusVersion1.CustomDatumRedeemerGuess
+                        PlutusExample.PlutusVersion1.DatumRedeemerGuess
+                        PlutusExample.PlutusVersion1.Loop
+                        PlutusExample.PlutusVersion1.MintingScript
+                        PlutusExample.PlutusVersion1.RedeemerContextScripts
+                        PlutusExample.PlutusVersion1.Sum
+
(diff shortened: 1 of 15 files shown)


james-iohk:
What do we do about existing Alonzo scripts that we want to compile as V2? I made this comment on #482: It would be good to be able to compile all existing V1 scripts as V2 without so much code repetition. Perhaps it would be better to keep just V2 functionality in the V2 dir and add V2 serialisation to the existing scripts? To make it less confusing, directories could be named after era and export multiple versions where supported.

plutus-example/src/PlutusExample/PlutusVersion1/RedeemerContextScripts.hs
```diff
@@ -0,0 +1,237 @@
...
+plutusV1RedeemerContextTestScript = Plutus.unValidatorScript validator
+
+pv1RedeemerContextTestScriptBs :: SBS.ShortByteString
+pv1RedeemerContextTestScriptBs = SBS.toShort . LB.toStrict $ serialise plutusV1RedeemerContextTestScript
+
+scriptContextTextPayingScript :: PlutusScript PlutusScriptV1
+scriptContextTextPayingScript = PlutusScriptSerialised pv1RedeemerContextTestScriptBs
+
+
+-- Minting script that checks the minting value, validty interval and
```
newhoggy, line 177: `s/validty/validity`

newhoggy (approved):

plutus-example/test/Test/PlutusExample/Gen.hs
```diff
@@ -16,8 +16,9 @@ import Cardano.Ledger.Era qualified as Ledger
 import Cardano.Ledger.Shelley.UTxO qualified as Ledger
 import Cardano.Ledger.TxIn qualified as Ledger
 import Gen.Cardano.Api.Typed
-import Plutus.V1.Ledger.Api qualified as Plutus hiding (singleton)
-import Plutus.V1.Ledger.Interval qualified as Plutus
+import Ledger qualified as Plutus
```
koslambrou, line 19: You don't seem to need this.

plutus-example/plutus-example.cabal
```diff
@@ -77,6 +82,7 @@ library
                       , containers
                       , ouroboros-consensus
                       , ouroboros-network
+                      , plutus-ledger
```
koslambrou, line 85: You shouldn't need this if I'm not mistaken.

koslambrou (approved):

PR #483, opened 2022-05-25
PLT-161 Updated cardano-node et al. dependencies to the latest version

koslambrou:
Updating the `cardano-node` dependency once this PR is merged: https://github.com/input-output-hk/cardano-node/pull/3818

@Jimbo4350 

Pre-submit checklist:
- Branch
    - [ ] Tests are provided (if possible)
    - [x] Commit sequence broadly makes sense
    - [x] Key commits have useful messages
    - [x] Relevant tickets are mentioned in commit messages
    - [x] Formatting, materialized Nix files, PNG optimization, etc. are updated
- PR
    - [x] Self-reviewed the diff
    - [x] Useful pull request description
    - [ ] Reviewer requested

PR #488, opened 2022-05-30
[chain-index]: add inline datums support

ak3n:
- Adds inline datum support to chain-index
- Subsumes #483 and improves it a bit.
- Updates cardano-wallet and other dependencies

---

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-chain-index-core/src/Plutus/ChainIndex/Tx.hs
```diff
@@ -42,20 +58,75 @@ import Data.Set qualified as Set
...
+fromCardanoTxOutRefScript :: C.ReferenceScript era -> ReferenceScript
+fromCardanoTxOutRefScript = \case
+    C.ReferenceScriptNone      -> ReferenceScriptNone
+    C.ReferenceScript _ script -> ReferenceScriptInAnyLang script
+
+data ChainIndexTxOut = ChainIndexTxOut
+  { citoAddress   :: Address -- ^ We can't use AddressInAnyEra here because of missing FromJson instance for Byron era
+  , citoValue     :: Value
+  , citoDatum     :: OutputDatum
+  , citoRefScript :: ReferenceScript
```
koslambrou, line 94: For these three, why not reuse the ones from `Cardano.Api`?

ak3n: They are very annoying as they are parametrized with the era.

We would need to convert `RefinementScript era` to `RefinementScript BabbageEra` anyway, so I decided to define chain index's ADT for easier code and better control.

Datum misses JSON instances for example. They are not exported.

If it's not worth it, I could try again to use the cardano-api's types.

koslambrou: Not at all. Let's keep it this way for now.

sjoerdvisscher: Another option could be `PlutusLedgerApi.V2.TxOut`?

koslambrou: `TxOut` is designed for on-chain use. Thus, the reference script part of the tx output only contains the hash and not the actual script (which is what the spec requires of it).

plutus-ledger/src/Ledger/Tx/CardanoAPI.hs
```diff
@@ -718,6 +754,11 @@ toCardanoScriptInEra script = C.ScriptInEra C.PlutusScriptV1InAlonzo . C.PlutusS
 fromCardanoPlutusScript :: C.HasTypeProxy lang => C.PlutusScript lang -> P.Script
 fromCardanoPlutusScript = Codec.deserialise . BSL.fromStrict . C.serialiseToRawBytes
 
+fromCardanoPlutusScriptOrReferenceInput :: C.HasTypeProxy lang => C.PlutusScriptOrReferenceInput lang -> P.Script
+fromCardanoPlutusScriptOrReferenceInput = \case
+  C.PScript plutusScript -> fromCardanoPlutusScript plutusScript
+  C.PReferenceScript _   -> error ("TODO: implement PReferenceScript")
```
ak3n, line 760: @koslambrou what should we do in this case?

koslambrou: Purge (or deprecate)? Ultimately, this shows a problem with the design of `TxIn` and `TxInType` in `plutus`. What's more, `fromCardanoTxInWitness` is not used anywhere else in the codebase, so I'm questionning it's use.

Another solution might be to move `TxInType` in `plutus-apps` and extend it to do something similar to `cardano-api`. I think `plutus` will probably remove it in the future as it's legacy code which isn't used by `cardano-ledger`. Am I right @michaelpj ?

michaelpj: Um, I'm not quite following. What's wrong with `TxInType`? It is used as part of the `ScriptContext` so we can't change it super easily.

koslambrou: Also, I can't seem to find where `TxInType` is used in the `ScriptContext`. Are you mixing it with `TxInInfo`?

michaelpj: Yep, I'm 100% confused. Indeed, it seems those types should get purged from `plutus`!

plutus-pab/src/Plutus/PAB/LocalCluster/Run.hs
```diff
@@ -179,6 +180,7 @@ runWith userContractHandler = withLocalClusterSetup $ \dir lo@LogOutputs{loClust
             void $ serveWallet
                 (NodeSource socketPath vData)
                 gp
+                tunedForMainnetPipeliningStrategy
```
ak3n, line 183: Should be configurable?

koslambrou: Configurable with testnet as well?

ak3n: Well, we may provide it as on option. I was more talking about should we provide a CLI option to choose the strategy for local cluster users? Or they can just patch it themselves.

PR #493, opened 2022-06-02
Utxo indexer

raduom:
This adds the Utxo indexer as an option to the Marconi binary.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


raduom:
There is still some thought to be put in `Indexer.hs`, but the indexer itself is good enough to start reviewing.

plutus-chain-index/app/Indexer.hs
```diff
@@ -84,44 +90,119 @@ parseHash hash =
...
+  withIsCardanoEra era $ concatMap (getDatum slotNo era) txs
+
+getDatum
+  :: C.IsCardanoEra era
+  => SlotNo
+  -> C.EraInMode era C.CardanoMode
+  -> C.Tx era
+  -> [(SlotNo, (DatumHash, Datum))]
+getDatum slotNo era tx =
+      let hashes = either (const []) (assocs . _citxData) $ fromCardanoTx era tx
```
andreabedini, line 103: I would avoid using `fromCardanoTx` here. It's defined in `plutus-chain-index-core` while both the inputs and the outputs of this function only need `plutus-ledger-api` (which is a dependency of `plutus-chain-index-core`).

plutus-chain-index/app/Indexer.hs
```diff
@@ -84,44 +90,119 @@ parseHash hash =
...
+getUpdate (BlockInMode (Block (BlockHeader slotNo _ _) txs) era) =
+  withIsCardanoEra era $ getUtxoUpdate slotNo era txs
+
+getOutputs
+  :: C.IsCardanoEra era
+  => C.EraInMode era C.CardanoMode
+  -> C.Tx era
+  -> Maybe [(TxOut, TxOutRef)]
+getOutputs era tx = do
+  tx' <- either (const Nothing) Just $ fromCardanoTx era tx
```
andreabedini, line 116: same as above

plutus-chain-index/app/Indexer.hs
```diff
@@ -84,44 +90,119 @@ parseHash hash =
...
+    outputs InvalidTx      = []
+    outputs (ValidTx outs) = outs
+
+getInputs
+  :: C.IsCardanoEra era
+  => C.EraInMode era C.CardanoMode
+  -> C.Tx era
+  -> Maybe (Set TxOutRef)
+getInputs era tx = do
+  tx' <- either (const Nothing) Just $ fromCardanoTx era tx
```
andreabedini, line 133: same as above

andreabedini (commented):
How does this integrate with the existing entry-point? As it stands now, you just replace the datum indexer with the new utxo indexer. How can I run both?

plutus-chain-index/app/Indexer.hs
koslambrou: We won't need this once we use the streaming component right?

andreabedini: It's not the streaming component that gets data out of the blocks, it's the stream consumer that does that.
That said, the current version on main doesn't use `fromCardanoTx` so perhaps this just needs a rebase
https://github.com/input-output-hk/plutus-apps/blob/main/plutus-chain-index/app/Marconi.hs#L103

Other functions below have been added though.

andreabedini:
@raduom 

> You can't. That is the next issue I have in the sprint.

I didn't mean "at the same time". Let me rephrase that. How do I run the datum indexer after this is done.

plutus-chain-index/app/Indexer.hs
andreabedini: I was thinking of this version

https://github.com/input-output-hk/plutus-apps/blob/main/plutus-streaming/examples/Example2.hs#L32-L38

raduom: I think this is a good idea, especially as we should be trying to move away from chain-index specific data structures in order to eventually phase them out as marconi starts to get center-stage.

raduom:
> Need work to integrate with existing entry point (or create a new one).

We will use some command line options to specify which indexer should be used.

andreabedini (approved):

koslambrou (approved):

PR #500, opened 2022-06-07
PLT-106: Add an executable for dumping script evaluation events

zliu41:
The executable does the following:
- Stream and apply blocks
- Periodically writes `ChainPoint` and `LedgerState` into `timestamp.state` files, and writes `LedgerEvent`s into `timestamp.event` files
- Upon restart, it looks for the latest `state` file, and resumes streaming from that point. If it fails (because the checkpointed block was rolled back), it then tries the second latest `state` file

This PR has a few placeholders (`error`) because the `cardano-node` commit is too old, but I tested it locally on a newer `cardano-node` commit.

The implementation of `ledgerStateEvents` was provided by @andreabedini (with minor modification).

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -14,6 +14,7 @@ packages: doc
           plutus-pab
           plutus-pab-executables
           plutus-playground-server
+          plutus-script-evaluation-test
           plutus-script-utils
           plutus-tx-constraints
           plutus-use-cases
--- a/plutus-script-evaluation-test/dump-script-events/Main.hs
+++ b/plutus-script-evaluation-test/dump-script-events/Main.hs
@@ -1,0 +1,20 @@
+module Main (main) where
+
+import Options.Applicative qualified as O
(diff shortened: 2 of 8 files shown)


koslambrou:
Instead of a new package, I would put this as an executable in `plutus-use-cases`. We already have a tool in there which writes validators scripts and partial transactions on disk, so it would be logical to have another executable which does what you want.

andreabedini (changes requested):
Can you move these ledger related functions to a different module? Eg. 
Plutus.Streaming.LedgerState ?
If you look in the git history I used to have a module like that but I had decided to remove it to keep things minimal. I take the fact that you need it again as a signal that I was wrong! :) 

I notice the hack around epochSlots, let's just accept it and refer that discussion to #499

koslambrou:
@zliu41 Still relevant, or can I close this PR?

zliu41:
Still relevant, but I need to get some upstream PRs merged before working further on this.

PR #501, opened 2022-06-07
ContractModel: Detect Double Satisfaction Problems

MaximilianAlgehed:
This PR introduces a new property to check for double satisfaction vulnerabilities in any contract with a `ContractModel`. This property is not complete, but as far as we know it is sound. See the `checkDoubleSatisfaction` property in `Plutus.Contract.Test.ContractModel`.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-contract/src/Plutus/Contract/Test/ContractModel/DoubleSatisfaction.hs
+++ b/plutus-contract/src/Plutus/Contract/Test/ContractModel/DoubleSatisfaction.hs
@@ -58,8 +58,7 @@ import Ledger.Value (adaOnlyValue)
 import Plutus.Contract.Test hiding (not)
 import Plutus.Contract.Test.ContractModel.Internal
 import Plutus.Script.Utils.V1.Scripts (datumHash, validatorHash)
-import Plutus.Trace.Emulator as Trace (EmulatorTrace, activateContract, callEndpoint, runEmulatorStream,
-                                       walletInstanceTag)
+import Plutus.Trace.Emulator as Trace (EmulatorTrace, activateContract, callEndpoint, runEmulatorStream)
 import Plutus.V1.Ledger.Address
 import Plutus.V1.Ledger.TxId
 import Streaming qualified as S
--- a/plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
+++ b/plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
@@ -215,7 +215,7 @@ import Plutus.V1.Ledger.Ada qualified as Ada
 import Ledger.Index as Index
 import Ledger.Scripts
 import Ledger.Slot
-import Ledger.Value (AssetClass, adaOnlyValue)
+import Ledger.Value (AssetClass)
 import Plutus.Contract (Contract, ContractError, ContractInstanceId, Endpoint, endpoint)
 import Plutus.Contract.Schema (Input)
 import Plutus.Contract.Test hiding (not)


MaximilianAlgehed:
@sjoerdvisscher we didn't touch this code, is something broken on main?

MaximilianAlgehed:
@sjoerdvisscher we didn't touch this code, is something broken on main?

locallycompact:
What is a double satisfaction problem?

ThomasArts:
> What is a double satisfaction problem?

https://plutus.readthedocs.io/en/latest/reference/common-weaknesses/double-satisfaction.html?highlight=double%20satisfication

sjoerdvisscher:
@MaximilianAlgehed Hmm, it looks like a legitimate failure, and I don't see it on other branches. Did you try to run `cabal test plutus-pab-test-full-long-running` locally?

MaximilianAlgehed:
... It's a cached failure?!

sjoerdvisscher:
As far as I can tell it is cached from a previous build of this PR.

MaximilianAlgehed:
@sjoerdvisscher there is a change to `Ledger.Index` here! But that's because the semantics of the old code are, as far as I can tell, wrong! The semantics of the old code is that you always pay the collateral, when the semantics should be that you only pay if phase 2 validation fails, no?

sjoerdvisscher:
> the semantics should be that you only pay if phase 2 validation fails, no?

Yes, that's definitely how it's supposed to work.

MaximilianAlgehed:
Then the fact that this test is broken tells us something. Is something broken in the PAB or is something broken elsewhere?

MaximilianAlgehed:
@sjoerdvisscher I'll make a separate PR with that fix and hopefully that means we can dig into the issue in the PAB separately.

MaximilianAlgehed:
Also, @sjoerdvisscher it would be good if you or someone could just have a look over this to make sure there is nothing absolutely insane going on :)

sjoerdvisscher:
> So then the old version was buggy (because of how do and infix notation interact).

😱

> it would be good if you or someone could just have a look over this to make sure there is nothing absolutely insane going on :)

As a comment in the code says `-- Double satisfaction magic`, and as with any magic it is hard to tell if something insane is going on or not! To be honest this is true for most of the ContractModel code. In can review it for code quality (and this is already quite a task since most PRs are quite big) but I can't review the logic.

Don't forget to rebase on main to get the fix for the OOM CI issue.

sjoerdvisscher (approved):

UlfNorell:
@sjoerdvisscher: 43b73aca2c changed `runValidation` to no longer compute an updated `UtxoIndex`. What's the best way to compute that now? We need to keep an up-to-date `UtxoIndex` when traversing an emulator trace looking for double satisfaction vulnerabilities.

sjoerdvisscher:
Ah sorry, I keep forgetting that we're not the only ones using these functions. You need something like this:

```haskell
        idx' = case e of
            Just (Index.Phase1, _) -> idx
            Just (Index.Phase2, _) -> Index.insertCollateral txn idx
            Nothing                -> Index.insert txn idx
```

UlfNorell:
Does that mean we should to switch to using `CardanoTx`s instead of the emulator `Tx`? Currently we're expecting to find an emulator `Tx` in `TxnValidate` events, but that will now be a `CardanoTx`?

PR #502, opened 2022-06-08
PLT-380 Replace TChan with bounded queue

andreabedini:
Someone made me notice that I was using an unbounded tchan where I could get away with a mvar. Basically I want to have the two threads (the one running chain-sync and the stream consumer) in lock-step.

In the current version, blocks are pushed in the channel at full speed independently from the speed at which they are consumed, this can cause memory issues if the consumer is slower than the network (which is likely).

### DIFF
--- a/plutus-streaming/src/Plutus/Streaming.hs
+++ b/plutus-streaming/src/Plutus/Streaming.hs
@@ -56,12 +56,12 @@ withChainSyncEventStream socketPath networkId point consumer = do
   -- other. The problem here is "properly-sized". A bounded queue like
   -- Control.Concurrent.STM.TBQueue allows us to specify a max queue length
   -- but block size can vary a lot (TODO quantify this) depending on the
-  -- era. We have an alternative implementation with customizable
-  -- (TBMQueue) but it needs to be extracted from the
+  -- era. We have an alternative implementation with customizable queue
+  -- size (TBMQueue) but it needs to be extracted from the
   -- plutus-chain-index-core package. Using a simple MVar doesn't seem to
   -- slow down marconi's indexing, likely because the difference is
-  -- negligeable compared to existing network and IO latencies.
-  -- Therefore, let's stick with a MVar now and revisit later.
+  -- negligeable compared to existing network and IO latencies.  Therefore,
+  -- let's stick with a MVar now and revisit later.
   nextBlockVar <- newEmptyMVar
 
   let client = chainSyncStreamingClient point nextBlockVar
@@ -102,7 +102,7 @@ chainSyncStreamingClient ::
   ChainPoint ->
   MVar (ChainSyncEvent e) ->
   ChainSyncClient e ChainPoint ChainTip IO ()
-chainSyncStreamingClient point nextBlockVar =
+chainSyncStreamingClient point nextChainEventVar =
   ChainSyncClient $ pure $ SendMsgFindIntersect [point] onIntersect
   where
     onIntersect =
@@ -121,10 +121,10 @@ chainSyncStreamingClient point nextBlockVar =
           ClientStNext
             { recvMsgRollForward = \bim ct ->
                 ChainSyncClient $ do
-                  putMVar nextBlockVar (RollForward bim ct)
+                  putMVar nextChainEventVar (RollForward bim ct)
                   sendRequestNext,
               recvMsgRollBackward = \cp ct ->
                 ChainSyncClient $ do
-                  putMVar nextBlockVar (RollBackward cp ct)
+                  putMVar nextChainEventVar (RollBackward cp ct)
                   sendRequestNext
             }
--- a/plutus-streaming/src/Plutus/Streaming.hs
+++ b/plutus-streaming/src/Plutus/Streaming.hs
@@ -74,14 +74,7 @@ withChainSyncEventStream socketPath networkId point consumer = do
       -- FIXME this comes from the config file but Cardano.Api does not expose readNetworkConfig!
       epochSlots = EpochSlots 40
 
-      clientThread = do
-        connectToLocalNode connectInfo localNodeClientProtocols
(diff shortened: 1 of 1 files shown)


sjoerdvisscher (approved):
I'm completely inexperienced when it comes to concurrency programming (in Haskell or any other language), but this all sounds very reasonable. And it is simpler, and simple is good 👍

ak3n:
You may use [`Control.Concurrent.STM.TBQueue`](https://hackage.haskell.org/package/stm-2.5.0.2/docs/Control-Concurrent-STM-TBQueue.html) to accumulate the items with a limited size queue to control the memory usage. That might have a bigger throughput than using a single `MVar`.

koslambrou:
I second @ak3n. The `TBMQueue` will solve the memory issues.

andreabedini:
> You may use [`Control.Concurrent.STM.TBQueue`](https://hackage.haskell.org/package/stm-2.5.0.2/docs/Control-Concurrent-STM-TBQueue.html) to accumulate the items with a limited size queue to control the memory usage. That might have a bigger throughput than using a single `MVar`.

You are right that throughput is important (which reminds me I should implement pipelining as well).

> UPD. For blocks with transactions we have [`Control.Concurrent.STM.TBMQueue`](https://github.com/input-output-hk/plutus-apps/blob/main/plutus-chain-index/src/Control/Concurrent/STM/TBMQueue.hs) which allows to use a measure function to count the number of transactions in blocks instead of the number blocks. The example of usage might be found in [chain index](https://github.com/input-output-hk/plutus-apps/blob/main/plutus-chain-index/src/Plutus/ChainIndex/Events.hs#L43).

`Control.Concurrent.STM.TBMQueue` looks nice and the ability of bounding the channel size in terms of transactions rather than blocks is interesting but it's in plutus-chain-index and plutus-streaming cannot depend on that. Maybe worth extracting it out at some point?

I'll change this PR to use `Control.Concurrent.STM.TBQueue`

ak3n:
`Control.Concurrent.STM.TBQueue` is unreliable in terms of memory usage in this case because the number of transactions in the blocks could take much more memory with a fixed number of blocks queue. We experienced that with chain index. 

>Maybe worth extracting it out at some point?

I have extracted it to a separate personal repo https://git.sr.ht/~ak3n/tbmqueue some time ago where I wanted to play with it and maybe tune but had no time and haven't uploaded to hackage (also it seems I made a mistake with the LICENSE file).

We can start using it which will make me a maintainer and maybe I will extend it. Or we can extract the module to a new package within the plutus-apps repo. Or just copy the module until the dependencies picture will be clearer.

andreabedini:
> `Control.Concurrent.STM.TBQueue` is unreliable in terms of memory usage in this case because the number of transactions in the blocks could take much more memory with a fixed number of blocks queue. We experienced that with chain index.

Do you happen to remember how big was the variation? I am a bit conflicted. I don't even have evicence of lower throughput with an MVar. I will run some tests tomorrow to decide what to do.

andreabedini:
@ak3n @koslambrou I appreciate your feedback but I have decided to stick with a dead simple MVar. I added a comment explaining the thread off. I am concerned about spending time optimising for the wrong thing. We can revise this later if it becomes a problem.

plutus-streaming/src/Plutus/Streaming.hs
```diff
@@ -41,21 +41,30 @@ withChainSyncEventStream ::
...
+  -- thread and the stream consumer will each block on each other and stay
+  -- in lockstep.
+  --
+  -- NOTE: choosing a MVar is a tradeoff towards simplicity. In this case a
+  -- (bounded) queue could perform better. Indeed a properly-sized buffer
+  -- can reduce the time the two threads are blocked waiting for each
+  -- other. The problem here is "properly-sized". A bounded queue like
+  -- Control.Concurrent.STM.TBQueue allows us to specify a max queue length
+  -- but block size can vary a lot (TODO quantify this) depending on the
+  -- era. We have an alternative implementation with customizable
```
koslambrou, line 59: "customizable queue size"?

plutus-streaming/src/Plutus/Streaming.hs
```diff
@@ -94,26 +96,22 @@ withChainSyncEventStream socketPath networkId point consumer = do
...
---
 -- If the starting point is such that an intersection cannot be found, this
 -- client will throw a NoIntersectionFound exception.
 chainSyncStreamingClient ::
   ChainPoint ->
-  TChan (ChainSyncEvent e) ->
+  MVar (ChainSyncEvent e) ->
   ChainSyncClient e ChainPoint ChainTip IO ()
-chainSyncStreamingClient point chan =
+chainSyncStreamingClient point nextBlockVar =
```
koslambrou, line 105: better name would be `nextChainEventVar` (or something similar) no?

koslambrou:
> @ak3n @koslambrou I appreciate your feedback but I have decided to stick with a dead simple MVar. I added a comment explaining the thread off. I am concerned about spending time optimising for the wrong thing. We can revise this later if it becomes a problem.

No problem, we can add a story for this. I may be wrong, but feels like using something like `TBMQueue` will indeed improve performance (it **did** improve the performance of `plutus-chain-index`, and the current use case is very similar. I might be wrong.

koslambrou (approved):

plutus-streaming/src/Plutus/Streaming.hs
andreabedini: I accidentally a word. Thank you.

andreabedini:
> I may be wrong, but feels like using something like `TBMQueue` will indeed improve performance (it **did** improve the performance of `plutus-chain-index`, and the current use case is very similar. I might be wrong.

I don't mean to say you are wrong! I saw the issue on JIRA :+1:

PR #509, opened 2022-06-10
PLT-381 Fix the value of epochSlots and add a comment

andreabedini:
As per internal discussion.

This a parameter needed only for the Byron era. Since the Byron era is over and the parameter has never changed it is ok to hardcode this.

### DIFF
--- a/plutus-streaming/src/Plutus/Streaming.hs
+++ b/plutus-streaming/src/Plutus/Streaming.hs
@@ -73,7 +73,8 @@ withChainSyncEventStream socketPath networkId point consumer = do
 
       -- This a parameter needed only for the Byron era. Since the Byron
       -- era is over and the parameter has never changed it is ok to
-      -- hardcode this.
+      -- hardcode this. See comment on `Cardano.Api.ConsensusModeParams` in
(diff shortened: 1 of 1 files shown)


plutus-streaming/src/Plutus/Streaming.hs
```diff
@@ -71,8 +71,10 @@ withChainSyncEventStream socketPath networkId point consumer = do
             localNodeSocketPath = socketPath
           }
 
-      -- FIXME this comes from the config file but Cardano.Api does not expose readNetworkConfig!
-      epochSlots = EpochSlots 40
+      -- This a parameter needed only for the Byron era. Since the Byron
+      -- era is over and the parameter has never changed it is ok to
+      -- hardcode this.
```
koslambrou, line 76: :+1:

koslambrou (approved):

plutus-streaming/src/Plutus/Streaming.hs
andreabedini: Added a reference to `Cardano.Api.ConsensusModeParams` :+1:

PR #511, opened 2022-06-10
A story of two indexers

raduom:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-chain-index/app/Marconi.hs
```diff
@@ -49,16 +55,12 @@ import Streaming.Prelude qualified as S
...
-  { optionsSocketPath   :: String,
-    optionsNetworkId    :: NetworkId,
-    optionsChainPoint   :: ChainPoint,
-    optionsDatabasePath :: FilePath,
-    optionsIndexerType  :: IndexerType
+  { optionsSocketPath :: String,
+    optionsNetworkId  :: NetworkId,
+    optionsChainPoint :: ChainPoint,
+    optionsUtxoPath   :: Maybe FilePath,
+    optionsDatumPath  :: Maybe FilePath
```
koslambrou, line 63: Curious, is it possible to put the data from the indexers in the same database?

plutus-chain-index/app/Marconi.hs
```diff
@@ -174,50 +156,106 @@ getUtxoUpdate slot txs =
                  , _slotNo  = slot
                  }
 
-utxoIndexer
-  :: FilePath
+data Coordinator = Coordinator
```
koslambrou, line 159: Comment. Why is this needed?

plutus-chain-index/app/Marconi.hs
```diff
@@ -174,50 +156,106 @@ getUtxoUpdate slot txs =
...
-
-    finish :: UtxoIndex -> IO ()
-    finish _index = pure ()
+    initial :: IO Coordinator
+    initial = do
+      let indexerCount = length . catMaybes $ [utxoPath, datumPath]
+      coordinator <- initialCoordinator indexerCount
+      when (isJust datumPath) $ do
+        ch <- atomically . dupTChan $ _channel coordinator
+        void . forkIO . datumWorker coordinator ch $ fromJust datumPath
```
koslambrou, line 228: Why not do the `dupTChan` inside the `datumWorker` instead of passing it as a parameter?

plutus-chain-index/app/Marconi.hs
```diff
@@ -174,50 +156,106 @@ getUtxoUpdate slot txs =
...
+      when (isJust utxoPath) $ do
+        ch <- atomically . dupTChan $ _channel coordinator
+        void . forkIO . utxoWorker coordinator ch $ fromJust utxoPath
```
koslambrou, line 231: Same as above

plutus-chain-index/app/Marconi.hs
```diff
@@ -174,50 +156,106 @@ getUtxoUpdate slot txs =
...
+          events <- Ix.getEvents (index ^. Ix.storage)
+          innerLoop $
+            fromMaybe index $ do
+              slot   <- chainPointToSlotNo cp
+              offset <- findIndex  (\u -> (u ^. Utxo.slotNo) < slot) events
+              Ix.rewind offset index
+
+combinedIndexer
+  :: Maybe FilePath
+  -> Maybe FilePath
```
koslambrou, line 217: This approach does not seem extensible for multiple indexers. Why not pass a list of workers? They seem to have a similar interface.

plutus-chain-index/app/Marconi.hs
```diff
@@ -174,50 +156,106 @@ getUtxoUpdate slot txs =
...
+  Coordinator <$> newBroadcastTChanIO
+              <*> newQSemN 0
+              <*> pure indexerCount
+
+datumWorker
+  :: Coordinator
+  -> TChan (ChainSyncEvent (BlockInMode CardanoMode))
+  -> FilePath
+  -> IO ()
+datumWorker Coordinator{_barrier} ch path = Datum.open path (Datum.Depth 2160) >>= innerLoop
```
koslambrou, line 176: The security param `2160` shouldn't be hardcoded right?

andreabedini (dismissed):
I am wondering if making the two indexes explicitly concurrent is worth the effort. If we keep concurrency implict (because most of the computation is pure) the entire PR boils down to 

```
withStream bla bla $ L.impurely S.foldM_ (mconcat [consumer1, consumer2]) 
```

It's not clear to me that explicit concurrency brings an advange in terms of performace. I have found some discussions [1](https://www.sqlite.org/threadsafe.html) [2](https://github.com/nurpax/snaplet-sqlite-simple/issues/5) about SQLite and multithreading but I haven't dug into it yet. 

AFAIU each indexer opens a new sqlite database so there should not be any contention at the connection level, is this right?

I have noticed you use `Control.Concurrent.STM.TChan` (like I did in plutus-streaming). This is an unbounded channel and it's memory usage will grow without limits if the producer is faster than the consumer. I believe this is likely here. Can you either use a bounded channel or make sure this arbitrary grow doesn't happen in other means?

plutus-chain-index/app/Marconi.hs
raduom: In the case of SQLite, most likely not due to higher contention.

plutus-chain-index/app/Marconi.hs
raduom: Actually thinking about how this works, this is a possible multi-threading bug. Let me fix it first :)

plutus-chain-index/app/Marconi.hs
raduom: It can probably be done, but it's not trivial. You need to provide specific configurations for each indexer, and the startup may also be different (not the case here). The issue mentioned 2 indexers and I am not sure if supporting multiple indexers is on the bird path, but if it is, it should be a separate issue, I think.

plutus-chain-index/app/Marconi.hs
raduom: I am not sure why it shouldn't be hardcoded in this case. Would you like to expose it as a command-line argument? One for all indexers, or one command-line argument for each? I can write the code if you think it's important to do so.

plutus-chain-index/app/Marconi.hs
raduom: Ok. So there is no bug. 

Moving the channel creation in the function would mean I need to define an inner loop function which I preferred not to do.

andreabedini (approved):

plutus-chain-index/app/Marconi.hs
koslambrou: :+1:

koslambrou (approved):

PR #519, opened 2022-06-17
Make LocalNodeConnectionInfo available to streamer consumer

zliu41:
This makes it easier to do such things as obtaining the epoch number of a block, which requires executing a local state query. See Example4.hs.  cc @andreabedini

### DIFF
--- a/plutus-chain-index/app/Marconi.hs
+++ b/plutus-chain-index/app/Marconi.hs
@@ -266,7 +266,7 @@ main = do
       optionsSocketPath
       optionsNetworkId
       optionsChainPoint
-      (combinedIndexer optionsUtxoPath optionsDatumPath . logging trace)
+      (\_connInfo -> combinedIndexer optionsUtxoPath optionsDatumPath . logging trace)
       `catch` \NoIntersectionFound ->
         logError trace $
           renderStrict $
--- a/plutus-streaming/README.md
+++ b/plutus-streaming/README.md
@@ -24,6 +24,11 @@ id, ins and outs) and datums. Then it prints them to stdout in JSON format.
 This application uses folds the stream into a UTxoState as defined in
 plutus-chain-index-core.
 
+## Example 4
+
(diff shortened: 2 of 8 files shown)


andreabedini:
Thanks @zliu41, connectInfo is basically built from the arguments (`networkId` and `socketPath`) and from the constant `epochSlots`. (Maybe we should call it `byronEpochSlots`?).

 ```
connectInfo = LocalNodeConnectInfo
          { localConsensusModeParams = CardanoModeParams epochSlots,
            localNodeNetworkId = networkId,
            localNodeSocketPath = socketPath
}
```

So what this change really does is "just" exposing `epochSlots`. In Example4 you reuse `connectInfo` for another connection which can indeed be handy.

I would suggest making a version of `withChainSyncEventStream` called that `withChainSyncEventStreamFromLocalNodeConnectInfo` (:rofl: we need a shorter name) that takes directly a `LocalNodeConnectInfo` as a parameter. So the user makes a `LocalNodeConnectInfo` on their own and then use it multiple times. 

E.g.

```
withChainSyncEventStream :: FilePath -> NetworkId -> ChainPoint -> (Stream (Of (ChainSyncEvent (BlockInMode CardanoMode))) IO r -> IO b) -> IO b

withChainSyncEventStreamFromLocalNodeConnectInfo :: LocalNodeConnectInfo -> ChainPoint -> (Stream (Of (ChainSyncEvent (BlockInMode CardanoMode))) IO r -> IO b) -> IO b
```

What do you think?

zliu41:
@andreabedini Sounds reasonable - I'll update the PR.

By the way, do you know if there's a remote counterpart of `LocalNodeConnectInfo` and `executeLocalStateQueryExpr`? I'd be hesitant to rely on this to obtain epoch numbers if this requires running a local node.

koslambrou:
@zliu41 Still relevant, or can I close this PR?

Issue #282, opened 2022-01-27
Mention --rollback-history in the integration test file.


raduom:
### Describe the feature you'd like

We should mention this there, as anyone that will try it without this option will get their RAM blown away.

### Describe alternatives you've considered

This will go away when we get persistence for the PAB working.

PR #524, opened 2022-06-20
Transaction state indexer.

raduom:
This fix introduces an alternative to the in-memory only indexer used in the PAB so far. I am hiding the new indexer under an option that can be specified in the PAB configuration file.

The code itself is a bunch of if branches that select the new indexer or the old one when running connected to a real Alonzo node. The new indexer is not made available for the mock node as that is used mostly for testing, with small data sets and needs to run very fast (so in-memory is a good choice for that use-case).

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-pab/src/Plutus/PAB/Core/ContractInstance.hs
+++ b/plutus-pab/src/Plutus/PAB/Core/ContractInstance.hs
@@ -71,9 +71,11 @@ import Wallet.Emulator.Wallet qualified as Wallet
 
 import Control.Monad.Freer.NonDet (NonDet)
 import Data.Monoid (Sum (Sum))
-import Plutus.ChainIndex (ChainIndexQueryEffect, Depth (..), RollbackState (..), TxConfirmedState (..), TxStatus,
-                          TxValidity (..))
-import Plutus.PAB.Core.ContractInstance.STM (Activity (Done, Stopped), BlockchainEnv,
+import Ledger (TxId, TxOutRef (..))
+import Plutus.ChainIndex (ChainIndexQueryEffect, Depth (..), RollbackState (..), TxConfirmedState (..), TxOutState (..),
+                          TxOutStatus, TxStatus, TxValidity (..), transactionOutputState)
+import Plutus.ChainIndex.UtxoState (UtxoState (_usTxUtxoData), utxoState)
+import Plutus.PAB.Core.ContractInstance.STM (Activity (Done, Stopped), BlockchainEnv (..),
                                              InstanceState (InstanceState, issStop), InstancesState,
                                              callEndpointOnInstance, emptyInstanceState)
 import Plutus.PAB.Core.ContractInstance.STM qualified as InstanceState
@@ -83,7 +85,6 @@ import Plutus.PAB.Effects.UUID (UUIDEffect, uuidNextRandom)
 import Plutus.PAB.Events.Contract (ContractInstanceId (ContractInstanceId))
 import Plutus.PAB.Types (PABError)
 import Plutus.PAB.Webserver.Types (ContractActivationArgs (ContractActivationArgs, caID, caWallet))
-import Plutus.V1.Ledger.Api (TxId)
 
 -- | Container for holding a few bits of state related to the contract
 -- instance that we may want to pass in.
@@ -196,7 +197,7 @@ processAwaitTimeRequestsSTM =
         )
 
 processTxStatusChangeRequestsSTM ::
-    forall effs m.
+    forall m effs.
     ( LastMember m effs
     , MonadIO m
     , Member (Reader BlockchainEnv) (NonDet : effs)
@@ -237,8 +238,10 @@ processTxStatusChangeRequestIO ixRef env txId = do
             TentativelyConfirmed (Depth n) TxValid ()
 
 processTxOutStatusChangeRequestsSTM ::
-    forall effs.
-    ( Member (Reader BlockchainEnv) effs
+    forall m effs.
+    ( LastMember m effs
+    , MonadIO m
+    , Member (Reader BlockchainEnv) effs
     )
     => RequestHandler effs PABReq (STM PABResp)
 processTxOutStatusChangeRequestsSTM =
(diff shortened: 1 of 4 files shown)


plutus-chain-index-core/src/Marconi/Index/TxConfirmationStatus.hs
```diff
@@ -0,0 +1,105 @@
...
+    searchOnDisk :: IO Result
+    searchOnDisk = do
+      txStatus :: [(TxId, BlockNumber)]
+        <- SQL.query (ix ^. Ix.handle) "SELECT (txId, blockNo) FROM tx_status WHERE txId = ?" (Only txId)
+      if null txStatus
+         then pure Nothing
+         else let slotNo = snd $ head txStatus
+              in  pure . Just $
+                TxConfirmedState { timesConfirmed = Sum 0
+                                 , blockAdded     = Last $ Just slotNo
```
andreabedini, line 89: is blockAdded the block number or the slot number? above there's `blockAdded = Last $ Just bn` where bn suggests block number.

eyeinsky (approved):
This PR looks good to me, given that:
- my PR #629 is now rebased on top of this and works
- tests will arrive in a future PR

Perhaps it would have been better to make this PR more granular (in three PRs: one moving modules to the library, second adding the new indexer, and third adding it to PAB) -- would have been easier to review, as the third part I can't really review because don't have knowledge of it.

plutus-pab/src/Plutus/PAB/Core.hs
```diff
@@ -106,11 +110,14 @@ import Ledger.Address (PaymentPubKeyHash)
...
 import Ledger.Value (Value)
-import Plutus.ChainIndex (ChainIndexQueryEffect, RollbackState (Unknown), TxOutStatus, TxStatus)
+import Plutus.ChainIndex (ChainIndexQueryEffect, Depth (..), RollbackState (..), TxConfirmedState (..), TxOutStatus,
+                          TxStatus, TxValidity (..))
 import Plutus.ChainIndex qualified as ChainIndex
 import Plutus.ChainIndex.Api qualified as ChainIndex
 import Plutus.Contract.Effects (ActiveEndpoint (ActiveEndpoint, aeDescription), PABReq)
 import Plutus.Contract.Wallet (ExportTx)
+import Plutus.HystericalScreams.Index.VSplit qualified as Ix
+-- import Plutus.HystericalScreams.Index.VSqlite qualified as Ix
```
koslambrou, line 120: Remove

plutus-pab/src/Plutus/PAB/Core/ContractInstance/STM.hs
```diff
@@ -153,19 +157,23 @@ data BlockchainEnv =
         , beCurrentSlot         :: TVar Slot -- ^ Actual current slot
         , beLastSyncedBlockSlot :: TVar Slot -- ^ Slot of the last synced block from 'startNodeClient'
         , beLastSyncedBlockNo   :: TVar BlockNumber -- ^ Last synced block number from 'startNodeClient'.
-        , beTxChanges           :: TVar (UtxoIndex TxIdState) -- ^ Map holding metadata which determines the status of transactions.
+        , beTxChanges           :: Either (TVar (UtxoIndex TxIdState)) (IORef TCSIndex)-- ^ Map holding metadata which determines the status of transactions.
         , beTxOutChanges        :: TVar (UtxoIndex TxOutBalance) -- ^ Map holding metadata which determines the status of transaction outputs.
```
koslambrou, line 161: Shouldn't the same be done with `UtxoIndex TxOutBalance`? This also uses a lot of memory

plutus-pab/src/Plutus/PAB/Core/ContractInstance.hs
```diff
@@ -198,7 +198,11 @@ processTxStatusChangeRequestsSTM =
     where
         handler txId = do
             env <- ask
-            pure (AwaitTxStatusChangeResp txId <$> InstanceState.waitForTxStatusChange Unknown txId env)
+                     -- This should never be called in a context where there is a
+                     -- Marconi indexer in the environment.
```
koslambrou, line 202: Fix indentation for the comment.

Also, I don't understand the "why" from the comment.

koslambrou:
Did you test the following?

- Run PAB on testnet and notice "low" memory usage?
- Run the LocalCluster with `IntegrationTest`? This uses the `awaitTxConfirmed` call which should test your indexer.

plutus-pab/src/Plutus/PAB/Core/ContractInstance/STM.hs
raduom: Actually it does not use that much memory. I would say, let's leave it like this for now as this one is a fairly large commit already and add another issue if you feel we should do that.

raduom:
> Run PAB on testnet and notice "low" memory usage?

I ran it on the mainnet and did notice the low memory usage.

> Run the LocalCluster with IntegrationTest? This uses the awaitTxConfirmed call which should test your indexer.

I did not do this and I am not sure how to do it. I can look into it, but is running this test a blocker for you?

koslambrou:
> I did not do this and I am not sure how to do it. I can look into it, but is running this test a blocker for you?

I think so. This integration does test a bit the indexer. But you need to modify the LocalCluster code in order to tell it to use the Marconi indexer.

To run the indexer, see `pab-local-cluster` section of `plutus-pab-executables/README.md`.

plutus-pab/src/Plutus/PAB/Core/ContractInstance/STM.hs
koslambrou: Aren't the changes to `beTxOutChanges` pretty much the exact same as `beTxChanges`?

raduom: They will need a new indexer. I don't think it's a bad idea, but maybe we should have a separate issue for it.

koslambrou: Will add a ticket

koslambrou (approved):

plutus-pab/src/Plutus/PAB/Core/ContractInstance/STM.hs
```diff
@@ -394,31 +405,44 @@ insertInstance :: ContractInstanceId -> InstanceState -> InstancesState -> STM (
...
+        let txStatus  = transactionStatus blockNumber txIdState tx
+        -- Succeed only if we _found_ a status and it was different; if
+        -- the status hasn't changed, _or_ there was an error computing
+        -- the status, keep retrying.
+        case txStatus of
+          Right s | s /= oldStatus -> pure s
+          _                        -> empty
+      -- This branch gets intercepted in `processTxStatusChangeRequestIO` and
+      -- handled separateley, so we should never reach this place.
+      Right _ ->
```
raduom, line 424: This is probably not my finest moment, but fixing it would mean refactoring quite a bit more code, and I am not sure that is warranted as the in-memory index should eventually go away which will make life a lot simpler.

raduom:
I ran the integration tests with and without Marconi and they both succed.

andreabedini (approved):

eyeinsky:
In the diff I don't see this issue resolved https://github.com/input-output-hk/plutus-apps/issues/282 (linked from Jira as acceptance criteria)

koslambrou (approved):

PR #533, opened 2022-06-22
PLT-333: Buildkite pipeline for script dump job

zliu41:
For pipeline: https://buildkite.com/input-output-hk/plutus-script-dump


.buildkite/plutus-script-dump.yml
```diff
@@ -0,0 +1,11 @@
+# A nightly job that runs a cardano node and dumps Plutus script evaluation events
```
michaelpj, line 1: How do we know it's scheduled nightly? Is that configured in the builkite UI or something?

zliu41: Yep, that's configured in the UI. I believe you can only define steps in the yml file.

PR #534, opened 2022-06-22
deleted Marlowe content, updated project name and copyright year

joseph-fajen:


doc/index.rst
```diff
@@ -1,12 +1,10 @@
-The Plutus Platform and Marlowe
+The Plutus Platform
 ===============================
 
 The Plutus Platform is an application development platform for the Cardano blockchain.
-To read more about the Platform, see :ref:`what_is_the_plutus_platform`.
+To read more about the Platform, see :ref:`what_is_the_plutus_platform`
```
koslambrou, line 5: The dot (.) probably makes sense here at the end.

koslambrou (approved):

doc/reference/glossary.rst
```diff
@@ -59,7 +59,7 @@ Glossary
 
       This is implemented in the Alonzo hard fork of the Cardano blockchain.
 
-      See :ref:`what_is_a_ledger`.
+      See :ref:`what_is_a_ledger`
```
michaelpj, line 62: why all these changes? is it not grammatical to have a full stop at the end?

michaelpj (approved):

joseph-fajen:
I wanted to remove the periods that followed headings that had question
marks included, to avoid instances of "?." because to my eye it looks a
little odd. If you disagree with that I'll look into it a little further.

doc/index.rst
joseph-fajen: @koslambrou @michaelpj 
I find it strange to see a question mark followed immediately by a period. Logically, I see that it makes sense because the heading being referenced includes a question mark, and a period at the end of the whole sentence is logical. However, it jumped out at me as something that didn't look right to my eye. For reference, have a look at this link: 

https://www.thepunctuationguide.com/terminal-points.html

It states that "The general rule is that a sentence ends with only one terminal punctuation mark."

I think either approach is legitimate. My preference is to use a single terminal punctuation mark. But it's a minor issue.

doc/reference/glossary.rst
michaelpj: An alternative approach would be to use a different link text in that circumstance.

PR #542, opened 2022-06-23
PLT-340: test minting policy issue

sjoerdvisscher:
This doesn't actually test that minting policy scripts are executed when calculating fees, but that would basically come down to writing tests for cardano-api, which I don't want to do unless we are sure something is wrong.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-ledger/src/Ledger/Test.hs
+++ b/plutus-ledger/src/Ledger/Test.hs
@@ -1,0 +1,27 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TemplateHaskell     #-}
+{-# LANGUAGE TypeApplications    #-}
+module Ledger.Test where
+
+import Ledger qualified
+import Plutus.Script.Utils.V1.Typed.Scripts.MonetaryPolicies qualified as MPS
+import Plutus.V1.Ledger.Api (Address, Validator)
+import Plutus.V1.Ledger.Scripts qualified as Ledger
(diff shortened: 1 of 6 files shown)


plutus-ledger/src/Ledger/Test.hs
```diff
@@ -0,0 +1,27 @@
+{-# LANGUAGE DataKinds           #-}
```
koslambrou, line 1: Why not put this file in the `test` directory?

plutus-ledger/src/Ledger/Tx.hs
```diff
@@ -167,7 +167,21 @@ pattern CardanoApiEmulatorEraTx tx <- (getEmulatorEraTx -> tx) where
 {-# COMPLETE CardanoApiEmulatorEraTx #-}
 
 instance Pretty CardanoTx where
-    pretty = onCardanoTx pretty (pretty . getCardanoApiTxId)
+    pretty tx =
```
koslambrou, line 170: Why did you need to modify `Pretty` for this PR?

koslambrou: Ah, was it to also print the minting policies?

koslambrou (approved):

plutus-ledger/src/Ledger/Test.hs
sjoerdvisscher: Because it doesn't contain tests, it contains utilities useful for testing. Just like `Plutus.Contract.Test`. The `test` directories we have are not structured in a way that's useable from another package, they are all the same.

PR #543, opened 2022-06-24
[emulator]: fix waitUntilSlot memory consumption

### LINKED ISSUE #318: Severe scaling issues with `waintUntilSlot` in emulator
### Summary

We have run into some scalability issues with running big QuickCheck tests in the emulator and I have identified at least one clear cause, `Plutus.Trace.Emulator.waitUntilSlot`. Basically, the more slots you wait in this action the more memory the emulator consumes - by a large factor. This happens even if `Trace.waitUntilSlot` _is the only thing you do in your emulator tests!_

The following test code reveals the problem (for different values of `n`):
```
module Test where
import Test.Tasty
import Plutus.Trace.Emulator as Trace
import Plutus.Contract.Test

main = defaultMain $ testGroup "tests" [
  checkPredicate "test"
    (walletFundsChange w1 mempty
    .&&. walletFundsChange w2 mempty
    .&&. walletFundsChange w3 mempty)
          $ void $ Trace.waitUntilSlot n]
```

### Actual Result

We see the following performance when running the code:
```
Wait time (n)  | time to execute test
  31                    0.02s
  310                  0.11s
  3100                0.92s
  31000              93.80s
  310000           {Ran out of memory at 16GB}
```

### Expected Result

We expect the runtime and memory consumption to not depend in any meaningful way on the number of slots waited.

_No response_

### System info

OS: Windows with WSL2
Plutus version: main
Machine: ThinkPad X1 with 11th gen i7 and 32GB internal memory (of which 16GB allocated to WSL2)

ak3n:
Should fix #318.

This reduces the memory consumption from 3gb to under 300mb (approximately) for me with this test:

```
run "test"
  (walletFundsChange w1 mempty
  .&&. walletFundsChange w2 mempty
  .&&. walletFundsChange w3 mempty) (void $ Trace.waitUntilSlot 100000)
```

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-chain-index-core/src/Plutus/ChainIndex/Emulator/Handlers.hs
+++ b/plutus-chain-index-core/src/Plutus/ChainIndex/Emulator/Handlers.hs
@@ -178,14 +178,16 @@ appendBlocks [] = pure ()
 appendBlocks blocks = do
     let
         processBlock (utxoIndexState, txs) (Block tip_ transactions) = do
-            case UtxoState.insert (TxUtxoBalance.fromBlock tip_ (map fst transactions)) utxoIndexState of
-                Left err -> do
-                    let reason = InsertionFailed err
-                    logError $ Err reason
-                    return (utxoIndexState, txs)
-                Right InsertUtxoSuccess{newIndex, insertPosition} -> do
-                    logDebug $ InsertionSuccess tip_ insertPosition
-                    return (newIndex, transactions ++ txs)
+            if null transactions then return (utxoIndexState, txs)
+            else do
+                case UtxoState.insert (TxUtxoBalance.fromBlock tip_ (map fst transactions)) utxoIndexState of
+                    Left err -> do
+                        let reason = InsertionFailed err
+                        logError $ Err reason
+                        return (utxoIndexState, txs)
+                    Right InsertUtxoSuccess{newIndex, insertPosition} -> do
+                        logDebug $ InsertionSuccess tip_ insertPosition
+                        return (newIndex, transactions ++ txs)
     oldState <- get @ChainIndexEmulatorState
     (newIndex, transactions) <- foldM processBlock (view utxoIndex oldState, []) blocks
     put $ oldState
--- a/plutus-contract/src/Plutus/Trace/Scheduler.hs
+++ b/plutus-contract/src/Plutus/Trace/Scheduler.hs
@@ -301,19 +301,14 @@ loop s = do
     case dequeue s of
         AThread EmThread{_continuation, _threadId, _tag} event schedulerState prio -> do
             let mkLog e = SchedulerLog{slEvent=e, slThread=_threadId, slPrio=prio, slTag = _tag}
-            logDebug (mkLog Resumed)
             result <- _continuation event
             case result of
-                Done () -> do
-                    logDebug (mkLog $ Stopped ThreadDone)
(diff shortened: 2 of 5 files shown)


plutus-chain-index-core/src/Plutus/ChainIndex/Emulator/Handlers.hs
```diff
@@ -187,11 +192,14 @@ appendBlocks blocks = do
                     logDebug $ InsertionSuccess tip_ insertPosition
                     return (newIndex, transactions ++ txs)
     oldState <- get @ChainIndexEmulatorState
+    let oldIndex = view utxoIndex oldState
     (newIndex, transactions) <- foldM processBlock (view utxoIndex oldState, []) blocks
-    put $ oldState
-            & set utxoIndex newIndex
-            & over diskState
-                (mappend $ foldMap (\(tx, opt) -> if tpoStoreTx opt then DiskState.fromTx tx else mempty) transactions)
+    -- This check reduces the emulator's memory consumption by 75%
```
sjoerdvisscher, line 197: Is it valid to not store transactions that don't have an effect on the index?

ak3n: I have checked to make sure that we don't miss anything and turned out this is the case when there are no transactions at all. But I suppose adding an if-branch to validate it in code would be a good idea.

plutus-chain-index-core/src/Plutus/ChainIndex/Emulator/Handlers.hs
```diff
@@ -178,20 +179,22 @@ appendBlocks [] = pure ()
...
-            case UtxoState.insert (TxUtxoBalance.fromBlock tip_ (map fst transactions)) utxoIndexState of
-                Left err -> do
-                    let reason = InsertionFailed err
-                    logError $ Err reason
-                    return (utxoIndexState, txs)
-                Right InsertUtxoSuccess{newIndex, insertPosition} -> do
-                    logDebug $ InsertionSuccess tip_ insertPosition
-                    return (newIndex, transactions ++ txs)
+            if null transactions then return (utxoIndexState, txs)
+            else do
```
ak3n, line 183: I have looked at `UtxoState.insert` and I suppose there is no point in adding empty `UtxoState`s to the fingertree. We split the fingertree a lot just to add fresh `UtxoState`s without any transactions.

But I would like a validation of the correctness of this.

raduom: I only see a possible problem with rollbacks.

raduom: I checked and it should be fine.

raduom (approved):

koslambrou:
Would it make sense to add your test as well? Although I don't know if it will be obvious to the developer that modifies the scheduler again that it uses too much memory.

koslambrou (approved):

ak3n:
>Would it make sense to add your test as well? Although I don't know if it will be obvious to the developer that modifies the scheduler again that it uses too much memory.

I thought about that. We might add the test and limit the memory usage of `plutus-contract-test` executable by some upper bound. That would help to track the memory usage and avoid the unexpected growth.

PR #547, opened 2022-06-28
PLT-40 Apply changes from Djed which optimize SQL queries utxoSetAtAddress and utxoSetWithCurrency

koslambrou:
Apply changes from Djed which optimize SQL queries of utxoSetAtAddress and utxoSetWithCurrency chain-index queries.

Co-authored-by: etiennjf <jean-frederic.etienne@iohk.io>

* Syncing until slot 40M in testnet, before this commit, takes 1h. Now, it takes 1h15m.

* Querying the utxo of a given address took, before this commit, 2min. Now it takes around 15ms.

* Tested the query: `curl -H "Content-Type: application/json" -d '{ "credential": { "tag": "PubKeyCredential", "contents": { "getPubKeyHash": "88ff402b0522f27649ac742238c697c579beeb344eb723099d1f16ce" } } }' http://localhost:9083/txo-at-address` which is available in slot 40M of testnet.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-chain-index/src/Plutus/ChainIndex/Lib.hs
```diff
@@ -93,6 +93,12 @@ withRunRequirements logConfig config cont = do
         \                                 AND input_row_out_ref = old.output_row_out_ref; \
         \END"
 
+    -- creating extra index to optimize utxoSetAtAddress and utxoSetWithCurrency queries
```
raduom, line 96: You could also use a `CREATE INDEX IF NOT EXISTS..` but not a big issue.

raduom (approved):

PR #551, opened 2022-06-28
PLT-484 Upgraded cardano-node version to the official 1.35.0 release.…

koslambrou:
Upgraded cardano-node version to the official 1.35.0 release. Also updated cardano-wallet, plutus et al. versions that work with v1.35.0 of the node.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-example/src/PlutusExample/ScriptContextChecker.hs
```diff
@@ -247,3 +254,11 @@ dummyPOSIXTimeRange = Plutus.from $ Plutus.POSIXTime 42
 dummyScriptPurpose :: Maybe Plutus.ScriptPurpose
 dummyScriptPurpose = Nothing
 
+getTxInInfoFromTxIn
+  :: Shelley.UTxO (Alonzo.AlonzoEra StandardCrypto)
+  -> Ledger.TxIn StandardCrypto
+  -> Maybe Plutus.TxInInfo
+getTxInInfoFromTxIn (Shelley.UTxO utxoMap) txIn = do
+  txOut <- Map.lookup txIn utxoMap
+  Alonzo.txInfoIn txIn txOut
```
koslambrou, line 263: @Jimbo4350 Not sure if this is the best approach to get the `TxInInfo` now, since the signature of `Alonzo.txInfoIn` changed. Can you validate?

whs-dot-hk:
Hi, do you need to update this line of `cabal.project`? https://github.com/input-output-hk/plutus-apps/blob/main/cabal.project#L2

koslambrou:
@whs-dot-hk Very good point! Thanks for the reminder :+1:

koslambrou:
@whs-dot-hk Just verified. The current index-state is the correct one. We can't go beyond that for now, because it breaks some of our dependencies like `cardano-prelude` which won't work anymore.

PR #556, opened 2022-07-05
PLT-84: make transaction body auto balance

sjoerdvisscher:
- Implements `makeTransactionBodyAutoBalance` with adjustments from marlowe-cardano
- Implements `MustSpendPubKeyOutput` in plutus-tx-contraints
- Adds testcase that uses both
- Removes old `FeeConfig`
- Refactors cardano-api based code in plutus-ledger

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-ledger/src/Ledger/Fee.hs
```diff
@@ -1,38 +1,78 @@
...
+import Data.Map qualified as Map
+import Ledger.Address (Address, PaymentPubKeyHash)
+import Ledger.Params (EmulatorEra, Params (pNetworkId, pProtocolParams), emulatorEraHistory, emulatorGlobals)
+import Ledger.Tx (ToCardanoError (TxBodyError), Tx)
+import Ledger.Tx.CardanoAPI (CardanoBuildTx (..), getCardanoBuildTx, toCardanoAddressInEra, toCardanoTxBodyContent)
+import Ledger.Validation (CardanoLedgerError, UTxO (..), makeTransactionBody)
+import Ledger.Value (Value)
+import Plutus.V1.Ledger.Ada (lovelaceValueOf)
+
+evaluateTransactionFee
```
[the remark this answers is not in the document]
sjoerdvisscher: Good idea!

koslambrou (approved):

plutus-ledger/src/Ledger/Fee.hs
```diff
@@ -1,38 +1,78 @@
...
-        , fcScriptsFeeFactor :: Double -- ^ Factor by which to multiply the size-dependent scripts fee
-        }
-    deriving (Show, Eq, Generic, ToJSON, FromJSON)
+makeTransactionBodyAutoBalance
+  :: Params
+  -> UTxO EmulatorEra
+  -> CardanoBuildTx
+  -> Address
+  -> Either CardanoLedgerError (C.Api.Tx C.Api.AlonzoEra)
+makeTransactionBodyAutoBalance params utxo (CardanoBuildTx txBodyContent) pChangeAddr = first Right $ do
```
ak3n, line 39: Shouldn't it be `makeTransactionWithAutoBalancedBody`?

Because I read it as "make the transaction body auto balance", but we return the transaction as a result, and make the transaction with auto balanced body.

Just clarifying.

ak3n (approved):

plutus-ledger/src/Ledger/Fee.hs
sjoerdvisscher: Yeah I'm just sticking with the name from cardano-api. I think it is meant to be read as "makeTransactionBody:, the AutoBalance version". I'll add a comment.

sjoerdvisscher: No actually you are right, I forgot I added a makeSignedTransaction call at the end. So `makeAutoBalancedTransaction` it is!

PR #570, opened 2022-07-07
PLT-502 Add locale `LC_ALL=C.utf8` to nix shell

zeme-iohk:
- Enforced `LC_ALL=C.utf8` in the nix-shell to avoid encoding problems on hosts with non-utf8 locales.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -150,6 +150,8 @@ haskell.project.shellFor {
   withHoogle = false;
 
   shellHook = ''
+    export LC_ALL=C.utf8
+    
     ${pre-commit-check.shellHook}
   ''
   # Work around https://github.com/NixOS/nix/issues/3345, which makes
--- a/shell.nix
+++ b/shell.nix
@@ -6,7 +6,7 @@
 }:
 let
   inherit (packages) pkgs plutus-apps plutus-playground pab-nami-demo docs webCommon;
-  inherit (pkgs) stdenv lib utillinux python3 nixpkgs-fmt;
+  inherit (pkgs) stdenv lib utillinux python3 nixpkgs-fmt glibcLocales;
   inherit (plutus-apps) haskell stylish-haskell sphinxcontrib-haddock sphinx-markdown-tables sphinxemoji nix-pre-commit-hooks cabal-fmt;
 
   # Feed cardano-wallet, cardano-cli & cardano-node to our shell.
--- a/shell.nix
+++ b/shell.nix
@@ -150,8 +150,6 @@ haskell.project.shellFor {
   withHoogle = false;
 
   shellHook = ''
-    export LC_ALL=C.utf8
-    
     ${pre-commit-check.shellHook}
   ''
   # Work around https://github.com/NixOS/nix/issues/3345, which makes
--- a/shell.nix
+++ b/shell.nix
@@ -165,4 +165,10 @@ haskell.project.shellFor {
   + ''
     export WEB_COMMON_SRC=${webCommon.cleanSrc}
   '';
+
+  # This is no longer set automatically as of more recent `haskell.nix` revisions,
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -150,6 +150,8 @@ haskell.project.shellFor {
   withHoogle = false;
 
   shellHook = ''
+    export LC_ALL=C.utf8
```
zeme-iohk, line 153: This looks harmless but might have subtle effects, as it kind of affects *everything* inside the shell.
Inside the nix-shell we only have `C`, `C.utf8` and `POSIX` locales.
Some users (for example those with `LC_CTYPE="en_AU.UTF-8"`) will not be able to run `cabal test plutus-contract` as it will break when printing non-ascii chars to stdout.
Adding this line to the shell fixes the problem, but I'm a little concerned it might blow up in our faces at some point.
What do you think? @michaelpj @angerman @koslambrou

michaelpj: Are you maybe missing this line? https://github.com/input-output-hk/plutus/blob/master/shell.nix#L113

This is kind of my fault - `haskell.nix` used to set this stuff in the shell automatically, but I removed it because no other nixpkgs shells set this stuff, we usually leave it to the user. But it seems to be needed for a bunch of Haskell stuff.

michaelpj: To be clear, I think including the locale archives should remove the need for this setting. Overriding the user's locale can be annoying for them.

zeme-iohk: Aaaa so `plutus` has it but `plutus-apps` does not. Let's try that. @andreabedini could you test out this branch and tell me if it works?

andreabedini: @zeme-iohk yes I will test. Adding the missing locale information sounds like a clearner fix, I wasn't aware that was a solved problem already.

andreabedini:
I tested this branch with `nix-shell` and `cabal test plutus-contract` no locale related failures :tada: but some `ScriptFailures`

```
Transactions failed to validate:
      +++ OK, passed 1 test.
686f605fd4e9b11d6eba0358471b2db00cd6e15da3383486a1a1aaabb3a08e60: ScriptFailure (EvaluationError ["PT8"] "CekEvaluationFailure")
      
      Actions (1 in total):
      100% FailFalse
Test failed.
    Failure due to head [] not allowed:                               Emulator log:
OK (0.01s)
      +++ OK, failed as expected. Assertion failed (after 1 test):
    Division by zero not allowed:                                     Transactions failed to validate:
7b90949beab21b05cce94704d227704870d9c8cbdfee81e124324937ed67a10f: ScriptFailure (EvaluationError [] "CekEvaluationFailure")
Test failed.
Emulator log:
OK (0.01s)
      +++ OK, failed as expected. Assertion failed (after 1 test):
    Can't trick division by zero check using trace:                   Chain events do not match predicate
TxnValidate 331170b4130e90cf688d52a395fe1a712c305f343587547d47ea1dd8071e17a6
SlotAdd Slot 1
TxnValidate 8c17eafeba739872dd9c94abcd62ba073277cd158ac34d1f00a49f56b223fc93
SlotAdd Slot 2
TxnValidationFail Phase2 e15c2abb35360f4f8746591bed75cc7680f33e497ebcf320ec9ae706f6656d4a: ScriptFailure (EvaluationError ["PT5"] "BuiltinEvaluationFailure of DivideInteger")
SlotAdd Slot 3
SlotAdd Slot 4
Transactions failed to validate:
e15c2abb35360f4f8746591bed75cc7680f33e497ebcf320ec9ae706f6656d4a: ScriptFailure (EvaluationError ["PT5"] "BuiltinEvaluationFailure of DivideInteger")
```

I am happy to take the locale problem has solved!

andreabedini (approved):
LGTM and also seems to work :)

PR #575, opened 2022-07-08
Reorganized the dependencies (build-depends) in the cabal files

koslambrou:
The dependencies in the cabal files are separated in three categories:
- Local components
- Other IOG dependencies
- Non-IOG dependencies

Feel free to propose different categories

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/doc/plutus-doc.cabal
+++ b/doc/plutus-doc.cabal
@@ -39,7 +39,7 @@ common lang
     ScopedTypeVariables
     StandaloneDeriving
 
-  -- See Plutus Tx readme for why we need the following flags: 
+  -- See Plutus Tx readme for why we need the following flags:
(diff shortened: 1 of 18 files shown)


ak3n (approved):
Love it!

One note. What do you think about using common stanzas to explicitly name them?

```
common local_packages
  build-depends: plutus-ledger
  ghc-options: -Wall

library
  import: local_packages
  exposed-modules: Foo
```

michaelpj:
> One note. What do you think about using common stanzas to explicitly name them?

I think this doesn't scale as well when you have multiple components in one cabal file, which we often do.

michaelpj (approved):
Looks nice!

PR #582, opened 2022-07-11
Make `getCardanoTxInputs` return `TxInType`

ak3n:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


ak3n:
Subsumed by #594. 

We should consider to remove the `TxInType` at all as it seems it's impossible to recover the information we need from the transaction alone.

PR #585, opened 2022-07-12
Raduom/marconi initiative

raduom:
Basically put down some thoughts on why Marconi makes sense.

PR: https://input-output.atlassian.net/jira/software/c/projects/PLT/boards/788?modal=detail&selectedIssue=PLT-503

This is an ADR.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
+=========================
+Making a case for Marconi
+=========================
+
+Plutus off-chain code oftentimes needs access to indexed portions of the blockchain. The Chain Index project is the initial solution meant to deliver access to this kind of data. However, after release, a couple of shortcomings were identified which prompted the development of an indexing solution that is based on a different set of architectural and functional constraints.
```
andreabedini, line 5:
```suggestion
Plutus off-chain code oftentimes needs to have access to data thats has been already stored on-chain. The Chain Index project is a solution to deliver access to this data by indexing portions of the blockchain. However, after release, few shortcomings were identified which prompted the development of an indexing solution based on different architectural and functional constraints.
```

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+  
+  B. The monolithic architecture lacks the features necessary to customise the indexed set of data, the Chain Index provides only all-or-nothing indexing. While this can be addressed the monolithic architecture makes it an uphill battle. 
+  
+  C. Also due to the monolithic architecture that assumes that there is only one index running we ran into trouble when we made the Plutus Application Backend collect and index information requested by smart contracts. Now we have two components that index information from the blockchain, but they are not synchronised. Querying the Chain Index about transactions received from the Plutus Application Backend often results in no data returned, since the Chain Index indexes data slower than the PAB.
```
andreabedini, line 15: To me this sounds just like a PAB bug which can be solved without a re-architecture. Like point B above.

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+  
+  D. The lack of non-functional requirements resulted in software that uses an unreasonable amount of resources and results in slow synchronisation speeds. And since everything is monolithic it is difficult to turn off indexing of data which is not required by our customers there is no way to limit the required resources.
```
andreabedini, line 17: Sounds a bit like point B but `lack of non-functional requirements` is importan on its own. Perhaps merge with point E below?

e.g.

```
B. The monolithic architecture lacks the features necessary to customise the indexed set of data, the Chain Index provides only all-or-nothing indexing. Without changing the code, it impossible to turn off indexing of data that is not required by our customers. While this can be addressed in principle, the monolithic architecture makes it an uphill battle.

E. The lack of non-functional requirements resulted in software that uses an unreasonable amount of resources and results in slow synchronisation speeds. Combined with the lack of a specification this makes the testing feel ad-hoc and like an afterthought.
```

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+
+Having a set of property-based tests to validate that the implementation conforms to the specification also means that the correctness of the implementation does not rely on type-level checks or complicated term-level machinery (we could even verify the correctness of a Rust implementation by leveraging the Rust to Haskell FFI).
+
+Because we have no reliance on type-level checks or complicated architectural patterns to validate the software (we use the specification and property tests for that), the code is much easier to understand, document and extend.
+
+=================
+Indexing solution
+=================
+
+The indexing solution has the following basic requirements: it needs to deal with rollbacks as elegantly as possible and provide a way to compromise between memory, disk and CPU usage.
```
andreabedini, line 43:
```suggestion
Any indexing solution has to deal with blockchain rollbacks, which can happen as the network achieves consensus. We introduce the following basic requirements: it needs to deal with rollbacks as elegantly as possible and provide a way to compromise between memory, disk and CPU usage.
```

Expanding a bit just because you haven't mentioned rollbacks before.

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+
+On the Cardano blockchain, there are frequent rollbacks, but they can only span a maximum of 2160 blocks (and most of them are < 10 blocks). We call the 2160 number the security parameter K (and we denote it by 'K' henceforth).
+
+Indexers are a store which is updated by events created from each block. The problem introduced by rollbacks is that we need to undo all state changes when a rollback occurs.
+
+We opted for a design where we keep K blocks in memory as the list of events that are fed into the function that merges them into the persisted state and persist them once they go beyond the K limit.
```
andreabedini, line 49: What is "them"? You save the persisted state to disk past the security parameter right? not the blocks themselves.

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+
+This architectural decision has some desirable effects:
+
+  1. Managing rollbacks is very simple and fast. We drop the events that were rolled back. (no complicated logic is required to undo the projection of the list of events on disk, which we would need if we would want to store everything on disk as fast as possible).
```
andreabedini, line 53:
```suggestion
  1. Managing rollbacks is very simple and fast. We drop the events that were rolled back. (no need to undo the application of blocks on the state stored on disk, which would be necessary if we were to store everything on disk as fast as possible).
```

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+
+Indexers are a store which is updated by events created from each block. The problem introduced by rollbacks is that we need to undo all state changes when a rollback occurs.
+
+We opted for a design where we keep K blocks in memory as the list of events that are fed into the function that merges them into the persisted state and persist them once they go beyond the K limit.
+
+This architectural decision has some desirable effects:
+
+  1. Managing rollbacks is very simple and fast. We drop the events that were rolled back. (no complicated logic is required to undo the projection of the list of events on disk, which we would need if we would want to store everything on disk as fast as possible). 
+  
+  2. Making 'K' configurable makes the design already quite scalable. Developers do not usually need to guard themselves against rollbacks by K blocks so they can choose to only store 10 events in memory thus sacrificing chain integrity for RAM.
```
andreabedini, line 55: > thus sacrificing chain integrity for RAM

I think this point is very important but "sacrificing chain integrity" with no extra detail sounds a bit scary. Can you expand?

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+  
+  3. In case of a restart recovery is very simple. We only store fully confirmed transactions so there is nothing to do other than resume operation.
```
andreabedini, line 57:
```suggestion
  3. In case of a restart recovery is very simple. We only store fully confirmed transactions so there is nothing to do other than loading the state persisted on disk and following the blockchain from the corresponding chain-point.
```

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+
+And some less desirable effects:
+
+  1. We must keep K events in memory, which (depending on how large events are) can waste some memory. Our educated guess is that this is a reasonable compromise, but depending on how large events can get that may not be the case for your use case.
```
andreabedini, line 61: > can waste some memory

I don't think it's necessarily a "waste".

```suggestion
  1. We must keep K events in memory, which (depending on how large events are) can waste some memory. Our educated guess is that this is a reasonable compromise, but depending on how large events can get that may not be the case for every use case.
```

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+  
+  2. Queries are more complicated as we need to query both events in memory and the state persisted on disk.
+
+=================
+Query and storage
+=================
+
+The indexed data is accessible through queries. There are no constraints on the format of queries or results. Both are identified by a type variable that the indexer exposes and the code for both of them has to be provided by the user.
+
+This is currently the only area where the users can customise the indexer by providing functions that handle queries and storage. Note that the queries have to run on both the in-memory data and the data that is stored.
```
andreabedini, line 71: We definitely need a specification for this API.

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+
+This allows us to associate any kind of storage type to the indexers, though, right now we are only using SQLite.
+
+==================
+Versioned contents
+==================
+
+We need some sort of versioning or sequencing of events, notifications and queries.
```
andreabedini, line 79: I am not sure what you mean by this.

andreabedini (commented):
This is good. I allowed myself to reword some parts but feel free to take what you like and discard the rest. I commented on a couple of points that are unclear to me.

doc/adr/0004-marconi-initiative.rst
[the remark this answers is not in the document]
joseph-fajen: Suggestion: 
"to have access to data that has already been stored..." 
"However, after release, a few shortcomings..."

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+
+A lot of the shortcomings are connected to the exploratory type of development that we used to deliver the Chain Index which was prompted by the lack of a clear specification and a lack of concern for non-functional and quality assurance requirements. The top-down design resulted in a monolithic and fairly complex architecture which made the code difficult to reuse, compose and understand.
+
+Some of the problems we identified due to the above-mentioned approaches are:
+
+  A. The use of free algebras (the `freer-simple` package) makes the code fairly complex and difficult to understand (quite a few type-level computations are happening). The separation between syntax and semantics imposed by the library also complicates matters for no clear reason (for example, if we write two semantics, one for pure code used for testing and one for production code, then there would be a lot of production code that would not be tested).
```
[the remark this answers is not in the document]
joseph-fajen: As a non-subject matter expert reader in this area, I don't know the meaning of the term "free algebras," but I imagine it is possible that the target audience may be expected to be familiar with this term.

doc/adr/0004-marconi-initiative.rst
joseph-fajen: Suggestion to add a couple of commas: 
Line 13: "While this can be addressed, the monolithic..." 
Line 15: "...that assumes that there is only one index running, we ran into..."

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+  
+  D. The lack of non-functional requirements resulted in software that uses an unreasonable amount of resources and results in slow synchronisation speeds. And since everything is monolithic it is difficult to turn off indexing of data which is not required by our customers there is no way to limit the required resources.
+  
+  E. The same lack of a specification and non-functional requirements makes the testing feel ad-hoc and like an afterthought.
+
+The Chain Index was meant to be a software application that supports the execution of smart contracts. And, in that, it succeeded. However, we found out that our customers would rather have a library of functionality that they can customise to build their indexers, only for data they care about for their application, using whatever storage engine that they prefer and supporting only the queries that they need to support.
```
joseph-fajen, line 21: The Chain Index was meant to be a software application that supports the execution of smart contracts. And, in that, it succeeded. However, we found that our customers would rather have a library of functionality that they can customize to do the following:
* to build their own indexers, 
* to work only with the data that they care about for their application, 
* to use whatever storage engine they prefer, and 
* to support only the queries that they need to support.

doc/adr/0004-marconi-initiative.rst
joseph-fajen: Suggestion: Use a capital "N" --> (No need to undo the application... )

doc/adr/0004-marconi-initiative.rst
joseph-fajen: Suggestion: reverse order of "only" and "store" --> 
"...so they can choose to store only 10 events in memory..."

doc/adr/0004-marconi-initiative.rst
joseph-fajen: Suggest adding a comma: 
"In case of a restart, recovery is very simple."

doc/adr/0004-marconi-initiative.rst
joseph-fajen: Suggestion: 
"Our educated guess is that this is a reasonable compromise; however, depending on how large events can get, that may not be..."

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+  2. Resume functionality (we need to know from which slot to resume)
+  3. Handling of rollbacks (now there is explicit handling of rollbacks)
+
+More information will become available in the next few sprints.
+
+=============
+Event streams
+=============
+
+To support PAB functionality which subscribes to a source for a set of event types we need a way to produce events from indexers.
```
joseph-fajen, line 91: Suggest adding a comma: 
"...for a set of event types, we need a way..."

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+
+They are also very useful for contracts that want to track rollbacks. Rollbacks are invisible from the point of view of the indexed data, but it may be the case that the internal state of a contract needs to know that the state has been reverted.
+
+============
+Similar work
+============
+
+One of the advantages that Marconi has over Oura and Scrolls is the fact that they are both a streaming solution and an indexer. So you can listen to a stream of events and know that those events are reflected in the index.
```
joseph-fajen, line 99: I'm slightly unclear on the meaning. Can you clarify a little bit? Is this saying that Marconi is both a streaming solution and an indexer? I'm not sure what the word "they" is referring to --> "they are both a streaming solution and an indexer" 

I like this document! Very informative.

joseph-fajen (approved):
Radu, I left some comments. I found this to be a very informative document!

doc/adr/0004-marconi-initiative.rst
koslambrou, line 5:
```suggestion
Plutus off-chain code oftentimes needs access to indexed portions of the blockchain. The plutus-chain-index project is the initial solution meant to deliver access to this kind of data. However, after release, a couple of shortcomings were identified which prompted the development of an indexing solution that is based on a different set of architectural and functional constraints.
```

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
+=========================
+Making a case for Marconi
+=========================
+
+Plutus off-chain code oftentimes needs access to indexed portions of the blockchain. The Chain Index project is the initial solution meant to deliver access to this kind of data. However, after release, a couple of shortcomings were identified which prompted the development of an indexing solution that is based on a different set of architectural and functional constraints.
+
+A lot of the shortcomings are connected to the exploratory type of development that we used to deliver the Chain Index which was prompted by the lack of a clear specification and a lack of concern for non-functional and quality assurance requirements. The top-down design resulted in a monolithic and fairly complex architecture which made the code difficult to reuse, compose and understand.
```
koslambrou, line 7:
```suggestion
A lot of the shortcomings are connected to the exploratory type of development that we used to deliver the plutus-chain-index which was prompted by the lack of a clear specification and a lack of concern for non-functional and quality assurance requirements. The top-down design resulted in a monolithic and fairly complex architecture which made the code difficult to reuse, compose and understand.
```

doc/adr/0004-marconi-initiative.rst
koslambrou, line 11: I understand the point, but I don't think I'm in agreement. The use of an effect system in `plutus-chain-index` was to, like you said, have two implementations: one for the emulator and one for production use. While I agree with the point that "there would be a lot of production code that would not be tested", saying that the effect system "makes the code fairly complex and difficult to understand" is too opinionated. I believe we could have a similar problem (havingcomplex code) even if we tried to reproduce the same behavior with `mtl` instead of `freer-simple` (there's even more boilerplate with `mtl` IMO).

doc/adr/0004-marconi-initiative.rst
koslambrou: To continue @andreabedini's point, this is a PAB architectural bug which is addresses here: #550 . Once the PR is merged I'll reference it to this ADR.

koslambrou, line 15:
```suggestion
  B. The monolithic architecture lacks the features necessary to customise the indexed set of data, the Chain Index provides only all-or-nothing indexing. While this can be addressed the monolithic architecture makes it an uphill battle. 
  
  C. Also due to the monolithic architecture that assumes that there is only one index running we ran into trouble when we made the Plutus Application Backend collect and index information requested by smart contracts. Now we have two components that index information from the blockchain, but they are not synchronised. For example, querying the Chain Index about a transaction, from which the PAB confirmed it's presence in a Cardano block, often results in no data returned, since the Chain Index indexes data slower than the PAB.
```

doc/adr/0004-marconi-initiative.rst
koslambrou, line 21: Well. "Succeeded" is probably not the right term. I would say that functional requirement were met, but non-functional requirements were not adequately thought about, which resulted in a bad experience for users.

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+
+So when we took all the feedback into account we decided that a redesign of the indexing solution using a much simplified and modular design is a worthwhile enterprise.
+
+We continue by introducing some of the design principles that guided us in the specification of Marconi.
+
+============================
+Design principles of Marconi
+============================
+
+I follow the Algebra Driven Design approach for Marconi components, so from the get-go, we will have a checked specification for the software that we develop.
```
koslambrou, line 31: I -> We?

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+
+Because we have no reliance on type-level checks or complicated architectural patterns to validate the software (we use the specification and property tests for that), the code is much easier to understand, document and extend.
+
+=================
+Indexing solution
+=================
+
+The indexing solution has the following basic requirements: it needs to deal with rollbacks as elegantly as possible and provide a way to compromise between memory, disk and CPU usage.
+
+On the Cardano blockchain, there are frequent rollbacks, but they can only span a maximum of 2160 blocks (and most of them are < 10 blocks). We call the 2160 number the security parameter K (and we denote it by 'K' henceforth).
```
koslambrou, line 45: We need to reconfirm this fact. I read an important detail about the K parameter. I always though that it meant that once a block is K blocks deep in the blockchain, then it is immutable. However, that is not the case. The real formula to determine how deep a block should be in order for it to be immutable is something like (I don't remember it exactly): 2 * K / activeSlotCoef. I'll need to research this more.

doc/adr/0004-marconi-initiative.rst
koslambrou: There's an ADR coming for this API in the next sprint probably. Right @raduom ?

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+
+This allows us to associate any kind of storage type to the indexers, though, right now we are only using SQLite.
+
+==================
+Versioned contents
+==================
+
+We need some sort of versioning or sequencing of events, notifications and queries.
+
+  1. Synchronisation of multiple indexers (queries have a validity interval)
```
koslambrou, line 81: Link to the other ADR once available.

koslambrou:
@raduom This ADR will probably get a lot of feedback/comments. I would suggest:

* to keep this PR open for some sufficient amount of time in order to allow everyone time to put in their comments
* mark it as draft in order to not merge it yet

What do you think?

doc/adr/0004-marconi-initiative.rst
koslambrou, line 99: This section should (maybe? probably?) be expanded.

koslambrou, line 99: Links to Oura and Scrolls would be helpful.

doc/adr/0004-marconi-initiative.rst
joseph-fajen: Question about line 69: "...the code for both of them has to be provided by the user." Can you clarify what "both of them" is referring to? Is it the queries and the type variable? Or the results?

doc/adr/0004-marconi-initiative.rst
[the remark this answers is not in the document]
raduom: You never rollback things that are on-disk. That is why the indexers are fairly simple.

raduom:
> 

Done.

raduom:
> * mark it as draft in order to not merge it yet

I have marked it as draft.

doc/adr/0004-marconi-initiative.rst
raduom: I have this slight problem which is the slot number changes which may add some updates to this document and I also don't know very well how oura/scrolls work.

doc/adr/0004-marconi-initiative.rst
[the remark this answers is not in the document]
koslambrou: Oh! Then with my current understand, I would heavily object to this decision. Here's my perspective:

If you never rollback things that are on-disk, then it makes no sense to give the user the option of defining how many blocks are kept in memory. You really don't want to affect the integrity of the on-disk data.

IF we want to give the user the option to specify the amount of blocks in memory (like 100 for example), then we definitely **need** to handle rollbacks on data that is on disk. Customizability is one of the main key points of Marconi. Doesn't made sense to break chain data integrity we need to  rollback, let's say, 500 blocks deep. Plus, I though the indexer (hysterical-screams) was meant to make the process of implementing rollbacks simple to handle between data that is on disk and data that is on memory.

doc/adr/0004-marconi-initiative.rst
koslambrou: Then, we can wait for Joseph's work which compares the different tools.

doc/adr/0004-marconi-initiative.rst
raduom: @andreabedini I am not really convinced about C not being an architectural bug. You can sort of solve it if your indexing solution fits into one machine (and note, that it may need to fit into a developer's machine, not necessarily a server), but if it does not, then you will need some kind of distribution anyway, in which case life gets complicated for chain-index.

doc/adr/0004-marconi-initiative.rst
raduom: B is about the inability of splitting up indexing into several components that can be optionally turned off, whereas E (which is the point I suppose you mean) is about the lack of a testing strategy.

doc/adr/0004-marconi-initiative.rst
andreabedini: I think we can move the discussion to #550.

doc/adr/0004-marconi-initiative.rst
raduom: > We definitely need a specification for this API

Is the type signature of the required functions what you had in mind?

raduom: > Question about line 69: "...the code for both of them has to be provided by the user." Can you clarify what "both of them" is referring to? Is it the queries and the type variable? Or the results?

I clarified (I hope) the meaning.

doc/adr/0004-marconi-initiative.rst
raduom: I rewrote it. Hopefully it's a little more clear now.

doc/adr/0004-marconi-initiative.rst
```diff
@@ -0,0 +1,100 @@
...
+  
+  2. Making 'K' configurable makes the design already quite scalable. Developers do not usually need to guard themselves against rollbacks by K blocks so they can choose to only store 10 events in memory thus sacrificing chain integrity for RAM.
+  
+  3. In case of a restart recovery is very simple. We only store fully confirmed transactions so there is nothing to do other than resume operation.
+
+And some less desirable effects:
+
+  1. We must keep K events in memory, which (depending on how large events are) can waste some memory. Our educated guess is that this is a reasonable compromise, but depending on how large events can get that may not be the case for your use case.
+  
+  2. Queries are more complicated as we need to query both events in memory and the state persisted on disk.
```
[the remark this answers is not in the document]
raduom: I removed both and replaced the word `query` with `scan`.

doc/adr/0004-marconi-initiative.rst
raduom: I need to rewrite this, but I think I need a little bit of time to look over the competing solutions to make sure I am not assuming too many things about them.

doc/adr/0004-marconi-initiative.rst
raduom: I renamed all the `Chain Index` strings to `plutus-chain-index`. I think it looks better.

doc/adr/0004-marconi-initiative.rst
raduom: I am not the only person that said that our code is "fairly complex and difficult to understand". At least one of our clients pointed that to us in a previous occasion. I was referring to the type-level programming here, which is not so prevalent in a MTL based solution. It was not about boilerplate as much as complexity.

doc/adr/0004-marconi-initiative.rst
raduom: What I meant here is that it was usable. And it still is usable. It can be faster, more modular, etc. But it works, which is a great achievement given the lack of specification and complexity of the problem, which is something I wanted to give props for.

doc/adr/0004-marconi-initiative.rst
raduom: As long as there is a formula, there is a K.

doc/adr/0004-marconi-initiative.rst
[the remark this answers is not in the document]
raduom: Updated no. 3.

doc/adr/0004-marconi-initiative.rst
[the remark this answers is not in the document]
raduom: > If you never rollback things that are on-disk, then it makes no sense to give the user the option of defining how many blocks are kept in memory. You really don't want to affect the integrity of the on-disk data.

I would not mind giving this option to developers working on machines that have very little ram, however for production-level indexers it would be pretty bad.

The way we did it before is that we hid this option under a 'development options' header. We could have a development-mode run configuration which allows setting K < 2160 and a production-mode run configuration that only allows for K > 2160.

As a different alternative, we can always complicate the indexers to handle on-disk rollbacks and provide that as an option, in case users don't mind a bit more complexity.

joseph-fajen (approved):
Looks like some helpful clarifications here.

koslambrou:
> I do think we need further discussion on handling rollbacks greater than K

@asutherlandus After our discussion with Mamba, it seems like there's no need to handle rollbacks greater than K, because there won't be any. Once the block is K (2160 on the current mainnet) blocks deep (with an average of 1 block every 20s), the block should be immutable. That's what I understood from the conversation.

andreabedini (approved):
LGTM

PR #586, opened 2022-07-12
ADR-0004 Common PAB API document

gege251:
See the end result here: https://plutus-apps--586.org.readthedocs.build/en/586/adr/0004-common-pab-api.html

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


doc/adr/0004-common-pab-api.rst
```diff
@@ -0,0 +1,44 @@
...
+
+The PAB Contract (defined in `plutus-apps`) is using the `freer-simple` effect system to define all the contract effects. This already allows us to separate the interface from the implementation, or to have multiple implementations for one interface.
+
+Our proposal, is to further separate these two components: create a separate repository with these contract effects and types, but without any implementation or any unnecessary dependencies. By moving the contracts out of the plutus-apps monorepository, any tool could update to newer version to their discretion. Without many dependencies, many tools could utilize the contract API without having to depend on the whole plutus-apps monorepo.
+
+The current implementation of Contract effects are tightly coupled to the PAB and chain-index implementations, so we also need to make changes in order to get a more general API, but this could be done in a further iteration.
+
+Architectural decisions
+-----------------------
+TBD
```
gege251, line 43: I left this part empty, because I think we need to spend some time to find which parts of the current implementation needs to be refactored.

doc/adr/0004-common-pab-api.rst
```diff
@@ -0,0 +1,44 @@
...
+
+A standard API for all Plutus contacts would help keeping the ecosystem on the same track with their implementation. As more and more off-chain tools implement the same contract interface in the future, it will be relatively easy to switch between different PAB implementations, or to use multiple of these tools at the same time without a need for serious code rewrites.
+
+Proposal
+--------
+
+The PAB Contract (defined in `plutus-apps`) is using the `freer-simple` effect system to define all the contract effects. This already allows us to separate the interface from the implementation, or to have multiple implementations for one interface.
+
+Our proposal, is to further separate these two components: create a separate repository with these contract effects and types, but without any implementation or any unnecessary dependencies. By moving the contracts out of the plutus-apps monorepository, any tool could update to newer version to their discretion. Without many dependencies, many tools could utilize the contract API without having to depend on the whole plutus-apps monorepo.
```
sjoerdvisscher, line 37: I think having the contract effects and types in a separate package should be enough right? I don't see the need for a separate repository.

ak3n: The disadvantage of this is that cabal will have to clone the whole repo to use the package (https://github.com/haskell/cabal/issues/7264).

sjoerdvisscher: That's a downside that seems pretty minor to us not having a monorepo anymore.

gege251: Another downside is that PAB will not be able to use a different version of the interface. Tools, that only use the interface would have to wait for the PAB implementation. And the reality is that the PAB implementation is lagging behind the hardfork, and we're unsure when we will be able to use Vasil features (this is the main reason for this proposal).

gege251: Ideally, the interface would be updated well before a testnet hardfork, so library developers can start building on it.

doc/adr/0004-common-pab-api.rst
```diff
@@ -0,0 +1,63 @@
...
+
+* We will create effects for the constraints-based transaction builder library (`plutus-apps/plutus-ledger-constraints`) in the Contract API.
+  Currently, the interface and the implementation in the transaction builder library are tightly coupled.
+  Therefore, we need to decouple them.
+
+* We will create a separate repository with the contract effects and types (the splitted `plutus-contract`).
+  By moving the Contract API out of the plutus-apps monorepository, any tool could update to newer version to their discretion.
+  Without many dependencies, many tools could utilize the Contract API without having to depend on the whole plutus-apps monorepo.
+
+* TODO: What about governance? Maybe IOG will still be the maintainer of this API for the time being. Then, we can create a new ADR where we make the decision to make this a community driven project.
```
koslambrou, line 53:
```suggestion
* We (IO Global) will continue to be the main maintainers of this new repository. However, a new ADR will need to be created if we ever decide to make this a community driven project.

* TODO: What about governance? How do we decide which interface changes are accepted? ADRs? Who ultimately accepts and rejects them?
```

doc/adr/0004-common-pab-api.rst
sjoerdvisscher: Right, we could do that with a separate branch, like we do now with the `next-node` branch, but that's a bit of a kludge.

TotallyNotChase: It'd also allow this specific "interface-only" package not depend on a great amount of packages that may also need to be updated for it to work. A separate repo will ensure the interface package is truly minimal, easy to depend on, and implement an interpreter for. Of course, it'll also make this interface-only repo much easier to maintain.

koslambrou (approved):

PR #588, opened 2022-07-12
Add script equivalence context test for the V2 context.

Jimbo4350:
Note:
- The mint field is still not tested
- Withdrawals are not tested
- When we feed a validity range via the redeemer we get failures when we try to compare for equality ((==) and (/=)) with the validity interval in the script context.

### DIFF
--- a/plutus-example/src/PlutusExample/ScriptContextChecker.hs
+++ b/plutus-example/src/PlutusExample/ScriptContextChecker.hs
@@ -4,7 +4,9 @@
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
 {-# LANGUAGE TypeFamilies        #-}
+
 
 module PlutusExample.ScriptContextChecker where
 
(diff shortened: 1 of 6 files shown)


koslambrou (approved):

plutus-example/src/PlutusExample/ScriptContextChecker.hs
```diff
@@ -108,41 +124,79 @@ createAnyCustomRedeemer sbe pparams utxo eInfo sStart (ShelleyTx ShelleyBasedEra
...
+      Alonzo.TxDats datumHashMap = getField @"txdats" $ getField @"wits" ledgerTx
+      wdrwls = getField @"wdrls" txBody
+      txwit = getField @"wits" ledgerTx
+      Alonzo.Redeemers rdmrs = getField @"txrdmrs" txwit
+      rdmrList = Map.toList rdmrs
+
+      bUtxo = toLedgerUTxO ShelleyBasedEraBabbage utxo
+      -- Plutus script context types
+  case pScriptVer of
+    PlutusScriptV1 -> Prelude.error "createAnyCustomRedeemer: PlutusScriptV1 custom redeemer not wired up yet"
```
james-iohk, line 153: V1 was previously working. Should this be reimplemented before merging?

PR #594, opened 2022-07-14
Replace `Set TxIn` with `[TxIn]` to fix `StateMachine.getInput`

ak3n:
This PR tries to fix the problem with `StateMachine.getInput` by properly querying the redeemer of the input by its index.

---

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-chain-index-core/src/Plutus/ChainIndex/Tx.hs
+++ b/plutus-chain-index-core/src/Plutus/ChainIndex/Tx.hs
@@ -18,6 +18,7 @@ module Plutus.ChainIndex.Tx(
     , txOutsWithRef
     , txOutRefMap
     , txOutRefMapForAddr
+    , txRedeemersWithHash
     -- ** Lenses
     , citxTxId
     , citxInputs
@@ -38,13 +39,13 @@ import Data.Set qualified as Set
 import Data.Tuple (swap)
 import Ledger (Address, OnChainTx (..), SomeCardanoApiTx (SomeTx), Tx (..), TxIn (..), TxInType (..),
                TxOut (txOutAddress), TxOutRef (..), onCardanoTx, txId)
+import Plutus.ChainIndex.Types
 import Plutus.Contract.CardanoAPI (fromCardanoTx, setValidity)
 import Plutus.Script.Utils.V1.Scripts (datumHash, mintingPolicyHash, redeemerHash, validatorHash)
 import Plutus.V1.Ledger.Scripts (Datum, DatumHash, MintingPolicy (getMintingPolicy),
-                                 MintingPolicyHash (MintingPolicyHash), Redeemer (..), RedeemerHash, Script,
-                                 ScriptHash (..), Validator (getValidator), ValidatorHash (ValidatorHash))
-
-import Plutus.ChainIndex.Types
+                                 MintingPolicyHash (MintingPolicyHash), Redeemer, RedeemerHash, Script, ScriptHash (..),
+                                 Validator (getValidator), ValidatorHash (ValidatorHash))
+import Plutus.V1.Ledger.Tx (RedeemerPtr (RedeemerPtr), Redeemers, ScriptTag (Spend))
 
 -- | Get tx output references from tx.
 txOutRefs :: ChainIndexTx -> [TxOutRef]
@@ -78,7 +79,7 @@ fromOnChainTx = \case
                 let (validatorHashes, otherDataHashes, redeemers) = validators txInputs in
                 ChainIndexTx
                     { _citxTxId = txId tx
-                    , _citxInputs = txInputs
+                    , _citxInputs = Set.toList txInputs
                     , _citxOutputs = ValidTx txOutputs
                     , _citxValidRange = txValidRange
                     , _citxData = txData <> otherDataHashes
@@ -94,7 +95,7 @@ fromOnChainTx = \case
                 let (validatorHashes, otherDataHashes, redeemers) = validators txInputs in
                 ChainIndexTx
                     { _citxTxId = txId tx
-                    , _citxInputs = txCollateral
+                    , _citxInputs = Set.toList txCollateral
                     , _citxOutputs = InvalidTx
                     , _citxValidRange = txValidRange
                     , _citxData = txData <> otherDataHashes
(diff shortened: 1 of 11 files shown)


plutus-chain-index-core/src/Plutus/ChainIndex/Tx.hs
```diff
@@ -78,7 +79,7 @@ fromOnChainTx = \case
                 let (validatorHashes, otherDataHashes, redeemers) = validators txInputs in
                 ChainIndexTx
                     { _citxTxId = txId tx
-                    , _citxInputs = txInputs
+                    , _citxInputs = Set.toList txInputs
```
ak3n, line 82: This is incorrect but to fix it we need to fix the `Tx` from `plutus-core`. Or create our own `Tx` with `[TxIn]` instead of `Set TxIn`.

plutus-chain-index-core/src/Plutus/Contract/CardanoAPI.hs
```diff
@@ -37,7 +38,9 @@ fromCardanoTx eraInMode tx@(C.Tx txBody@(C.TxBody C.TxBodyContent{..}) _) = do
     let scriptMap = plutusScriptsFromTxBody txBody
         isTxScriptValid = fromTxScriptValidity txScriptValidity
         (datums, redeemers) = scriptDataFromCardanoTxBody txBody
-        inputs =
+        -- We need to sort the inputs as the order is important
+        -- to find the corresponding redeemers
+        inputs = sort $
```
ak3n, line 43: Here is the correct order of inputs as we sort cardano-api's `TxIn` with an indented `Ord` instance.

plutus-contract/src/Plutus/Contract/StateMachine.hs
```diff
@@ -128,8 +127,9 @@ getInput ::
     -> ChainIndexTx
     -> Maybe i
 getInput outRef tx = do
-    (_validator, Ledger.Redeemer r, _) <-
-      listToMaybe $ mapMaybe Tx.inScripts $ filter (\Tx.TxIn{Tx.txInRef} -> outRef == txInRef) $ Set.toList $ _citxInputs tx
+    -- We retrieve the correspondent redeemer according to the index of txIn in the list
+    let findRedeemer (ix, _) = Map.lookup (Tx.RedeemerPtr Tx.Spend ix) (_citxRedeemers tx)
+    Ledger.Redeemer r <- listToMaybe $ mapMaybe findRedeemer $ filter (\(_, Tx.TxIn{Tx.txInRef}) -> outRef == txInRef) $ zip [0..] $ _citxInputs tx
```
ak3n, line 132: Here is where we use the knowledge of inputs order to get the redeemer. I hope I got the idea correctly that index of `TxIn` is binded to the redeemer pointer.

Source: https://github.com/input-output-hk/cardano-node/issues/3389#issuecomment-1043165343

Maybe instead of relying on their order it's better to include the index. `Set TxIn ~> Set (TxInIndx, TxIn)`, that would make it more explicit.

plutus-ledger/src/Ledger/Tx/CardanoAPI.hs
```diff
@@ -308,15 +308,26 @@ scriptDataFromCardanoTxBody
...
+                         $ fromCardanoScriptData
+                         $ C.fromAlonzoData
+                         $ fst rdmr
+                        )
+                      )
+                $ Map.toList reds
    in (datums, redeemers)
 
+redeemerPtrFromCardanoRdmrPtr :: Alonzo.RdmrPtr -> P.RedeemerPtr
+redeemerPtrFromCardanoRdmrPtr (Alonzo.RdmrPtr rdmrTag ptr) = P.RedeemerPtr t (toInteger ptr)
```
ak3n, line 323: We already had `RedeemerPtr` but didn't use it before.

plutus-chain-index-core/src/Plutus/ChainIndex/Tx.hs
```diff
@@ -117,13 +118,21 @@ mintingPolicies = Map.fromList . fmap withHash . Set.toList
     withHash mp = let (MintingPolicyHash mph) = mintingPolicyHash mp
                    in (ScriptHash mph, getMintingPolicy mp)
 
-validators :: Set TxIn -> (Map ScriptHash Script, Map DatumHash Datum, Map RedeemerHash Redeemer)
-validators = foldMap (maybe mempty withHash . txInType) . Set.toList
+validators :: Set TxIn -> (Map ScriptHash Script, Map DatumHash Datum, Redeemers)
+validators txIns = foldMap (\(ix, txIn) -> maybe mempty (withHash ix) $ txInType txIn) $ zip [0..] (Set.toList txIns)
   where
-    withHash (ConsumeScriptAddress val red dat) =
+    -- TODO: the index of the txin is probably incorrect as we take it from the set.
```
andreabedini, line 124: Is this going to have bad consequences?

ak3n: My understanding is that this PR makes a move from a completely broken state to a semi-broken state. Instead of not getting redeemers at all, we will start choosing them but with a big chance of selecting the incorrect ones.

ak3n: Speaking of `fromOnChainTx` in particular, I think `withHash` function was useless already since we use the `Tx` here, and the inputs have an empty `TxInType` here as I understand.

plutus-contract/src/Plutus/Contract/StateMachine.hs
sjoerdvisscher: The fact that we now have a list instead of a set makes it clear enough that the order is significant in my opinion.

plutus-chain-index-core/src/Plutus/ChainIndex/Tx.hs
koslambrou: Can you add a TODO and a story for this? I guess it will need to be done once `next-node` is merged in `main`.

sjoerdvisscher: No need imho, this is good enough.

plutus-chain-index-core/src/Plutus/ChainIndex/Tx.hs
ak3n: >As long as the redeemer pointer points to the right input, I think this is totally fine.

There is a problem here that we take a set that could break the order. I looked at usages of the `fromOnChainTx` and there is no anything that directly relies on that order. But would be nice to fix `Tx` type.

PR #608, opened 2022-07-19
Add zero ada to mint and fee fields to match cardano-ledger behaviour

### LINKED ISSUE #604: Discrepancy between `plutus-apps` and `cardano-ledger` on handling values.
### Summary

The ledger implementation seems to always add a 0 lovelace entry to values, as seen [here](https://github.com/input-output-hk/cardano-ledger/blob/0726ec1fb2bbb3cf434aa46d3501109e2143de35/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs#L317). Yet, `plutus-apps` doesn't add this entry to values. In fact, [this](https://github.com/input-output-hk/plutus/blob/master/plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs) module never mentions that restriction anywhere.
As a result, the following script can distinguish whether we're running on the emulator or on-chain:

```
validate _ _ _ scriptCtx = adaSymbol `Map.member` txInfoMint (txInfo scriptCtx)
```

That script will be equivalent to `validate _ _ _ = True` on-chain, but `validate _ _ _ = False` on the emulator.
I wonder whether this is a bug in `plutus-apps` or whether the ledger just happen to add that entry to values, which seems not to change the semantic of values in any way. It depends on what the specification is and would be great to know more.

See the script example above.

### Actual Result

Script fails on the emulator.

### Expected Result

Script succeeds on the emulator.

_No response_

### System info

n/a

ak3n:
fix #604

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
+++ b/plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
@@ -642,7 +642,7 @@ processConstraint = \case
 
     MustMintValue mpsHash red tn i -> do
         mintingPolicyScript <- lookupMintingPolicy mpsHash
-        -- We add zero ada here to be consistent with cardano-ledger behaviour.
+        -- See note [Mint and Fee fields must have ada symbol].
         let value = Ada.lovelaceValueOf 0 <> Value.singleton (Value.mpsSymbol mpsHash) tn
         -- If i is negative we are burning tokens. The tokens burned must
         -- be provided as an input. So we add the value burnt to
--- a/plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
+++ b/plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
@@ -642,7 +642,8 @@ processConstraint = \case
 
     MustMintValue mpsHash red tn i -> do
         mintingPolicyScript <- lookupMintingPolicy mpsHash
-        let value = Value.singleton (Value.mpsSymbol mpsHash) tn
+        -- We add zero ada here to be consistent with cardano-ledger behaviour.
+        let value = Ada.lovelaceValueOf 0 <> Value.singleton (Value.mpsSymbol mpsHash) tn
         -- If i is negative we are burning tokens. The tokens burned must
         -- be provided as an input. So we add the value burnt to
         -- 'valueSpentInputs'. If i is positive then new tokens are created
--- a/plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
+++ b/plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
@@ -643,7 +643,7 @@ processConstraint = \case
     MustMintValue mpsHash red tn i -> do
         mintingPolicyScript <- lookupMintingPolicy mpsHash
         -- See note [Mint and Fee fields must have ada symbol].
-        let value = Ada.lovelaceValueOf 0 <> Value.singleton (Value.mpsSymbol mpsHash) tn
+        let value = ((<>) (Ada.lovelaceValueOf 0)) . Value.singleton (Value.mpsSymbol mpsHash) tn
         -- If i is negative we are burning tokens. The tokens burned must
         -- be provided as an input. So we add the value burnt to
         -- 'valueSpentInputs'. If i is positive then new tokens are created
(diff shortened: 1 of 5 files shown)


plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
```diff
@@ -642,7 +642,8 @@ processConstraint = \case
 
     MustMintValue mpsHash red tn i -> do
         mintingPolicyScript <- lookupMintingPolicy mpsHash
-        let value = Value.singleton (Value.mpsSymbol mpsHash) tn
+        -- We add zero ada here to be consistent with cardano-ledger behaviour.
```
koslambrou, line 645: Point to the relevant ledger code? https://github.com/input-output-hk/cardano-ledger/blob/0726ec1fb2bbb3cf434aa46d3501109e2143de35/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs

Unless we shouldn't do that to prevent having broken links in our docs?
Up to you.

plutus-ledger-constraints/src/Ledger/Constraints/OnChain.hs
```diff
@@ -104,7 +104,8 @@ checkTxConstraint ctx@ScriptContext{scriptContextTxInfo} = \case
         $ isJust (V.findTxInByTxOutRef txOutRef scriptContextTxInfo)
     MustMintValue mps _ tn v ->
         traceIfFalse "L9" -- "Value minted not OK"
-        $ Value.valueOf (txInfoMint scriptContextTxInfo) (Value.mpsSymbol mps) tn == v
+        -- We add zero ada here to be consistent with cardano-ledger behaviour.
```
koslambrou, line 107: This is the same comment as `OffChain.hs`. How about creating a [note](https://github.com/input-output-hk/plutus-apps/blob/main/CONTRIBUTING.adoc#notes) at the bottom of `OnChain.hs` and then refering to this note?

plutus-ledger/src/Ledger/Index.hs
```diff
@@ -381,8 +382,9 @@ mkTxInfo tx = do
     let ptx = TxInfo
             { txInfoInputs = txins
             , txInfoOutputs = txOutputs tx
-            , txInfoMint = txMint tx
-            , txInfoFee = txFee tx
+            -- We add zero ada to follow the intentional logic of `Cardano.Ledger.Alonzo.TxInfo.txInfo`.
```
koslambrou, line 385: Refer to a note here too?

koslambrou (approved):

koslambrou:
LGTM :)

plutus-ledger-constraints/src/Ledger/Constraints/OnChain.hs
ak3n: Great idea! Thanks!

koslambrou: Good point. I would say, why not? However, since there's no specification for how to construct the `ScriptContext`, this check feels ad-hoc.

L-as:
>  Add zero ada to mint and fee fields to match cardano-ledger behaviour #608 

This isn't correct. *All* values in `ScriptContext` must have 0 ADA, and must all be sorted lexicographically.

ak3n:
@L-as could you please specify what exactly is incorrect? We are following the logic of cardano-ledger here, if it's incorrect, then we should notify them. As we are planning to switch to their implementation and this PR is a quick fix of the bug.

plutus-ledger/src/Ledger/Index.hs
```diff
@@ -381,8 +382,9 @@ mkTxInfo tx = do
     let ptx = TxInfo
             { txInfoInputs = txins
             , txInfoOutputs = txOutputs tx
-            , txInfoMint = txMint tx
-            , txInfoFee = txFee tx
+            -- See note [Mint and Fee fields must have ada symbol]
+            , txInfoMint = Ada.lovelaceValueOf 0 <> txMint tx
+            , txInfoFee = Ada.lovelaceValueOf 0 <> txFee tx
```
sjoerdvisscher, line 387: This is not necessary, the fee is always in Ada and always more than 0.

sjoerdvisscher (approved):

PR #612, opened 2022-07-20
Added checks for database-locked and table-locked

MartorSkull:
When running the pab for long periods of time we came across these two errors:
```
SqlError "SQLite3 returned ErrorBusy while attempting to perform prepare \"SELECT \\\"t0\\\".\\\"instance_id\\\" AS \\\"res0\\\", \\\"t0\\\".\\\"instance_contract_id\\\" AS \\\"res1\\\", \\\"t0\\\".\\\"instance_wallet\\\" AS \\\"res2\\\", \\\"t0\\\".\\\"instance_state\\\" AS \\\"res3\\\", \\\"t0\\\".\\\"instance_active\\\" AS \\\"res4\\\" FROM \\\"instances\\\" AS \\\"t0\\\" WHERE (\\\"t0\\\".\\\"instance_active\\\")=(?)\": database is locked"
```
and
```
SqlError "SQLite3 returned ErrorBusy while attempting to perform step: database is locked"
```
Both of these seem to happen due to the asyncrounous nature of the PAB, when a coroutine tries to access the database while another operation is already happening. 
We first came found these issues when running the pab with the `-theaded` flag. We were able to mitigate this error by removing this option mostly, but from time to time is happening either way. This fix adds handlers for these SQLErrors on the BeamHandler in freer-effects. We believe this is the correct approach since on the original code a comment explained that the ErrorError constructor would deal with these cases, now we have added these cases specifically. From our testing this fix seems to mitigate the issue.

### DIFF
--- a/freer-extras/src/Control/Monad/Freer/Extras/Beam.hs
+++ b/freer-extras/src/Control/Monad/Freer/Extras/Beam.hs
@@ -194,9 +194,19 @@ runBeam trace action = do
       resultEither <- try $ Sqlite.withTransaction conn $ runBeamSqliteDebug traceSql conn action
       case resultEither of
           -- 'Database.SQLite.Simple.ErrorError' corresponds to an SQL error or
-          -- missing database. When this exception is raised, we suppose it's
-          -- because the another transaction was already running.
-          Left (Sqlite.SQLError Sqlite.ErrorError _ _) | retries > 0 -> do
+          -- missing database. When this exception is raised.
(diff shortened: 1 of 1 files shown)


koslambrou:
To fix the failing readthedocs build, update your plutus-apps repo to the latest commit of `main`.

koslambrou:
LGTM :) Thanks!

koslambrou (approved):

MartorSkull:
This PR is becomes obsolete with #713.

PR #613, opened 2022-07-21
[PLT-81] plutus-chain-index: support inline scripts when querying TxOut of a TxOutRef

andreabedini:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-chain-index-core/src/Plutus/ChainIndex/Handlers.hs
+++ b/plutus-chain-index-core/src/Plutus/ChainIndex/Handlers.hs
@@ -200,10 +200,10 @@ makeChainIndexTxOut txout@ChainIndexTxOut{..} =
         OutputDatumHash dh -> do
           v <- maybe (Left vh) Right <$> getScriptFromHash vh
           d <- maybe (Left dh) Right <$> getDatumFromHash dh
-          pure $ Just $ L.ScriptChainIndexTxOut citoAddress v d citoValue
+          pure $ Just $ L.ScriptChainIndexTxOut citoAddress citoValue d v
         OutputDatum d -> do
           v <- maybe (Left vh) Right <$> getScriptFromHash vh
-          pure $ Just $ L.ScriptChainIndexTxOut citoAddress v (Right d) citoValue
+          pure $ Just $ L.ScriptChainIndexTxOut citoAddress citoValue (Right d) v
         NoOutputDatum -> do
           -- If the txout comes from a script address, the Datum should not be Nothing
           logWarn $ NoDatumScriptAddr txout
--- a/plutus-chain-index-core/src/Plutus/ChainIndex/Handlers.hs
+++ b/plutus-chain-index-core/src/Plutus/ChainIndex/Handlers.hs
@@ -5,7 +5,6 @@
 {-# LANGUAGE NamedFieldPuns        #-}
 {-# LANGUAGE QuantifiedConstraints #-}
 {-# LANGUAGE RankNTypes            #-}
-{-# LANGUAGE RecordWildCards       #-}
 {-# LANGUAGE TypeApplications      #-}
 {-# LANGUAGE TypeOperators         #-}
 {-# LANGUAGE UndecidableInstances  #-}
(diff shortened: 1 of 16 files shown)


plutus-ledger/src/Ledger/Tx.hs
```diff
@@ -101,13 +103,14 @@ type PrivateKey = Crypto.XPrv
...
+                             , _ciTxOutValue          :: V1.Value
+                             , _ciTxOutDatumPublicKey :: V2.OutputDatum
                              }
-  | ScriptChainIndexTxOut { _ciTxOutAddress   :: Address
-                          , _ciTxOutValue     :: V1.Value
-                          , _ciTxOutDatum     :: Either V1.DatumHash V1.Datum
-                          , _ciTxOutValidator :: Either V1.ValidatorHash V1.Validator
+  | ScriptChainIndexTxOut { _ciTxOutAddress     :: Address
+                          , _ciTxOutValue       :: V1.Value
+                          , _ciTxOutDatumScript :: Either V1.DatumHash V1.Datum
```
andreabedini: I split `_ciTxOutDatum` into `_ciTxOutDatumPublicKey` and `_ciTxOutDatumScript` because the types are now different. If you have a better idea let me know.

plutus-chain-index-core/src/Plutus/ChainIndex/Handlers.hs
```diff
@@ -192,22 +192,27 @@ makeChainIndexTxOut ::
...
         OutputDatum d -> do
           v <- maybe (Left vh) Right <$> getScriptFromHash vh
-          pure $ Just $ L.ScriptChainIndexTxOut citoAddress v (Right d) citoValue
-        _ -> do
+          pure $ Just $ L.ScriptChainIndexTxOut address value (Right d) script v
+        NoOutputDatum -> do
           -- If the txout comes from a script address, the Datum should not be Nothing
           logWarn $ NoDatumScriptAddr txout
           pure Nothing
+  where
```
sjoerdvisscher, line 212: Ok, this is the 3rd time, it deserves a reusable function.

plutus-ledger/src/Ledger/Tx.hs
```diff
@@ -96,16 +101,17 @@ type PrivateKey = Crypto.XPrv
...
 -- 'Ledger.Constraints.processConstraint', specifically with the constraints
 -- 'MustSpendPubKeyOutput' and 'MustSpendScriptOutput'.
---
--- TODO Add 'Either DatumHash Datum' field for 'PublicKeyChainIndexTxOut'.
 data ChainIndexTxOut =
-    PublicKeyChainIndexTxOut { _ciTxOutAddress :: Address
-                             , _ciTxOutValue   :: Value
+    PublicKeyChainIndexTxOut { _ciTxOutAddress         :: Address
+                             , _ciTxOutValue           :: V1.Value
+                             , _ciTxOutDatumPublicKey  :: V2.OutputDatum
```
sjoerdvisscher, line 107: This reads wrong, how about `_ciTxOutPublicKeyDatum` and `_ciTxOutScriptDatum`

sjoerdvisscher (changes requested):

plutus-ledger/src/Ledger/Tx.hs
andreabedini: Yes, it sounds better.

plutus-ledger/src/Ledger/Tx.hs
```diff
@@ -96,16 +101,17 @@ type PrivateKey = Crypto.XPrv
...
                              }
-  | ScriptChainIndexTxOut { _ciTxOutAddress   :: Address
-                          , _ciTxOutValidator :: Either ValidatorHash Validator
-                          , _ciTxOutDatum     :: Either DatumHash Datum
-                          , _ciTxOutValue     :: Value
+  | ScriptChainIndexTxOut { _ciTxOutAddress         :: Address
+                          , _ciTxOutValue           :: V1.Value
+                          , _ciTxOutScriptDatum     :: Either V1.DatumHash V1.Datum
+                          , _ciTxOutReferenceScript :: Maybe V1.Script
+                          , _ciTxOutValidator       :: Either V1.ValidatorHash V1.Validator
```
koslambrou, line 114: I think it would be helpful to add a comment explaining the difference between the two. I got confused as well :)

plutus-ledger/src/Ledger/Tx.hs
```diff
@@ -115,22 +121,28 @@ makePrisms ''ChainIndexTxOut
...
+  V1.Tx.TxOut addr v (Just $ datumHash d)
 
 -- | Converts a plutus-ledger-api transaction output to the chain index
 -- transaction output.
-fromTxOut :: TxOut -> Maybe ChainIndexTxOut
-fromTxOut TxOut { txOutAddress, txOutValue, txOutDatumHash } =
-  case addressCredential txOutAddress of
-    PubKeyCredential _ -> pure $ PublicKeyChainIndexTxOut txOutAddress txOutValue
-    ScriptCredential vh ->
+fromTxOut :: V1.Tx.TxOut -> Maybe ChainIndexTxOut
```
koslambrou, line 137: May I suggest testing the  _partial_ inverse property test between `toTxOut` and `fromTxOut`? In the test, you could generate a `ChainIndexTxOut`, convert it to a `TxOut` and back to `ChainIndexTxOut`. Then you can verify what the expected result would be.

andreabedini: > In the test, you could generate a `ChainIndexTxOut`, convert it to a `TxOut` and back to `ChainIndexTxOut`. Then you can verify what the expected result would be.

This is wrong. From `ChainIndexTxOut` to `TxOut` is lossy as `ChainIndexTxOut` can have the full datum while `TxOut` only its hash. You probably meant the other way around `TxOut` -> `ChainIndex` -> `TxOut` and we have already a test for that. https://github.com/input-output-hk/plutus-apps/blob/6d1600f084e894a257e02d3d0095b1f7a752b464/plutus-ledger/test/Spec.hs#L229-L234

koslambrou: That's right! I think that's the point, you can also test the lossy conversion to make sure we lose just enough.

andreabedini: :thinking:

koslambrou: Ultimately, it's up to you :). If you think it doesn't make sense, no need to do it.

sjoerdvisscher (approved):

PR #617, opened 2022-07-22
Add missing field

eyeinsky:
The missing field prevented running marconi on testnet against a local node.

The bash shebang fix is just a minor fix. Nixos doesn't have a `/bin/bash`, but `/usr/bin/env bash` seems to be supported on all unixes so switch to that (in the entire repo, not sure if that's ok).

Pre-submit checklist:
- Branch
- PR
    - [X] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-streaming/src/Plutus/Streaming.hs
+++ b/plutus-streaming/src/Plutus/Streaming.hs
@@ -8,7 +8,7 @@ where
 import Cardano.Api (BlockInMode, CardanoMode, ChainPoint, ChainSyncClient (ChainSyncClient), ChainTip,
                     ConsensusModeParams (CardanoModeParams), EpochSlots (EpochSlots),
                     LocalChainSyncClient (LocalChainSyncClient),
-                    LocalNodeClientProtocols (LocalNodeClientProtocols, localChainSyncClient, localStateQueryClient, localTxSubmissionClient),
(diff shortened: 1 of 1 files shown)


andreabedini:
@eyeinsky is this PR supposed to be against the next-node branch rather than main?

eyeinsky:
@andreabedini it should be against main, am rebuilding on main to see if the change is still correct.

koslambrou:
If the PR is meant on `main`, then you should create your branch based on `main`. By looking at the commits, seems like you created your branch based on `next-node`. So you can either restart with a new branch based on `main`, or do `git pull --rebase origin main` and fix potential conflicts.

eyeinsky:
Right, so the situation is as follows:
- the localTxMonitoringClient comes from cardano-api package in cardano-node

eyeinsky:
Right, so the situation is as follows:
- localTxMonitoringClient field comes from cardano-api package from the cardano-node repo
- it's not present in the cardano-api which plutus-apps' main branch depends on

So it seems that adding the `localTxMonitoringClient` field does indeed belong do the next-node branch (unless we update main branches dependency to a more recent cardano-api), and the bash shebang belongs to master.

eyeinsky:
Created another PR for shebang fix here https://github.com/input-output-hk/plutus-apps/pull/618

Changed this PR to be based on next-node and only kept the "add field" commit.

koslambrou (approved):

raduom (approved):
Assuming we merge this into next-node.

PR #618, opened 2022-07-22
Replace `#!/bin/bash` with `#!/usr/bin/env bash`

eyeinsky:
Pre-submit checklist:
- Branch
- PR
    - [X] Self-reviewed the diff
    - [ ] Reviewer requested


koslambrou (approved):

whs-dot-hk:
It seems that the person wrote the script have `bash` in `/bin/bash`. Which the binaries are in  `/bin`, but this commit assume the binaries (i mean env) are in  `/usr/bin`

PR #624, opened 2022-07-25
Add plutus-contract tests for required signer and...

james-iohk:
...moved some tests into "TxConstraints directory.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-contract/plutus-contract.cabal
```diff
@@ -239,10 +239,11 @@ test-suite plutus-contract-test
     Spec.Plutus.Contract.Oracle
     Spec.Plutus.Contract.Wallet
     Spec.Rows
+    Spec.ScriptContext.V1.RequiredSigner
+    Spec.ScriptContext.V1.TimeValidity
```
koslambrou, line 243: @ak3n @sjoerdvisscher @andreabedini How should we name testing modules in plutus-contract? Let's put ideas here.

koslambrou: First off, I would suggest using the package name as the module name and make it consistent:
```
Spec.Plutus.Contract.ScriptContext.V1.RequiredSigner
```

`ScriptContext` might make us think that it is testing on-chain code only, but we're also testing off-chain using the `Contract` monad. Maybe `Spec.Plutus.Contract.IntegrationTest.V1.RequiredSigner` or simply `Spec.Plutus.Contract.V1.RequiredSigner`?

Other than that, I don't have any awesome ideas.

plutus-contract/test/Spec/ScriptContext/V1/RequiredSigner.hs
```diff
@@ -0,0 +1,167 @@
...
+    ledgerTx2 <- submitTxConstraintsWith @Void lookups2 tx2
+    awaitTxConfirmed $ Tx.getCardanoTxId ledgerTx2
+
+ownWallet :: TestTree
+ownWallet =
+    let pk  = mockWalletPaymentPubKey     w1
+        pkh = mockWalletPaymentPubKeyHash w1
+        trace = do
+            void $ Trace.activateContractWallet w1 $ contract pk pkh
+            void $ Trace.waitNSlots 3
```
koslambrou, line 77: Why 3? Shouldn't 1 be enough?

plutus-contract/test/Spec/ScriptContext/V1/RequiredSigner.hs
```diff
@@ -0,0 +1,167 @@
...
+otherWallet =
+    let pk  = mockWalletPaymentPubKey     w2
+        pkh = mockWalletPaymentPubKeyHash w2
+        trace = do
+            Trace.setSigningProcess w1 (Just $ signPrivateKeys [paymentPrivateKey $ fromJust $ walletToMockWallet w1, paymentPrivateKey $ fromJust $ walletToMockWallet w2])
+            void $ Trace.activateContractWallet w1 $ contract pk pkh
+            void $ Trace.waitNSlots 3
+    in checkPredicateOptions defaultCheckOptions "other wallet's signature passes on-chain validation" (assertValidatedTransactionCount 2) (void trace)
+
+missingRequiredSignature :: TestTree -- shouldn't this fail in the contract before script execution because it already knows it'll fail due to other mustBeSignedBy pubkey?
```
koslambrou, line 90: What do you mean by "it'll fail due to other mustBeSignedBy pubkey"?

plutus-contract/plutus-contract.cabal
james-iohk: I think I like `IntegrationTest` or something similar because these tests are covering a few areas (e.g. Contract, Lookups and ScriptContect's TxInfo)

plutus-contract/test/Spec/ScriptContext/V1/RequiredSigner.hs
james-iohk: The contract should be able to determine failure before executing the script because the required signer constraint has not been fulfilled. Why does it need to execute the script to determine phase-2 failure? If the script had a bug (i.e. didn't correctly check for required signer) then it'd pass phase-2 locally and proceed to submit, but then ledger would probably fail phase-1 because of the missing required signer. Why not catch this earlier?

plutus-contract/plutus-contract.cabal
andreabedini: Well, I'd say prepend `Spec` or `Test` to the tested module and use the longest prefix when testing multiple modules but this kinda assumes the module names itself are structured hierarchically.

plutus-contract/test/Spec/ScriptContext/V1/RequiredSigner.hs
koslambrou: @sjoerdvisscher Quickly looking at `Ledger.Validation`, seems like we are using the ledger's phase 1 validation, right? Shouldn't it catch that the signature is missing?

sjoerdvisscher: Yes, but we only show the first error and currently phase 2 emulator validation happens before phase 1 ledger validation.

koslambrou: I guess fixing this simply entails running phase 1 validation first?

koslambrou: Ok, found what causes this. If you look at the function `Wallet.Emulator.Chain.validateEm`, you'll notice that we run our custom validation rules before running the real ledger rules. Thus, in our custom validation rules, we don't have a rule which detects this phase1 error, that's why the script was run. By commented out our custom validation rules and running the tests, I'm getting:

```
      [WARNING] Slot 2: TxnValidationFail Phase1 124516ce2730493a63bf6af45e941162d3e33eba84a179d89890eea857f33013: CardanoLedgerValidationError "ApplyTxError [UtxowFailure (MissingRequiredSigners (fromList [KeyHash \"76aaef06f38cc98ed08ceb168ddb55bab2ea5df43a6847a99f086fc1\"]))]"
```

So when the real ledger rules are used, we are correctly getting a phase 1 error.

Therefore, I would propose you keep the current test case. We have a medium to high priority story in our backlog which is about completely removing our custom validation rules and using only the real ledger rules.

james-iohk: Perfect. I'm excited for using only ledger validation. It may not be necessary to inverse the rules if it's going to be changed again in the coming weeks.

koslambrou: True, but I'm proposing this so that you can fix the `assertFailedTransaction` right away in this PR. Up to you then :)

james-iohk: Let's leave it as is, the way it's been behaving for a while, and I'll comment each test I expect to be changed. Will wait for the final solution.

plutus-contract/plutus-contract.cabal
james-iohk: I've moved them into a "Constraints" directory for now, we can decided whether to split by version etc later.

plutus-contract/test/Spec/ScriptContext/V1/RequiredSigner.hs
james-iohk: Left the following comment on two tests: "this should fail as phase-1 in the contract before script execution. We need to run phase-1 before phase-2 or only use ledger validation."

james-iohk:
I have just added a few more tests using the on-chain constraint.

There are currently 2 failing tests.
@sjoerdvisscher I expect `withoutOffChainMustBeSignedBy` and `withoutOffChainMustBeSignedByOnChainConstraint` (both similar)  to fail because I'm not using off-chain constraint `mustBeSignedBy`, and the trace logs show `Requires signatures:` with no signatures. The fact it passes phase-2 makes me think the contract is silently including the expected required signer before plutus script evaluation. It should at least log the required signers added to the txbody. Any thoughts?

plutus-contract/test/Spec/TxConstraints/RequiredSigner.hs
```diff
@@ -0,0 +1,217 @@
...
+        , ownWalletOnChainConstraint
+        , missingRequiredSignatureOnChainConstraint
+        , withoutOffChainMustBeSignedByOnChainConstraint
+        ]
+
+mustBeSignedByContract :: Ledger.PaymentPubKey -> Ledger.PaymentPubKeyHash -> Scripts.TypedValidator UnitTest -> Contract () Empty ContractError ()
+mustBeSignedByContract pk pkh tv = do
+    let validator = Scripts.validatorScript tv
+        lookups1 = Constraints.plutusV1TypedValidatorLookups tv
+        tx1 = Constraints.mustPayToTheScript () (Ada.lovelaceValueOf 25000000)
```
koslambrou, line 58: Large numbers like these are hard to read. You can the NumericUnderscores language extension here and do `25_000_ 000`

plutus-contract/test/Spec/TxConstraints/RequiredSigner.hs
```diff
@@ -0,0 +1,217 @@
...
+
+    utxos <- utxosAt (Ledger.scriptHashAddress $ Scripts.validatorHash tv)
+    let orefs = fst <$> Map.toList utxos
+        redeemer = Redeemer $ toBuiltinData pkh
+        lookups2 =
+            Constraints.plutusV1OtherScript validator
+            <> Constraints.unspentOutputs utxos
+            <> Constraints.paymentPubKey pk
+        tx2 =
+            foldMap (\oref -> Constraints.mustSpendScriptOutput oref redeemer) orefs
```
koslambrou, line 70: Use `collectFromPlutusV1Script` or `collectFromTheScript`?

plutus-contract/test/Spec/TxConstraints/RequiredSigner.hs
```diff
@@ -0,0 +1,217 @@
...
+import PlutusTx qualified
+import PlutusTx.Prelude qualified as P
+import Prelude
+import Wallet.Emulator.Wallet (signPrivateKeys, walletToMockWallet)
+
+tests :: TestTree
+tests =
+    testGroup "required signer"
+        [
+        -- tests that use txSignedBy on-chain
```
koslambrou, line 42: Why test the `txSignedBy` along `mustBeSignedBy` constraint? `mustBeSignedBy` uses `txSignedBy`.

plutus-contract/test/Spec/TxConstraints/RequiredSigner.hs
```diff
@@ -0,0 +1,217 @@
...
+        lookups1 = Constraints.plutusV1TypedValidatorLookups tv
+        tx1 = Constraints.mustPayToTheScript () (Ada.lovelaceValueOf 25000000)
+    ledgerTx1 <- submitTxConstraintsWith lookups1 tx1
+    awaitTxConfirmed $ Tx.getCardanoTxId ledgerTx1
+
+    utxos <- utxosAt (Ledger.scriptHashAddress $ Scripts.validatorHash tv)
+    let orefs = fst <$> Map.toList utxos
+        redeemer = Redeemer $ toBuiltinData pkh
+        lookups2 =
+            Constraints.plutusV1OtherScript validator
```
koslambrou, line 66: Since you're using TypedValidator for your contract, I suggest using `Constraints.plutusV1TypedValidatorLookups`. Which implies that you can use `collectFromTheScript` later on

plutus-contract/test/Spec/TxConstraints/RequiredSigner.hs
```diff
@@ -0,0 +1,217 @@
...
+            <> Constraints.unspentOutputs utxos
+            <> Constraints.paymentPubKey pk
+        tx2 =
+            foldMap (\oref -> Constraints.mustSpendScriptOutput oref redeemer) orefs
```
koslambrou, line 93: Same here.

plutus-contract/test/Spec/TxConstraints/RequiredSigner.hs
koslambrou: Also, using typed validator means that you can use `pkh` as the redeemer instead of doing `Redeemer $ toBuiltingData pkh`.

plutus-contract/test/Spec/TxConstraints/RequiredSigner.hs
james-iohk: Initially thought there's be more different outcomes to check. Have changed to use `mustBeSignedBy` for most tests. One test exists to for `txSignedBy` phase-2 failure.

koslambrou: I don't understand. Both validators are the exact same no? `mustBeSigned` uses the same implementation you used for `txSignedBy`.

james-iohk: You're right. Now removed.

plutus-contract/plutus-contract.cabal
```diff
@@ -122,6 +122,7 @@ library
     , freer-extras
     , plutus-chain-index-core
     , plutus-ledger
+    , plutus-ledger-api
```
koslambrou, line 125: This should go in the "Other IOG dependencies" down below.

koslambrou (approved):

plutus-contract/plutus-contract.cabal
james-iohk: Ah, forgot to revert this. Was trying to fix my breaking HLS.

PR #625, opened 2022-07-25
Refactored the tx constraints library to prepare for PlutusV2

koslambrou:
* Changed functions names in plutus-ledger-constraints to reflect the Plutus version
  * `typedValidatorLookups` -> `plutusV1TypedValidatorLookups`
  * `mintingPolicy` -> `plutusV1MintingPolicy`
  * `otherScript` -> `plutusV1OtherScript`

* Removed reexport of `Ledger.Constraints.OnChain.V1.checkScriptContext` from module `Ledger.Constraints`.

* Moved functions in `Plutus.Contract.Typed.Tx` and `Plutus.Contract.Tx`
  to plutus-ledger-constraints.

* Changed 'Ledger.Tx.ChainIndexTxOut' from Either to Tuple for the datum
  and validator. This prevents from having to recompute the hash from
  the value when we have `Right ...` since the hashing function of
  PlutusV1 is not the same as the one in PlutusV2 (for scripts, not for
  datums, but for consistency we apply the same refactoring for datums).

* Backported some refactoring changes from `next-node` such as the
  function 'Plutus.ChainIndex.Handlers.makeChainIndexTxOut' and the
  refactoring surrounding `Ledger.Tx.ChainIndexTxOut`.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


andreabedini:
Damn, I should have published my refactoring branch. Changes to `ChainIndexTxOut` will be a pain to merge :)

koslambrou:
@andreabedini No problem. I can adapt my code considering your changes to `ChainIndexTxOut`. The only difference I added in `ChainIndexTxOut` in `main` was to use a Tuple instead of Either for the datum and validator.

andreabedini:
> @andreabedini No problem. I can adapt my code considering your changes to `ChainIndexTxOut`. The only difference I added in `ChainIndexTxOut` in `main` was to use a Tuple instead of Either for the datum and validator.

I'll book some time to discuss both our chages ASAP. I'm still doing minor fixes all through the code base :exploding_head:

plutus-use-cases/src/Plutus/Contracts/Uniswap/OffChain.hs
```diff
@@ -259,9 +259,9 @@ close us CloseParams{..} = do
         lVal     = valueOf lC liquidity
         redeemer = Redeemer $ PlutusTx.toBuiltinData Close
 
-        lookups  = Constraints.typedValidatorLookups usInst        <>
-                   Constraints.otherScript usScript                <>
-                   Constraints.mintingPolicy (liquidityPolicy us) <>
+        lookups  = Constraints.plutusV1TypedValidatorLookups usInst        <>
+                   Constraints.otherPlutusV1Script usScript                <>
```
ak3n, line 263: Maybe should be `plutusV1OtherScript`? To be consistent in naming.

ak3n (commented):
Looks good but I have a comment about the naming.

I dislike the usage of `plutusV` prefix as a namespace approach because there are modules for this task.

Actually, we already have started mixing them:

```
import Ledger.Constraints.OnChain.V1 as Constraints

...

Constraints.checkPlutusV1OwnOutputConstraint 
```

I think it's tedious, as a user I would need to remember when a function starts with `plutusV1` or there is `plutusV1` somewhere in the name. Also, we might be inconsistent in the future or just not careful as with `otherPlutusV1Script` and that might lead to many functions with different names.

I prefer modules to deal with namespacing and scopes.

1. We still can use one version in one module:

```
import Ledger.Constraints.OnChain.V1 as Constraints

...

Constraints.checkOwnOutputConstraint 
```

and it will be easier to switch to a different version by changing the import.

2. We can mix the versions in one module without polluting the scope:

```
import Ledger.Constraints.OnChain.V1 as Constraints.V1
import Ledger.Constraints.OnChain.V2 as Constraints.V2

...

Constraints.V1.checkOwnOutputConstraint 
```

The names of functions will be the same, we don't need to change them, only the name of the module.

koslambrou:
@ak3n Very good point. One of the reason I decided to go with `plutusV...` was to keep global modules which reexport everything like `Ledger.Constraints`. How would this module reexport a function which has the same name, but is in two different modules? Unless we should split it with `Ledger.Constraints.V1` and `Ledger.Constraints.V2`?

ak3n:
@koslambrou Yeah, should we split like in #628?

koslambrou:
@ak3n For now, I went back to the old naming for`checkScriptContext`, `checkOwnInputConstraint` and `checkOwnOutputConstraint`, and removed the reexport in `Ledger.Constraints`. Therefore, something wanting to use this on-chain bit should import `Ledger.Constraints.OnChain.V1.checkScriptContext` or `Ledger.Constraints.OnChain.V1.checkScriptContext` in the future.

sjoerdvisscher (approved):
At some point I guess it will be useful to have a `TypedValidator` that can contain a script in any plutus version, and then maybe this can use that? We'll learn by experience when it's useful to split code into one for each plutus version and when a sum type will do.

koslambrou:
@sjoerdvisscher Yep, I agree

PR #629, opened 2022-07-27
PLT-172: Marconi index transactions that have been issued to a particular script

eyeinsky:
This is a draft PR for the jira issue [marconi indexer: transactions that have been issued to a particular script](https://input-output.atlassian.net/browse/PLT-172). (This PR is also wrongly based off of `next-node` branch because it can understand the blocks since Vasil hard-fork -- will rebase once I manage to resolve the issues below.)

A concrete code thing I need help with is extracting script hashes from a transaction [here](https://github.com/input-output-hk/plutus-apps/blob/9a2e24eb1de814c33d42c7b2ecc11e8e62620361/plutus-chain-index/app/Marconi.hs#L261): how/where would I find the type(s)?

Another question is that the Jira issue says to set script address as the primary key, but if this is an index from transaction to script, then this is an 1:n relationship because (as far as I understand) a single script will be possibly targeted/ran by many transactions.

I've also read the indexer's code (the hysterical-screams) but can't say it's fully clear yet :), so would be good to see if the onInsert/store/query triple is correct.

Any other comments welcome as well!

(The main commits are 9a2e24eb1 and 79a9d7fad, the two previous ones already exist as separate PRs.)

---

Pre-submit checklist:
- Branch
- PR
    - [X] Self-reviewed the diff
    - [X] Reviewer requested

### DIFF
--- a/plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
+++ b/plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
@@ -1,0 +1,197 @@
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE GADTs                      #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+
+module Marconi.Index.ScriptTx where
+
+import Codec.Serialise (deserialiseOrFail)
+import Control.Lens.Operators ((^.))
+import Data.ByteString qualified as BS
+import Data.Foldable (forM_, toList)
+import Data.Maybe (catMaybes, fromJust)
+import Database.SQLite.Simple qualified as SQL
+import Database.SQLite.Simple.FromField qualified as SQL
+import Database.SQLite.Simple.ToField qualified as SQL
+import GHC.Generics (Generic)
+
+import Cardano.Api (SlotNo)
+import Cardano.Api qualified as C
+import Cardano.Api.Shelley qualified as Shelley
+import Cardano.Ledger.Alonzo.Language qualified as Alonzo
+import Cardano.Ledger.Alonzo.Scripts qualified as Alonzo
+import Cardano.Ledger.Core qualified
+import Cardano.Ledger.Crypto qualified as LedgerCrypto
+import Cardano.Ledger.Keys qualified as LedgerShelley
+import Cardano.Ledger.Shelley.Scripts qualified as LedgerShelley
+import Cardano.Ledger.ShelleyMA.Timelocks qualified as Timelock
+
+import RewindableIndex.Index.VSqlite (SqliteIndex)
+import RewindableIndex.Index.VSqlite qualified as Ix
+
+
+newtype Depth = Depth Int
+
+newtype ScriptAddress = ScriptAddress Shelley.ScriptHash
+  deriving (Show, Eq)
+newtype TxCbor = TxCbor BS.ByteString
+  deriving (Show)
+  deriving newtype (SQL.ToField, SQL.FromField)
+
+-- * SQLite
+
+data ScriptTxRow = ScriptTxRow
+  { scriptAddress :: !ScriptAddress
+  , txCbor        :: !TxCbor
+  } deriving (Generic)
+
+instance SQL.ToField ScriptAddress where
+  toField (ScriptAddress hash)  = SQL.SQLBlob . Shelley.serialiseToRawBytes $ hash
+instance SQL.FromField ScriptAddress where
+  fromField f = SQL.fromField f >>=
+    either
+      (const cantDeserialise)
+      (\b -> maybe cantDeserialise (return . ScriptAddress) $ Shelley.deserialiseFromRawBytes Shelley.AsScriptHash b) . deserialiseOrFail
+    where
+      cantDeserialise = SQL.returnError SQL.ConversionFailed f "Cannot deserialise address."
+
+instance SQL.ToRow ScriptTxRow where
+  toRow o = [SQL.toField $ scriptAddress o, SQL.toField $ txCbor o]
+
+-- * Indexer
+
+type Query = ScriptAddress
+type Result = [TxCbor]
+
+data ScriptTxUpdate = ScriptTxUpdate
+  { txScripts :: [(TxCbor, [ScriptAddress])]
+  , slotNo    :: !SlotNo
+  } deriving (Show)
+
+type ScriptTxIndex = SqliteIndex ScriptTxUpdate () Query Result
+
+
+toUpdate :: forall era . C.IsCardanoEra era => [C.Tx era] -> SlotNo -> ScriptTxUpdate
+toUpdate txs = ScriptTxUpdate txScripts'
+  where
+    txScripts' = map (\tx -> (TxCbor $ C.serialiseToCBOR tx, getTxScripts tx)) txs
+
+getTxBodyScripts :: forall era . C.TxBody era -> [ScriptAddress]
+getTxBodyScripts body = let
+    hashesMaybe :: [Maybe C.ScriptHash]
+    hashesMaybe = case body of
+      Shelley.ShelleyTxBody shelleyBasedEra _ scripts _ _ _ -> flip map scripts $ \script ->
+        case fromShelleyBasedScript shelleyBasedEra script of
+          Shelley.ScriptInEra _ script' -> Just $ C.hashScript script'
+      _ -> [] -- Byron transactions have no scripts
+    hashes = catMaybes hashesMaybe :: [Shelley.ScriptHash]
+  in map ScriptAddress hashes
+
+getTxScripts :: forall era . C.Tx era -> [ScriptAddress]
+getTxScripts (C.Tx txBody _ws) = getTxBodyScripts txBody
+
+open :: FilePath -> Depth -> IO ScriptTxIndex
+open dbPath (Depth k) = do
+  ix <- fromJust <$> Ix.newBoxed query store onInsert k ((k + 1) * 2) dbPath
+  let c = ix ^. Ix.handle
+  SQL.execute_ c "CREATE TABLE IF NOT EXISTS script_transactions (scriptAddress TEXT NOT NULL, txCbor BLOB NOT NULL)"
+  SQL.execute_ c "CREATE INDEX IF NOT EXISTS script_address ON script_transactions (scriptAddress)"
+  pure ix
+
+  where
+    onInsert :: ScriptTxIndex -> ScriptTxUpdate -> IO [()]
+    onInsert _ix _update = pure []
+
+    store :: ScriptTxIndex -> IO ()
+    store ix = do
+      persisted <- Ix.getEvents $ ix ^. Ix.storage
+      buffered <- Ix.getBuffer $ ix ^. Ix.storage
+      let updates = buffered ++ persisted :: [ScriptTxUpdate]
+          rows = do
+            ScriptTxUpdate txScriptAddrs _slotNo <- updates
+            (txCbor', scriptAddrs) <- txScriptAddrs
+            scriptAddr <- scriptAddrs
+            pure $ ScriptTxRow scriptAddr txCbor'
+      forM_ rows $
+        SQL.execute (ix ^. Ix.handle) "INSERT INTO script_transactions (scriptAddress, txCbor) VALUES (?, ?)"
+
+    query :: ScriptTxIndex -> Query -> [ScriptTxUpdate] -> IO Result
+    query ix scriptAddress' memory = let
+        filterByScriptAddress :: [ScriptTxUpdate] -> [TxCbor]
+        filterByScriptAddress updates' = do
+          ScriptTxUpdate update _slotNo <- updates'
+          map fst $ filter (\(_, addrs) -> scriptAddress' `elem` addrs) (update )
+      in do
+      persisted :: [SQL.Only TxCbor] <- SQL.query (ix ^. Ix.handle)
+        "SELECT txCbor FROM utxos WHERE scriptAddress = ?" (SQL.Only scriptAddress')
+
+      buffered <- Ix.getBuffer $ ix ^. Ix.storage
+      let
+        both :: [TxCbor]
+        both = filterByScriptAddress memory
+          <> filterByScriptAddress buffered
+          <> map (\(SQL.Only txCbor') -> txCbor') persisted
+
+      return both
+
+-- * Copy-paste
+--
+-- | TODO: Remove when the following function is exported from Cardano.Api.Script
+-- PR: https://github.com/input-output-hk/cardano-node/pull/4386
+fromShelleyBasedScript  :: Shelley.ShelleyBasedEra era
+                        -> Cardano.Ledger.Core.Script (Shelley.ShelleyLedgerEra era)
+                        -> Shelley.ScriptInEra era
+fromShelleyBasedScript era script =
+  case era of
+    Shelley.ShelleyBasedEraShelley ->
+      Shelley.ScriptInEra Shelley.SimpleScriptV1InShelley $
+      Shelley.SimpleScript Shelley.SimpleScriptV1 $
+      fromShelleyMultiSig script
+    Shelley.ShelleyBasedEraAllegra ->
+      Shelley.ScriptInEra Shelley.SimpleScriptV2InAllegra $
+      Shelley.SimpleScript Shelley.SimpleScriptV2 $
+      fromAllegraTimelock Shelley.TimeLocksInSimpleScriptV2 script
+    Shelley.ShelleyBasedEraMary ->
+      Shelley.ScriptInEra Shelley.SimpleScriptV2InMary $
+      Shelley.SimpleScript Shelley.SimpleScriptV2 $
+      fromAllegraTimelock Shelley.TimeLocksInSimpleScriptV2 script
+    Shelley.ShelleyBasedEraAlonzo ->
+      case script of
+        Alonzo.TimelockScript s ->
+          Shelley.ScriptInEra Shelley.SimpleScriptV2InAlonzo $
+          Shelley.SimpleScript Shelley.SimpleScriptV2 $
+          fromAllegraTimelock Shelley.TimeLocksInSimpleScriptV2 s
+        Alonzo.PlutusScript Alonzo.PlutusV1 s ->
+          Shelley.ScriptInEra Shelley.PlutusScriptV1InAlonzo $
+          Shelley.PlutusScript Shelley.PlutusScriptV1 $
+          Shelley.PlutusScriptSerialised s
+        Alonzo.PlutusScript Alonzo.PlutusV2 s ->
+          Shelley.ScriptInEra Shelley.PlutusScriptV2InAlonzo $
+          Shelley.PlutusScript Shelley.PlutusScriptV2 $
+          Shelley.PlutusScriptSerialised  s
+
+  where
+  fromAllegraTimelock :: Shelley.TimeLocksSupported lang
+                      -> Timelock.Timelock LedgerCrypto.StandardCrypto
+                      -> Shelley.SimpleScript lang
+  fromAllegraTimelock timelocks = go
+    where
+      go (Timelock.RequireSignature kh) = Shelley.RequireSignature
+                                            (Shelley.PaymentKeyHash (LedgerShelley.coerceKeyRole kh))
+      go (Timelock.RequireTimeExpire t) = Shelley.RequireTimeBefore timelocks t
+      go (Timelock.RequireTimeStart  t) = Shelley.RequireTimeAfter  timelocks t
+      go (Timelock.RequireAllOf      s) = Shelley.RequireAllOf (map go (toList s))
+      go (Timelock.RequireAnyOf      s) = Shelley.RequireAnyOf (map go (toList s))
+      go (Timelock.RequireMOf      i s) = Shelley.RequireMOf i (map go (toList s))
+
+  fromShelleyMultiSig :: LedgerShelley.MultiSig LedgerCrypto.StandardCrypto -> Shelley.SimpleScript lang
+  fromShelleyMultiSig = go
+    where
+      go (LedgerShelley.RequireSignature kh)
+                                  = Shelley.RequireSignature
+                                      (Shelley.PaymentKeyHash (LedgerShelley.coerceKeyRole kh))
+      go (LedgerShelley.RequireAllOf s) = Shelley.RequireAllOf (map go s)
+      go (LedgerShelley.RequireAnyOf s) = Shelley.RequireAnyOf (map go s)
+      go (LedgerShelley.RequireMOf m s) = Shelley.RequireMOf m (map go s)
--- a/plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
+++ b/plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
@@ -106,11 +106,9 @@ open dbPath (Depth k) = do
 
     store :: ScriptTxIndex -> IO ()
     store ix = do
-      persisted <- Ix.getEvents $ ix ^. Ix.storage
       buffered <- Ix.getBuffer $ ix ^. Ix.storage
-      let updates = buffered ++ persisted :: [ScriptTxUpdate]
-          rows = do
-            ScriptTxUpdate txScriptAddrs _slotNo <- updates
+      let rows = do
+            ScriptTxUpdate txScriptAddrs _slotNo <- buffered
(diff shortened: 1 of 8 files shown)


koslambrou:
This PR should be done in the `main` branch.

koslambrou:
> (This PR is also wrongly based off of next-node branch because it can understand the blocks since Vasil hard-fork -- will rebase once I manage to resolve the issues below.)

I don't think you need to understand blocks from Vasil HF to make this PR work. You can just run the indexer on Cardano mainnet (which there is no Vasil HF), but not on testnet.

plutus-chain-index/app/Marconi/Index/ScriptTx.hs
```diff
@@ -0,0 +1,107 @@
...
+  { _scriptAddress :: !ScriptAddress
+  , _txCbor        :: !TxCbor
+  } deriving (Generic)
+
+$(makeLenses ''ScriptTxRow)
+
+instance SQL.ToField ScriptAddress where
+  toField (ScriptAddress hash)  = SQL.SQLBlob . toStrict . serialise $ hash
+instance SQL.FromField ScriptAddress where
+  fromField _ = undefined -- todo
```
raduom, line 48: Should we do this before merging?

plutus-chain-index/app/Marconi/Index/ScriptTx.hs
```diff
@@ -0,0 +1,107 @@
...
+
+$(makeLenses ''ScriptTxUpdate)
+
+type ScriptTxIndex = SqliteIndex ScriptTxUpdate () Query Result
+
+open :: FilePath -> Depth -> IO ScriptTxIndex
+open dbPath (Depth k) = do
+  ix <- fromJust <$> Ix.newBoxed query store onInsert k ((k + 1) * 2) dbPath
+  let c = ix ^. Ix.handle
+  SQL.execute_ c "CREATE TABLE IF NOT EXISTS script_transactions (scriptAddress TEXT NOT NULL, txCbor BLOB NOT NULL)"
```
raduom, line 71: I think I would rather have a primary key index on `scriptAddress`. Is there a reason why we are not doing that?

plutus-chain-index/app/Marconi/Index/ScriptTx.hs
```diff
@@ -0,0 +1,107 @@
...
+  SQL.execute_ c "CREATE INDEX IF NOT EXISTS script_address ON script_transactions (scriptAddress)"
+  pure ix
+
+  where
+    onInsert :: ScriptTxIndex -> ScriptTxUpdate -> IO [()]
+    onInsert _ix _update = pure []
+
+    store :: ScriptTxIndex -> IO ()
+    store ix = do
+      persisted <- Ix.getEvents $ ix ^. Ix.storage
```
raduom, line 81: This is a bug I just found. We only want to save things in the buffer, not in the events.

plutus-chain-index/app/Marconi/Index/ScriptTx.hs
```diff
@@ -0,0 +1,107 @@
...
+
+instance SQL.ToRow ScriptTxRow where
+  toRow o = [SQL.toField $ o ^. scriptAddress, SQL.toField $ o ^. txCbor]
+
+-- * Indexer
+
+type Query = ScriptAddress
+type Result = [TxCbor]
+
+data ScriptTxUpdate = ScriptTxUpdate
```
raduom, line 58: I think we can probably make everything strict since these will eventually be persisted to disk.

plutus-chain-index/app/Marconi/Index/ScriptTx.hs
```diff
@@ -0,0 +1,107 @@
...
+        "SELECT txCbor FROM utxos WHERE scriptAddress = ?" (SQL.Only scriptAddress')
+
+      let
+        buffered :: [TxCbor]
+        buffered = do
+          ScriptTxUpdate update _slotNo <- updates
+          map fst $ filter (\(_, addrs) -> scriptAddress' `elem` addrs) update
+
+        both :: [TxCbor]
+        both = buffered <> map (\(SQL.Only txCbor') -> txCbor') persisted
```
raduom, line 105: I think an `Iso` may look better here, but this also works.

plutus-chain-index-core/notes/ConnectingToTheTestnet.md
```diff
@@ -14,7 +14,7 @@ You need to download the configurations for the node, genesis blocks and the top
 I am using a shell script to start the node that I will paste here:
 
 ```shell
-#!/bin/bash
+#!/usr/bin/env bash
```
raduom, line 17: In the future if you see a quick fix that is possible create a new PR for it as there are many more people who can review and accept it.

plutus-chain-index/app/Marconi/Index/ScriptTx.hs
eyeinsky: Yes, it's finding the script hash (linked in the PR description) that I haven't been able to find yet (but working on it!).

plutus-chain-index/app/Marconi/Index/ScriptTx.hs
eyeinsky: I might be wrong here, but since a script can be targeted by many transactions (?), then the `scriptAddress` won't be unique and so it can't be a primary key. (Though it still can have a index on it though, so it would be faster to search for with sql).

eyeinsky: (The `scriptAddress` primary key issue is also mentioned in the PR description.)

plutus-chain-index-core/notes/ConnectingToTheTestnet.md
eyeinsky: Yup, already did #618! (As that one is now merged, then can rebase it now and the commit won't be part of this PR anymore)

eyeinsky:
@koslambrou @raduom With regarding to adding tests: marconi is currently an executable within the plutus-chain-index package, but to add tests I think it needs to be converted into a library (+ executable)? so that in the tests I could import functions from it.

eyeinsky:
In discussion with @raduom we found that:
- there isn't anything to property-test specifically in the script transactions indexer and that an integration test would be what is needed
- there is a PR #524 that moves Marconi to the plutus-chain-index-core package, and which makes the Marconi.* modules a library (so that they could be imported by tests)

Thus, would it make sense to wait for #524 to be merged and then add an integration in a later PR/Jira to test that all works as expected? @koslambrou

eyeinsky:
This branch is ready for review, given that:
- attempted to replace [this](https://github.com/input-output-hk/plutus-apps/pull/629/files#diff-3884d69af3672d023232b3f69c6c097f4a0452bb63cd80196a4dc9673a29e6d9R264-R265) with [this](https://github.com/eyeinsky/plutus-apps/commit/0dd428d1e691fcdbe1b7de11e59a8c57ff86d1b7#diff-3884d69af3672d023232b3f69c6c097f4a0452bb63cd80196a4dc9673a29e6d9R283-R294), but it adds dependencies and doesn't actually improve performance much (plutusScriptsFromTxBody isn't doing extra work to also return scritps as map values)
- tests will come in a future PR

koslambrou:
@eyeinsky 

> there isn't anything to property-test specifically in the script transactions indexer and that an integration test would be what is needed

Right, not in the indexer itself. The pure function that would be nice to test is `txScripts`, because then we can actually use in the integration test itself. You're using `plutusScriptsFromTxBody` , but I don't think Marconi should be using that. I'll comment more on the code itself.

> Thus, would it make sense to wait for https://github.com/input-output-hk/plutus-apps/pull/524 to be merged and then add an integration in a later PR/Jira to test that all works as expected? @koslambrou

That's right, you can work on the integration test on a later PR :)

plutus-chain-index/app/Marconi.hs
```diff
@@ -223,23 +228,64 @@ utxoWorker Coordinator{_barrier} ch path = Utxo.open path (Utxo.Depth 2160) >>=
...
+      where
+        txScripts' = map (\tx -> (txCbor tx, txScripts tx)) txs
+
+    txCbor :: forall era . C.IsCardanoEra era => Tx era -> ScriptTx.TxCbor
+    txCbor tx = ScriptTx.TxCbor $ C.serialiseToCBOR tx
+
+    txScripts :: forall era . Tx era -> [ScriptTx.ScriptAddress]
+    txScripts tx = let
+        Tx (body :: C.TxBody era) _ws = tx
+        map' = plutusScriptsFromTxBody body :: M.Map Ledger.ScriptHash Ledger.Script
```
koslambrou, line 264: I don't suggest using `plutusScriptsFromTxBody`. Marconi is an off-chain component, so it should be using types for off-chain use (like `cardano-api`). Anything with with `plutus-ledger-api` types are really meant for on-chain use.

I suggest reimplementing it in Marconi (something named like `getScriptHashesFromTx`) and write tests for it.

plutus-chain-index-core/src/Plutus/Contract/CardanoAPI.hs
```diff
@@ -34,7 +33,7 @@ fromCardanoTx
   -> Either FromCardanoError ChainIndexTx
 fromCardanoTx eraInMode tx@(C.Tx txBody@(C.TxBody C.TxBodyContent{..}) _) = do
     txOutputs <- traverse fromCardanoTxOut txOuts
-    let scriptMap = plutusScriptsFromTxBody txBody
+    let scriptMap = plutusScriptsFromTxBody txBody -- HERE
```
koslambrou, line 36: Why "HERE"

eyeinsky: Forgot to remove it, removed 97bcb71b4

plutus-chain-index/app/Marconi.hs
eyeinsky: `plutusScriptsFromTxBody` hashes the script by
- deserialising it into a plutus-ledger `Script` (defined in Plutus.V1.Ledger.Scripts)
- scriptHash serialises it again to bytes and makes a cardano-api
  `Script` (defined in Cardano.Api.Script; that script wraps the
  serialised bytes), then hashes that script

If the goal is to skip plutus-ledger `Script` then is it ok if I wrap the received bytes into cardano-api's `Script` directly? I do this [here](https://github.com/input-output-hk/plutus-apps/blob/d376c2835deec98f8f8475ee9a0e50472c3cefd4/plutus-chain-index/app/Marconi.hs#L277-L287), specifically within the `mkCardanoApiScript` which wraps the incoming bytes into a cardano-api `Script` (the previous way to get this script included going through deserialisation/serialisation steps).

Otherwise, I haven't found a way to go from the bytes I receive to a cardano-api `Script` without using anything from the ledger.

---

Regarding tests, since it's a property test, then if I generate arbitrary scripts, serialize them and get a hash, and then put the scripts into a transaction to see if I get the same scripts (= same hashes) out again, then what would it be testing? (they are the same by definition) Or, perhaps there are multiple locations to put those scripts and I'd see if I lose any by not handling a branch?

koslambrou: Hmm... ideally we would use types from `cardano-api`, not `cardano-ledger`. I recommend not pattern matching on the body of the transaction. I recommend pattern matching with `pattern TxBody :: TxBodyContent ViewTx era -> TxBody era` in `Cardano.Api.TxBody`.

There are some examples in `Ledger.Tx.CardanoAPI`.

Regarding property testing, here's a starting point. 
You can test the `txScripts` function as follows. Generate a random number of plutus scripts (see `genPlutusScript` from `Gen.Cardano.Api.Typed`), then generate an `TxIns` which spends funds from these scripts, then generate a transaction with this `TxIns` (based on `genTxBodyContent`), run your `txScripts` on this generated tx, and verify that the initial generated plutus scripts are part of the function's output.

eyeinsky: It seems that the scripts I'm picking out and using are not included in the pattern TxBody: https://github.com/input-output-hk/cardano-node/blob/master/cardano-api/src/Cardano/Api/TxBody.hs#L2070

Question is, if the scripts that I pick out are the right ones? As I found also that inside TxBody the Script type also exist under:
- TxBodyContent{txOuts} -> TxOut -> ReferenceScript -> ScriptInAnyLang
- TxBodyContent{txReturnCollateral} -> TxReturnCollateral -> TxOut
- TxBodyContent{txAuxScripts} -> TxAuxScripts -> ScriptInEra

I would guess the scripts at the inputs (the ones that I'm using already and that come in as serialised) are the important ones, as those are the ones that are run.

eyeinsky: Another idea is that maybe plutus-streaming library (the one that provides the `withChainSyncEventStream` that provides the blocks) can convert scripts to the correct type such that only downstream using cardano-api is sufficient.

eyeinsky: @koslambrou Have managed to come up with this property test:
https://github.com/input-output-hk/plutus-apps/blob/edc8da432275c017ceba3a3aea24bbe9cb8e5e4b/plutus-chain-index-core/test/MarconiSpec.hs#L26-L66

There is still a piece missing: what does it mean to "spend a script that I created"? In the above I've tried to generate the same number of TxIns as there are scripts and replace the scripts' hashes into the `TxId` field of `TxIn` -- is this the correct way to think about "spending a script"?

eyeinsky:
Pushed another iteration where I create the [first tx](https://github.com/input-output-hk/plutus-apps/blob/5e6866f93bb7f73be179c793446a16dc7e37e4ab/plutus-chain-index-core/test/MarconiSpec.hs#L40-L57) in which the scripts I've generated are part of the outputs. Then create the [second transaction](https://github.com/input-output-hk/plutus-apps/blob/5e6866f93bb7f73be179c793446a16dc7e37e4ab/plutus-chain-index-core/test/MarconiSpec.hs#L59-L70) where the inputs are the outputs of the previous transaction, so I could "spend the scripts".

It's still very much WIP as the test can't possibly work as running [makeTransactionBody on the second transaction](https://github.com/input-output-hk/plutus-apps/blob/5e6866f93bb7f73be179c793446a16dc7e37e4ab/plutus-chain-index-core/test/MarconiSpec.hs#L70) doesn't have access to the generated scripts nor the hashes, because the inputs (`TxIn = TxId X TxIx`) only refer to the scripts indirectly.

Thus I wonder: is it implied that I need to run a local testnet in this property test too? so that when running the second transaction the TxIns could at least be looked up and the script field thus populated?

eyeinsky:
@koslambrou I've got the test passing and am now off to adding non-plutus scripts similar to here d253cfbd4b5449e9158da6d776a4c0a1f5f347fb, then will need to adapt the test as well, does that seem like the right direction?

eyeinsky:
Ready for review; I now index all types of scripts and also test them, and the tests pass.

plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
```diff
@@ -0,0 +1,193 @@
...
+    txScripts' = map (\tx -> (TxCbor $ C.serialiseToCBOR tx, getTxScripts tx)) txs
+
+getTxBodyScripts :: forall era . C.TxBody era -> [ScriptAddress]
+getTxBodyScripts body = let
+    hashesMaybe :: [Maybe C.ScriptHash]
+    hashesMaybe = case body of
+      Shelley.ShelleyTxBody shelleyBasedEra _ scripts _ _ _ -> flip map scripts $ \script ->
+        case fromShelleyBasedScript shelleyBasedEra script of
+          Shelley.ScriptInEra _ script' -> Just $ C.hashScript script'
+      _ -> [] -- ^ Byron transactions have no scripts
```
koslambrou, line 88: Remove the `^`

plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
```diff
@@ -0,0 +1,193 @@
...
+          map fst $ filter (\(_, addrs) -> scriptAddress' `elem` addrs) update
+
+        both :: [TxCbor]
+        both = buffered <> map (\(SQL.Only txCbor') -> txCbor') persisted
+
+      return both
+
+-- * Copy-paste
+--
+-- | TODO: Remove when the following function is exported from Cardano.Api.Script
```
koslambrou, line 138: Point the link to the PR here.

plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
```diff
@@ -0,0 +1,194 @@
...
+import Cardano.Ledger.ShelleyMA.Timelocks qualified as Timelock
+
+import RewindableIndex.Index.VSqlite (SqliteIndex)
+import RewindableIndex.Index.VSqlite qualified as Ix
+
+
+newtype Depth = Depth Int
+
+newtype ScriptAddress = ScriptAddress Shelley.ScriptHash
+  deriving (Show, Eq)
```
koslambrou, line 38: Why the newtype? Because there's no `Show` and `Eq` for `ScriptHash`?

koslambrou (approved):
LGTM!

plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
eyeinsky: It's so to not have it be an orphan instance of the ToField/FromField classes.

plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
```diff
@@ -0,0 +1,194 @@
...
+
+type ScriptTxIndex = SqliteIndex ScriptTxUpdate () Query Result
+
+
+toUpdate :: forall era . C.IsCardanoEra era => [C.Tx era] -> SlotNo -> ScriptTxUpdate
+toUpdate txs = ScriptTxUpdate txScripts'
+  where
+    txScripts' = map (\tx -> (TxCbor $ C.serialiseToCBOR tx, getTxScripts tx)) txs
+
+getTxBodyScripts :: forall era . C.TxBody era -> [ScriptAddress]
```
raduom, line 81: I think this may be useful to someone else. Should we not move this to something like `CardanoAPI`?

plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
```diff
@@ -0,0 +1,194 @@
...
+  SQL.execute_ c "CREATE INDEX IF NOT EXISTS script_address ON script_transactions (scriptAddress)"
+  pure ix
+
+  where
+    onInsert :: ScriptTxIndex -> ScriptTxUpdate -> IO [()]
+    onInsert _ix _update = pure []
+
+    store :: ScriptTxIndex -> IO ()
+    store ix = do
+      persisted <- Ix.getEvents $ ix ^. Ix.storage
```
raduom, line 109: We should not store the events, only what we have in the buffer.

plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
```diff
@@ -0,0 +1,194 @@
...
+      forM_ rows $
+        SQL.execute (ix ^. Ix.handle) "INSERT INTO script_transactions (scriptAddress, txCbor) VALUES (?, ?)"
+
+    query :: ScriptTxIndex -> Query -> [ScriptTxUpdate] -> IO Result
+    query ix scriptAddress' updates = do
+      persisted :: [SQL.Only TxCbor] <- SQL.query (ix ^. Ix.handle)
+        "SELECT txCbor FROM utxos WHERE scriptAddress = ?" (SQL.Only scriptAddress')
+
+      let
+        buffered :: [TxCbor]
```
raduom, line 126: When querying we also need to account for `buffered` events.

plutus-chain-index-core/test/MarconiSpec.hs
```diff
@@ -0,0 +1,320 @@
...
+  let (witnesses, scriptHashes) = unzip witnessesHashes
+      collateral = case collateralSupportedInEra era of
+        Just yes -> TxInsCollateral yes txIns
+        _        -> TxInsCollateralNone
+
+  txBody <- forAll $ genTxBodyWithTxIns era (zip txIns $ map BuildTxWith witnesses) collateral
+  let hashesFound = map coerce $ ScriptTx.getTxBodyScripts txBody :: [ScriptHash]
+  assert $ S.fromList scriptHashes == S.fromList hashesFound
+
+genTxBodyWithTxIns
```
raduom, line 51: Doesn't the node export generators for these data structures already?

raduom (commented):
I think there may be some bugs in the indexer.

plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
koslambrou: Yes! For now, I'd say we keep it here until we decide how to structure cardano-api related code. For example, might make sense to create a decidated `cardano-api-extended` package which contains a bunch of these functions which can then be pushed upstream when stabilized and tested.

plutus-chain-index-core/test/MarconiSpec.hs
koslambrou: Not exactly. They are very limited and would really need to be reworked on. Again, it would make sense to put these generators in a common package so that they can be pushed upstream.

raduom: My concern is that generators are pretty difficult to get right. You need to think about efficient ways of shrinking them, and what statistical distribution you need to test your scenarios, and how to check that the distribution you selected is actually testing those scenarios. I am not sure we are the ones that should be doing this.

koslambrou: I agree. The problem is that this is *not* a priority for the node team. We can probably do what the Hydra team did, use the generators from `cardano-ledger` and reconvert the datatype to a cardano-api type.

Ultimately, cardano-api would need to have good generators in them, and I think this is a cross-team effort (Hydra, Djed, Plutus Tools). So I think it's okay for us to start contributing in that direction IMO.

plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
eyeinsky: Maybe this could be done in another PR?

But I guess yes, if the function would just return `Shelley.ScriptHash` (`ScriptAddress` is a newtype around it) then I guess Cardano.Api.Shelley could be a good location for it.

plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
eyeinsky: I now only store the buffered events here 83ac994f4f7b457f6d5517adca3d590a070fbc09 and think there was the same bug in the Utxo indexer as well, does it look ok now there as well?

plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
eyeinsky: Here I get the transaction cbors for a script address from the sqlite database:
https://github.com/input-output-hk/plutus-apps/blob/83ac994f4f7b457f6d5517adca3d590a070fbc09/plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs#L121 and here I get transaction cbors from the buffered events' sequence https://github.com/input-output-hk/plutus-apps/blob/83ac994f4f7b457f6d5517adca3d590a070fbc09/plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs#L127 

I.e I do query both and return them from the function (?).

plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
eyeinsky: A related thought: if we will get a hoogle, then writing this function could be done in two steps: one would be `getTxBodyScriptHashes :: forall era . C.TxBody era -> [ScriptHash]` and the other one would be the function we currently have, which just wraps `ScriptHash` into a newtype. This way hooglers can find the function, and if useful enough, can move them into a more suitable module.

raduom: I think we can leave it for another PR.

plutus-chain-index-core/src/Marconi/Index/ScriptTx.hs
raduom: The `updates` parameter only contains the events though. So basically you need to do what you did before for `store`. The way this works right now is that we keep K events in memory plus some buffered events so we can batch updates to the database (which makes inserting rows much more efficient than if we would do it one by one).

So when we store things on disk, we only store the things that are buffered and leave the K blocks in memory (since they can still be rolled back); but when we query we want to account for all events, including the ones that are currently being buffered.

eyeinsky: Right, fixes here 1886e125f, 800d8f9a8

raduom (approved):

eyeinsky:
@koslambrou This MR is now ready merge!

PR #636, opened 2022-07-29
Move plutus-ledger to use V2 types where available

andreabedini:
This moves plutus-ledger to use plutus-ledger-api V2 types where available. Also adjusts a bit the dependecy between modules (making every plutus ledger reexport the plutus-ledger-api types it refers to, saving the consumer from having to explicitly depend on plutus-ledger-api).

Still in draft because there are some parts I am not sure how to solve. I'll leave comments inline.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-ledger/src/Ledger/Index.hs
+++ b/plutus-ledger/src/Ledger/Index.hs
@@ -42,10 +42,10 @@ module Ledger.Index(
     -- * Script validation events
     ScriptType(..),
     ScriptValidationEvent(..),
-    Api.ExBudget(..),
-    Api.ExCPU(..),
-    Api.ExMemory(..),
-    Api.SatInt,
+    ExBudget(..),
+    ExCPU(..),
+    ExMemory(..),
+    SatInt,
     ValidatorMode(..),
     getScript
     ) where
@@ -74,22 +74,15 @@ import Ledger.Crypto
 import Ledger.Index.Internal
 import Ledger.Orphans ()
 import Ledger.Params (Params (pSlotConfig))
+import Ledger.Scripts
 import Ledger.Slot qualified as Slot
 import Ledger.TimeSlot qualified as TimeSlot
 import Ledger.Tx
 import Ledger.Validation (evaluateMinLovelaceOutput, fromPlutusTxOutUnsafe)
-import Plutus.Script.Utils.V1.Scripts
+import Ledger.Value
 import Plutus.Script.Utils.V1.Scripts qualified as PV1
-import Plutus.V1.Ledger.Address
-import Plutus.V1.Ledger.Api qualified as Api
-import Plutus.V1.Ledger.Contexts (ScriptContext (..), ScriptPurpose (..), TxInfo (..))
-import Plutus.V1.Ledger.Contexts qualified as Validation
-import Plutus.V1.Ledger.Credential (Credential (..))
 import Plutus.V1.Ledger.Interval qualified as Interval
-import Plutus.V1.Ledger.Scripts
-import Plutus.V1.Ledger.Scripts qualified as Scripts
-import Plutus.V1.Ledger.Value qualified as V
-import PlutusTx (toBuiltinData)
+import Plutus.V2.Ledger.Api
 import PlutusTx.Numeric qualified as P
 
 -- | Context for validating transactions. We need access to the unspent
@@ -129,7 +122,7 @@ runValidation :: Validation (Maybe ValidationErrorInPhase) -> ValidationCtx -> (
 runValidation l ctx = runWriter $ fmap (either (\e -> Just (Phase1, e)) id) $ runExceptT $ runReaderT (_runValidation l) ctx
 
 -- | Determine the unspent value that a ''TxOutRef' refers to.
-lkpValue :: ValidationMonad m => TxOutRef -> m V.Value
+lkpValue :: ValidationMonad m => TxOutRef -> m Value
 lkpValue = fmap txOutValue . lkpTxOut
 
 -- | Find an unspent transaction output by its reference. Assumes that the
@@ -218,9 +211,9 @@ the blockchain.
 checkMintingAuthorised :: ValidationMonad m => Tx -> m ()
 checkMintingAuthorised tx =
     let
-        mintedCurrencies = V.symbols (txMint tx)
+        mintedCurrencies = symbols (txMint tx)
 
-        mpsScriptHashes = Scripts.MintingPolicyHash . V.unCurrencySymbol <$> mintedCurrencies
+        mpsScriptHashes = MintingPolicyHash . unCurrencySymbol <$> mintedCurrencies
 
         lockingScripts = Map.keys $ txMintScripts tx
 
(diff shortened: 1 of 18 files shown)


plutus-ledger/src/Ledger/Index.hs
```diff
@@ -378,24 +372,26 @@ mkTxInfo tx = do
...
-            , txInfoMint = txMint tx
-            , txInfoFee = txFee tx
-            , txInfoDCert = [] -- DCerts not supported in emulator
-            , txInfoWdrl = [] -- Withdrawals not supported in emulator
-            , txInfoValidRange = TimeSlot.slotRangeToPOSIXTimeRange slotCfg $ txValidRange tx
-            , txInfoSignatories = fmap pubKeyHash $ Map.keys (tx ^. signatures)
-            , txInfoData = Map.toList (tx ^. datumWitnesses)
-            , txInfoId = txId tx
+            { txInfoInputs          = txins
+            , txInfoReferenceInputs = mempty      -- FIXME
```
andreabedini, line 376: `txInfoReferenceInputs` and `txInfoRedeemers` are new fields. I haven't thought yet what goes in here.

plutus-ledger/test/Spec.hs
```diff
@@ -229,7 +229,8 @@ ciTxOutRoundTrip :: Property
 ciTxOutRoundTrip = property $ do
   txOuts <- Map.elems . Gen.mockchainUtxo <$> forAll Gen.genMockchain
   forM_ txOuts $ \txOut -> do
-    Hedgehog.assert $ Tx.toTxOut (fromJust $ Tx.fromTxOut txOut) == txOut
+    -- Hedgehog.assert $ Tx.toTxOut (fromJust $ Tx.fromTxOut txOut) == txOut
+    fail "FIXME" -- FIXME
```
andreabedini, line 233: As mentioned elsewhere `fromTxOut` is broken because `TxOut` do not carry the reference script (but only its hash).

plutus-ledger/src/Ledger/Tx.hs
```diff
@@ -107,94 +105,86 @@ data ChainIndexTxOut =
...
-        pure $ ScriptChainIndexTxOut txOutAddress txOutValue (Left dh) Nothing (Left vh)
+toTxOut :: ChainIndexTxOut -> TxOut
+toTxOut (PublicKeyChainIndexTxOut addr v outputDatum referenceScript) =
+  TxOut addr v outputDatum (fmap scriptHash referenceScript)
+toTxOut (ScriptChainIndexTxOut addr v (Left dh) referenceScript _validator) =
+  TxOut addr v (OutputDatumHash dh) (fmap scriptHash referenceScript)
+toTxOut (ScriptChainIndexTxOut addr v (Right d) referenceScript _validator) =
+  TxOut addr v (OutputDatum d) (fmap scriptHash referenceScript)
+
+-- fromTxOut :: TxOut -> Maybe ChainIndexTxOut
```
andreabedini, line 145: unfortunately we cannot make a `ChainIndexTxOut` out of a `TxOut` because the reference script is gone. We could return `Nothing` here but I don't thing the meaning of this function is clear at all. We need to revisit.

plutus-ledger/src/Ledger/Tx.hs
```diff
@@ -217,83 +207,84 @@ instance Pretty CardanoTx where
...
 
-getCardanoTxData :: CardanoTx -> Map V1.DatumHash V1.Datum
+getCardanoTxData :: CardanoTx -> Map.Map DatumHash Datum
 getCardanoTxData = onCardanoTx txData
-    (\(SomeTx (C.Tx (C.TxBody C.TxBodyContent {..}) _) _) ->
+    (\(CardanoAPI.SomeTx (C.Tx (C.TxBody C.TxBodyContent {..}) _) _) ->
         Map.fromList $ mapMaybe (\(C.TxOut _ _ d _) -> fromCardanoTxOutDatum d) txOuts)
     where
-        fromCardanoTxOutDatum :: C.TxOutDatum C.CtxTx era -> Maybe (V1.DatumHash, V1.Datum)
+      -- FIXME duplicated code
```
andreabedini, line 275: We already have a `fromCardanoTxOutDatum` somewhere with a slightly different signature.

plutus-ledger/src/Ledger/Tx.hs
```diff
@@ -15,11 +15,12 @@
...
 module Ledger.Tx
     ( module Ledger.Tx.Internal
-    , module Plutus.V1.Ledger.Tx
+    , module Plutus.V2.Ledger.Tx
+    , module Plutus.Script.Utils.V2.Scripts
     -- * ChainIndexTxOut
     , ChainIndexTxOut(..)
     , toTxOut
-    , fromTxOut
+    -- , fromTxOut FIXME we need to figure this out
```
andreabedini, line 23: See comment at `fromTxOut`

plutus-ledger/src/Ledger/Tx/CardanoAPI.hs
```diff
@@ -423,6 +424,21 @@ toCardanoTxBodyContent P.Params{P.pProtocolParams, P.pNetworkId} sigs P.Tx{..} =
         , txUpdateProposal = C.TxUpdateProposalNone
         }
 
+  where
+    -- FIXME what do we need to do here if
```
andreabedini, line 428: :facepalm:

> FIXME what do we need to do here if the `P.Tx` refers to a datum that is not in txData? Do we need to fail with a `ToCardanoError` or we can include the datum hash in the outputs anyway? Here I am just reproducing the previous behaviour which is include the hash anyway.

andreabedini: I moved `lookupDatum` here because it's an helper function to `toCardanoTxBodyContent`

plutus-ledger/src/Ledger/Tx/CardanoAPI.hs
```diff
@@ -480,52 +496,51 @@ toCardanoTxInWitness
...
 toCardanoTxOut
     :: C.NetworkId
-    -> (Maybe P.DatumHash -> Either ToCardanoError (C.TxOutDatum ctx C.BabbageEra))
-    -> PV1.TxOut
+    -> (OutputDatum -> Either ToCardanoError (C.TxOutDatum ctx C.BabbageEra))
+    -> (Maybe ScriptHash -> Either ToCardanoError (C.ReferenceScript C.BabbageEra))
+    -> TxOut
     -> Either ToCardanoError (C.TxOut ctx C.BabbageEra)
-toCardanoTxOut networkId fromHash (PV1.TxOut addr value datumHash) =
+toCardanoTxOut networkId fromOutputDatum fromScriptHash (TxOut addr value outputDatum refScript) =
```
andreabedini, line 527: I am extending the mechanism for datum lookups to the reference scripts, hoping we have a solution for it.

andreabedini:
After a discussion with @koslambrou and @sjoerdvisscher we are thinking of using `Cardano.Api.TxOut` instead.

Issue #471, opened 2022-05-22
Add missing fields into Tx.


zmrocze:
### Describe the feature you'd like

Currently the `Tx` type taken from `plutus-ledger-api` lacks:
 - [ ] metadata
 - [ ] withdrawals
 - [ ] certificates
 
These should be added together with updating:
  - `toCardanoTxBody` translation
  - validation in emulator
  - adding related constraints in `plutus-ledger-constraints`
  - ...

zmrocze:
I've added this issue as I lacked a place discussing changing the `Tx` and the old one taken as is from plutus-ledger-api is as I understand only temporary.

related issues:
 - metadata #93
  - withdrawals #395 
  - Should witnesses be stored on the side? #459

zmrocze:
I went with such a change: [Ledger/Tx/Internal/Tx.hs](https://github.com/zmrocze/plutus-apps/blob/e1248a02dc87b799aeba0869043fe27e9a3936ed/plutus-ledger/src/Ledger/Tx/Internal.hs#L78) here #468.

Making these changes: 
 - moving out all script and datum witnesses into seperate txScripts and txData fields, leaving hashes in inputs/etc. Before these were attached to inputs, but in a Maybe field, so this doesn't change that expected witnesses might or might not be present.
 - attaching redeemers to inputs/mints (conversely ^). Before the `Redeemers` map was very loosely related to scripts with `RedeemerPtr`'s, which makes it hard to manually construct transactions with more than 1 mint script. I took suggestion from TxConstraints where constraints contain redeemers but refer to scripts only by hashes. 
 - adding certificates/withdrawals/metadata fields
 
 I mostly fixed compiling and added withdrawals translation in `toCardanoTxBody` based on [plutus-simple-model](https://github.com/mlabs-haskell/plutus-simple-model/blob/main/src/Plutus/Test/Model/Fork/CardanoAPI.hs).
 
Adding validation for withdrawals/certificates is harder as `ValidationContext` doesn't contain state needed to validate these (like rewards/delegations/registered credentials).

I would enjoy greatly discussing the expected shape for `Tx` and this proposition for it.

ak3n:
Closing because of https://github.com/input-output-hk/plutus-apps/issues/459#issuecomment-1485138902

PR #468, opened 2022-05-20
Update tx with missing fields.

zmrocze:
`Tx` lacks metadata/withdrawals/certificates fields #471. 
Together with adding these I refactored script and datum witnesses out of inputs/etc into seperate maps [see #459 ]. 

Pre-submit checklist:
- Branch
    - [ ] Tests are provided (if possible)
    - [ ] Commit sequence broadly makes sense
    - [ ] Key commits have useful messages
    - [x] Relevant tickets are mentioned in commit messages
    - [x] Formatting, materialized Nix files, PNG optimization, etc. are updated
- PR
    - [x] Self-reviewed the diff
    - [x] Useful pull request description
    - [ ] Reviewer requested

This pr introduces these changes:

### Redefines `Tx` type
 - `Tx` gets withdrawals, certificates(1), metadata(1) fields
 - script and datum witnesses get put into maps and refered to by hashes ([related comment](https://github.com/input-output-hk/plutus-apps/issues/471#issuecomment-1133941543) and [issue](#459) )
 - redeemers get put together with the MintingPolicy/Certificate/Input (^ same links)
 - defines `TxInput` and `TxInputType` simmilar to `TxIn` and `TxInType`.
 - defines `Withdrawal` type
 - defines `Certificate` type
 - splits Ledger.Tx.Internal module

### Updates pab accordingly
 - `TxIn` remains used by some functions only in auxiliary way
 - most changes are automatic: imports and minor updates that don't change logic only update the `Tx`
 - `toCardanoTxBodyContent` from Ledger.Tx.CardanoAPI gets updated with withdrawal translation (based on [plutus-simple-model](https://github.com/mlabs-haskell/plutus-simple-model/blob/main/src/Plutus/Test/Model/Fork/CardanoAPI.hs)).

### Updates purescript accordingly
 - code generation generates added types: `Certificate`, `Withdrawal`, `TxInput`, `TxIn`
 - minor tweak in purescript source code

### Certificate and Metadata is wrong for now (1)
 - `DCert` in `Certificate` has some data like `StakePoolParameters` in registration certificate digested. Best would be to reproduce [cardano-api/`Certificate`](https://github.com/input-output-hk/cardano-node/blob/6a465b09b5dfd5fe281061b94f69faee1227800a/cardano-api/src/Cardano/Api/Certificate.hs#L80) but using plutus-friendly types.
[...]

Issue #459, opened 2022-05-13
Tx type is inconsistent in how it stores data witnesses.


zmrocze:
### Summary

There are few inconsistencies in the [`Tx`](https://github.com/input-output-hk/plutus-apps/blob/f250a7ffb89e2554c08aff6b7105a77c0d75e85a/plutus-ledger/src/Ledger/Tx/Internal.hs#L34) type and now it's been moved into plutus-apps. Possibly all of them stem only from my misunderstanding. `Tx` lacks support for withdrawals or certifications which I'd like to add, but probably it is better to first address these.

### 1. Should datums be stored on the side in `Map DatumHash Datum` field or instead packed together with inputs?

`TxIn` for now CAN contain datum as:
```haskell
data TxInType =
      -- TODO: these should all be hashes, with the validators and data segregated to the side
      ConsumeScriptAddress !Validator !Redeemer !Datum -- ^ A transaction input that consumes a script address with the given validator, redeemer, and datum.
    | ConsumePublicKeyAddress -- ^ A transaction input that consumes a public key address.
    | ConsumeSimpleScriptAddress -- ^ Consume a simple script

-- | A transaction input, consisting of a transaction output reference and an input type.
data TxIn = TxIn {
    txInRef  :: !TxOutRef,
    txInType :: Maybe TxInType
    }
```
The TODO comment is not mine. 
But `Transaction` also has the `txData` field for various datums (which ones exactly if datums can be stored with TxIn?). 
So which is better: 
a) datum in TxIn or 
b) datum in a map and referenced by hash in TxIn'?

Also could someone explain at what stage is can the `txInType` field be `Nothing` and when is it expected to be `Just ...`? For example `toCardanoTxBody` fails if it's `Nothing`.

### 2. Should validators be stored on the side in seperate container or instead packed together with inputs/withdrawals/etc?

This is simmilar to the previous one but here validator scripts apply not only to inputs but also minting, certyfying and withdrawing.

Currently input validators are stored alongside inputs ^, minting policies are stored in a set and withdrawals or delegation certificates are not yet accounted for.

Should we:
a) keep input validators with inputs, certyficate validators with certyficates, withdrawal validator with withdrawals, minting validators in a set
b) input validators in one set, certyficate validators in second set, ..., minting policies in 4th set
c) all scripts in a single set
d) other combination of above solutions

[...]

PR #637, opened 2022-07-29
Update tx with missing fields [main].

zmrocze:
`Tx` lacks metadata/withdrawals/certificates fields #471. 
Together with adding these I refactored script and datum witnesses out of inputs/etc into seperate maps [see #459 ]. 

*This pr is made cherry-picking from #468 that targeted `next-node`.* 

I expect imports/exports needing to be fixed.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

## Notes for reviewer

 - Now one can import `Tx` both from `Plutus.V1.Ledger.Api` and from `Ledger` where first one is wrong. 
 - `Plutus.V1.Ledger.Api` re-exports from `Ledger` are messy. 
    Everything having to do with old `Tx` should not be re-exported from `Ledger`, but what about rest? Can we just leave users to import things from `Plutus.V1.Ledger` if its not in `Ledger`?
 - `TxInput` superseded `TxIn` in `Tx` but it is still used as auxiliary type here and there.

This pr introduces these changes:

### Redefines `Tx` type
 - `Tx` gets withdrawals, certificates(1), metadata(1) fields
 - script and datum witnesses get put into maps and refered to by hashes ([related comment](https://github.com/input-output-hk/plutus-apps/issues/471#issuecomment-1133941543) and [issue](#459) )
 - redeemers get put together with the MintingPolicy/Certificate/Input (^ same links)
 - defines `TxInput` and `TxInputType` simmilar to `TxIn` and `TxInType`.
 - defines `Withdrawal` type
 - defines `Certificate` type
 - splits Ledger.Tx.Internal module

### Updates pab accordingly
 - `TxIn` remains used by some functions
 - most changes are automatic: imports and minor updates that don't change logic only update the `Tx`
 - `toCardanoTxBodyContent` from Ledger.Tx.CardanoAPI gets updated with withdrawal translation (based on [plutus-simple-model](https://github.com/mlabs-haskell/plutus-simple-model/blob/main/src/Plutus/Test/Model/Fork/CardanoAPI.hs)).

### Updates purescript accordingly
 - code generation generates added types: `Certificate`, `Withdrawal`, `TxInput`, `TxIn`
 - minor tweak in purescript source code

### Certificate and Metadata is wrong for now (1)
 - `DCert` in `Certificate` has some data like `StakePoolParameters` in registration certificate digested. Best would be to reproduce [cardano-api/`Certificate`](https://github.com/input-output-hk/cardano-node/blob/6a465b09b5dfd5fe281061b94f69faee1227800a/cardano-api/src/Cardano/Api/Certificate.hs#L80) but using plutus-friendly types.
 - Metadata is more like `Map Word64 JSONLikeValue`, definitely not bytestring. Best would be [cardano-api/`TxMetadata`](https://github.com/input-output-hk/cardano-node/blob/6a465b09b5dfd5fe281061b94f69faee1227800a/cardano-api/src/Cardano/Api/TxMetadata.hs#L83), but we can't use for example Word64 onchain, correct? We could reproduce cardano-api type using plutus-friendly types.


zmrocze:
summarising I need feedback on two issues from above: 

 - what should `Ledger` re-export from `Plutus.V1.Ledger` given that 1) `Tx` shouldn't be, because confusion 2) all export will be sort of manual (with explicit hiding or explicit exporting)
 
  - what to use for `DCert` and `Metadata` types? Should we define in `Ledger` analogues of `Cardano.Api` types, just using Plutus friendly type analogues? Or use `Cardano.Api` types itself (not plutus friendly).

PR #640, opened 2022-08-02
PLT-445 Add `mustReferencePubKeyOutput` in constraints library

koslambrou:
* Added the `mustReferencePubKeyOutput` constraint in `plutus-ledger-constraints` and `plutus-tx-constraints`
    
* Added a test case for the `mustReferencePubKeyOutput`, but they are not run in CI until we remove our custom transaction validation rules in favor of using fully `cardano-ledger` validation rules
    
* Moved `TxIn` and `TxInType`  from `plutus-ledger-api` into `plutus-ledger` (these types will be removed from `plutus`)

* Added plutus version to `ConsumeScriptAddress` constructor of `TxInType`

* Renamed `Ledger.Index.mkTxInfo` into `mkPV1TxInfo`

* Added reference inputs field in `Ledger.Tx.Internal`

* Added constraints `MustUseOutputCollateral,` `MustSpendScriptOutput` and `MustPayToOtherScript` constraints in order to implement the test case with `plutus-tx-constraints` 

Pre-submit checklist:
- Branch
    - [ ] Tests are provided (if possible)
    - [ ] Commit sequence broadly makes sense
    - [ ] Key commits have useful messages
    - [ ] Formatting, PNG optimization, etc. are updated
- PR
    - [ ] Self-reviewed the diff
    - [ ] Useful pull request description
    - [ ] Reference the ADR in the PR and reference the PR in the ADR (if revelant)
    - [ ] Reviewer requested

PR #641, opened 2022-08-03
Add mustSpendAtLeast Constraint tests

james-iohk:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-contract/test/Spec/MustSpendAtLeast.hs
+++ b/plutus-contract/test/Spec/MustSpendAtLeast.hs
@@ -9,8 +9,6 @@
 module Spec.MustSpendAtLeast(tests) where
 
 import Control.Monad (void)
-import Data.Map qualified as Map
-import Data.Void (Void)
 import Test.Tasty (TestTree, testGroup)
 
 import Ledger qualified
@@ -18,18 +16,18 @@ import Ledger.Ada qualified as Ada
 import Ledger.Constraints.OffChain qualified as Constraints (MkTxError (OwnPubKeyMissing), plutusV1OtherScript,
                                                              plutusV1TypedValidatorLookups, unspentOutputs)
 import Ledger.Constraints.OnChain.V1 qualified as Constraints (checkScriptContext)
-import Ledger.Constraints.TxConstraints qualified as Constraints (mustIncludeDatum, mustPayToTheScript,
-                                                                  mustSpendAtLeast, mustSpendScriptOutput)
+import Ledger.Constraints.TxConstraints qualified as Constraints (collectFromTheScript, mustIncludeDatum,
+                                                                  mustPayToTheScript, mustSpendAtLeast)
 import Ledger.Tx qualified as Tx
 import Ledger.Typed.Scripts qualified as Scripts
 import Plutus.Contract as Con
-import Plutus.Contract.Test (assertContractError, assertValidatedTransactionCount, checkPredicateOptions,
-                             defaultCheckOptions, w1, (.&&.))
+import Plutus.Contract.Test (assertContractError, assertFailedTransaction, assertValidatedTransactionCount,
+                             checkPredicateOptions, defaultCheckOptions, w1, (.&&.))
 import Plutus.Trace qualified as Trace
 import Plutus.V1.Ledger.Api (Datum (Datum), ScriptContext, Validator, ValidatorHash)
-import Plutus.V1.Ledger.Scripts (unitRedeemer)
+import Plutus.V1.Ledger.Scripts (ScriptError (EvaluationError))
 import PlutusTx qualified
-import PlutusTx.Prelude as P (traceError)
+import PlutusTx.Prelude qualified as P
 import Prelude hiding (not)
 
 tests :: TestTree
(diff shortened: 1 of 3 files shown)


plutus-contract/test/Spec/MustSpendAtLeast.hs
```diff
@@ -0,0 +1,116 @@
...
+
+higherThanScriptBalance :: TestTree -- This should fail with a more accurate contract error than "OwnPubKeyMissing"
+higherThanScriptBalance =
+    let amt = scriptBalance + 1
+    in  checkPredicateOptions
+            defaultCheckOptions
+            "Fail validation when mustSpendAtLeast is greater than script's balance"
+            (assertContractError (contract amt) (Trace.walletInstanceTag w1) (\case { ConstraintResolutionContractError Constraints.OwnPubKeyMissing -> True; _ -> False }) "failed to throw error"
+            .&&. assertValidatedTransactionCount 1)
+            (void $ trace amt)
```
james-iohk, line 93: @sjoerdvisscher Do you agree that this contract error is inaccurate for the scenario?

koslambrou: It's actually accurate considering the implementation. In the library, if the total `Value` of the outputs is higher than the `Value` of the inputs, then the library will try to compensate the difference by adding inputs with funds from the pubkey hash provided in the lookups (e.g. `OwnPubKey`).

Typically, you would call `ownPaymentPubKeyHash` in the `Contract` and add it in the lookups with `paymentPubKey`.

plutus-contract/test/Spec/MustSpendAtLeast.hs
```diff
@@ -0,0 +1,116 @@
...
+
+{-# INLINEABLE mkValidator #-}
+mkValidator :: Integer -> () -> () -> ScriptContext -> Bool
```
koslambrou, line 96: Did you actually meant `mkValidator :: Integer -> () -> ScriptContext -> Bool`?

Because when trying to unlock the script, you used `Constraints.mustIncludeDatum (Datum $ PlutusTx.toBuiltinData amt)`

plutus-contract/test/Spec/MustSpendAtLeast.hs
```diff
@@ -0,0 +1,116 @@
...
+        tx1 = Constraints.mustPayToTheScript () (Ada.lovelaceValueOf scriptBalance)
+    ledgerTx1 <- submitTxConstraintsWith lookups1 tx1
+    awaitTxConfirmed $ Tx.getCardanoTxId ledgerTx1
+    utxos <- utxosAt $ scrAddress amt
+    let orefs = fst <$> Map.toList utxos
+        lookups2 =
+            Constraints.plutusV1OtherScript (validatorScript amt)
+            <> Constraints.unspentOutputs utxos
+        tx2 =
+            foldMap (\oref -> Constraints.mustSpendScriptOutput oref unitRedeemer) orefs
```
koslambrou, line 58: Use `Constraints.collectFromPlutusV1Script (Map.singleton scrAddress utxos) (validatorScript amt) unitRedeemer`

plutus-contract/test/Spec/MustSpendAtLeast.hs
```diff
@@ -0,0 +1,116 @@
...
+            <> Constraints.mustIncludeDatum (Datum $ PlutusTx.toBuiltinData amt)
```
koslambrou, line 59: Necessary?

koslambrou:
I would also add a test case which results in a failing validation script.

koslambrou:
I would also take add (duplicate?) the test cases for the `plutus-tx-constraints` library. We do something similar in here #640

plutus-contract/test/Spec/MustSpendAtLeast.hs
james-iohk: Thanks for the explanation, I still think UX could be improved though.

plutus-contract/test/Spec/MustSpendAtLeast.hs
james-iohk: Changed to use datum

plutus-contract/test/Spec/MustSpendAtLeast.hs
james-iohk: It is now

plutus-contract/test/Spec/MustSpendAtLeast.hs
```diff
@@ -0,0 +1,127 @@
...
+contract :: Integer -> Integer -> Contract () Empty ContractError ()
+contract offAmt onAmt = do
+    let lookups1 = Constraints.plutusV1TypedValidatorLookups typedValidator
+        tx1 = Constraints.mustPayToTheScript onAmt (Ada.lovelaceValueOf scriptBalance)
+    ledgerTx1 <- submitTxConstraintsWith lookups1 tx1
+    awaitTxConfirmed $ Tx.getCardanoTxId ledgerTx1
+
+    utxos <- utxosAt scrAddress
+    let lookups2 = Constraints.plutusV1TypedValidatorLookups typedValidator
+            <> Constraints.plutusV1OtherScript validatorScript
```
koslambrou, line 54: I don't think this is needed if you're using `Constraints.plutusV1TypedValidatorLookups typedValidator`

plutus-contract/test/Spec/MustSpendAtLeast.hs
```diff
@@ -0,0 +1,127 @@
...
+            (assertValidatedTransactionCount 2)
+            (void $ trace amt amt )
+
+higherThanScriptBalance :: TestTree
+higherThanScriptBalance =
+    let amt = scriptBalance + 1
+    in  checkPredicateOptions
+            defaultCheckOptions
+            "Fail validation when mustSpendAtLeast is greater than script's balance"
+            (assertContractError (contract amt amt) (Trace.walletInstanceTag w1) (\case { ConstraintResolutionContractError Constraints.OwnPubKeyMissing -> True; _ -> False }) "failed to throw error"
```
koslambrou, line 90: This expected result doesn't fit with the description "Fail validation when mustSpendAtLeast is greater than script's balance".

If you locked 25ADA in the script, and you want the transaction to spend 26ADA, the 1ADA that is left will be compensated by the wallet provided in the lookups.

Then, you can either change the test case description, or add the pubkey in the lookups, and show that we don't get a failing test (or add a new test case for that)

james-iohk: I've split this test out into two, using `ownPaymentPubKeyHash` lookup. They are now both failing due to `ValueContainsLessThanMinAda` by 1 lovelace.

- `higherThanScriptBalanceWithWalletPubkeyLookup` is using the correct own wallet PaymentPubKeyHash. Shouldn't it try and abide by this rule?

- `higherThanScriptBalanceWithoutWalletPubkeyLookup` is using other wallet's PaymentPubKeyHash. Shouldn't it just fail with `OwnPubKeyMissing`?

koslambrou: Yes! That's because you can't create an output with less than (I think it's hardcoded to 2) the minAda required for a tx output. You need to make sure that the amount that the provided pubkey needs is higher than that value.

Do `let amt = scriptBalance + 5` instead of `let amt = scriptBalance + 1`

james-iohk: Ok, changing to `+ 5` is only lovelace so same error. Chaning to `+ 5_000_000` results in passing contract validation but failing phase-2. Are there two problems here:
1. Constraint passing off-chain validation but failing on-chain
2. `higherThanScriptBalanceWithoutWalletPubkeyLookup` doesn't use w1 pubkeyhash so I'd expect it to throw `OwnPubKeyMissing` error.

james-iohk: Bug identified in offchain code, see PLT-665
Have commented out the two affected tests for now.

koslambrou (approved):

PR #648, opened 2022-08-05
Fix onchain MustPayToPubKeyAddress check

sjoerdvisscher:
Returning `True` in an `any` function meant that the check would succeed if there was any other output.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


ak3n (approved):
Good catch!

PR #654, opened 2022-08-09
PLT-682 ADR for the constraint interface change following support for reference inputs

koslambrou:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/doc/adr/0007-support-reference-inputs-in-constraint-library.rst
+++ b/doc/adr/0007-support-reference-inputs-in-constraint-library.rst
@@ -1,0 +1,47 @@
+.. _support_reference_inputs_in_constraint_library:
+
+ADR 7: Support reference inputs in constraint library
+=====================================================
+
+Date: 2022-08-09
+
+Authors
+-------
+
+koslambrou <konstantinos.lambrou@iohk.io>
+
+Status
+------
+
+Proposed
+
+Context
+-------
+
+After the Vasil HF, the Cardano blockchain will support `reference` inputs by adding a new field in the transaction data type.
+With reference inputs, transactions can take a look at UTXOs without actually spending them.
+
+Thus, we need to adapt our transaction constraint data type (`TxConstraints`) to support referencing UTXOs.
+
+Decision
+--------
+
+* We will add the data constuctor `MustReferenceOutput TxOutRef` to the `TxConstraints` data type.
+
+* The PlutusV1 on-chain implementation of this new constraint will simply return `False`.
+  However, `cardano-ledger` throws a phase-1 validation error if transactions that use the some of the new features (reference inputs, inline datums and reference scripts) try to execute PlutusV1 scripts.
+  See the `Babbage era ledger specification <https://hydra.iohk.io/job/Cardano/cardano-ledger/specs.babbage-ledger/latest/download-by-type/doc-pdf/babbage-changes>`_.
+  Therefore, the only way to get a phase-2 validation error would be to use this constraint on-chain in a PlutusV1 script, without using any of the new Babbage era features off-chain.
+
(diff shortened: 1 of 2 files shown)


doc/adr/0007-support-reference-inputs-in-constraint-library.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Decision
+--------
+
+* We will add the data constuctor `MustReferenceOutput TxOutRef` to the `TxConstraints` data type.
+
+* The PlutusV1 on-chain implementation of this new constraint will simply return `False`.
+  However, `cardano-ledger` throws a phase-1 validation error if transactions that use the some of the new features (reference inputs, inline datums and reference scripts) try to execute PlutusV1 scripts.
+  See the `Babbage era ledger specification <https://hydra.iohk.io/job/Cardano/cardano-ledger/specs.babbage-ledger/latest/download-by-type/doc-pdf/babbage-changes>`_.
+  Therefore, this on-chain implementation should **never** result in a phase-2 validation error.
```
ak3n, line 34: What does it mean that it will never result in a phase-2 validation error? It's important that it will be caught during the first phase?

koslambrou: Basically, if the user tries to submit a transaction which:
- uses reference inputs, inline datums or reference scripts
- includes a V1 scripts

This will result in a phase-1 error by `cardano-ledger` - meaning that the transaction will fail before being submitted in the blockchain. If it was a phase-2 error, it means that the transaction will be part of the blockchain and will be failing (meaning collateral inputs will be used).

doc/adr/0007-support-reference-inputs-in-constraint-library.rst
```diff
@@ -0,0 +1,47 @@
...
+
+* The PlutusV2 on-chain implementation of this new constraint will check that the provided `TxOutRef` is part of the `ScriptContext`'s reference inputs.
+
+Argument
+--------
+
+At first glance, we might think that we need two data constructors for reference inputs such as `MustReferencePubKeyOutput` and `MustReferenceScriptOutput` in contrast to the existing `MustSpendPubKeyOutput` and `MustSpendScriptOutput` constraints.
+However, we do not need to make the distinction between public key outputs and script outputs because we're not spending the output, therefore, we don't need to provide a redeemer nor the actual script as a witness to the transaction input.
```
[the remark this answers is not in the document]
koslambrou: Yeah exactly. That's what I'm saying. We can just have `MustReferenceOutput TxOutRef` because we don't need to know the difference between a pubkey and script output.

doc/adr/0007-support-reference-inputs-in-constraint-library.rst
ak3n: Oh, didn't know about the validation phases, thanks!

ak3n (approved):

doc/adr/0007-support-reference-inputs-in-constraint-library.rst
james-iohk, line 34: The ledger will only throw this phase-1 error in the presence of reference inputs, so the on-chain implementation should only result in a phase-2 validation error in the absence of the offchain constraint.

james-iohk (approved):

doc/adr/0007-support-reference-inputs-in-constraint-library.rst
koslambrou: Oh right! I'll rephrase this. Thanks :)

koslambrou: How about the following?

```suggestion
  Therefore, the only to get a phase-2 validation error would be to use this constraint on-chain, without using any of the new features off-chain.
```

PR #655, opened 2022-08-09
PAB <> Blockfrost integration

TomasOsiecki:
# Introduction
The following PR adds the possibility of using the PAB with Blockfrost instead of the Chain-Index. The interface is exactly the same, but internally the PAB connects with a different service. Therefore, any function implementation using the Contract effects doesn't need to be adjusted. Checkout [README](https://github.com/joinplank/plutus-apps/blob/plank-pab-blockfrost/plutus-blockfrost/README.md)

# Implementation details
One of the main challenges of this implementation was related to not changing the API defined on the plutus-chain-index-core library. So, this involves respecting the input and output format of each handler implementation. Our general strategy was to implement different input and output process functions. In that way, the general approach for implementing each handler, of what we call a chain-query action, is divided into three main parts: translate the input (`Utils` module), then call the blockfrost action(s) (`Queries` module), and finally process the output (`Responses` module). For instance, for the DatumFromHash handler we have to first translate the input with `toBlockfrostDatumHash`, then call the blockfrost action with `getDatumBlockfrost` and finally process the output with `processGetDatum`.

Snippet from Plutus.Blockfrost.Client module
```haskell
...
case event of
   DatumFromHash d -> (runClientMaybe . getDatumBlockfrost . toBlockfrostDatumHash) d >>= processGetDatum
...
```

# Disclaimers
Throughout the implementation a couple of decisions were made to keep everything for the sake of simplicity:
- The `UtxoSetAtAddress`, `UtxoSetWithCurrency`, `TxoSetAtAddress`, and `UnspentTxOutSetAtAddress` don't support pagination on the blockfrost implementation, we simply return the complete result considering all pages.
- For the `validatorFromHash`, `mintingPolicyFromHash`, and `stackeValidatorFromHash` actions, we found some strange behavior with the blockfrost response. The solution was to drop the first 6 chars. Maybe this is a bug from the blockfrost implementation and we need to stay alert for any news related with that.

Pre-submit checklist:
- Branch
    - [ ]  Tests are provided (if possible)
    - [x]  Commit sequence broadly makes sense
    - [x]  Key commits have useful messages
    - [x]  Formatting, PNG optimization, etc. are updated
- PR
    - [x]  Self-reviewed the diff
    - [x]  Useful pull request description
    - [ ]  Reference the ADR in the PR and reference the PR in the ADR (if revelant)
    - [ ]  Reviewer requested

Co-Authored-By: Valentino Cerutti <valentino.cerutti@joinplank.com>
Co-Authored-By: Ale Gadea <ale.gadea@joinplank.com>


koslambrou (approved):

sorki:
Nicely done!

> * For the `validatorFromHash`, `mintingPolicyFromHash`, and `stackeValidatorFromHash` actions, we found some strange behavior with the blockfrost response. The solution was to drop the first 6 chars. Maybe this is a bug from the blockfrost implementation and we need to stay alert for any news related with that.

I'm late to the party but this is fine - the response is serialized via `cardano-api` which is why there seem to be additional chars - they are just a tag for e.g. `Api.AsPlutusScript Api.AsPlutusScriptV1` (check out deserialization in my [old attempt at PAB + Blockfrost integration](https://github.com/cardano-foundation/dab/pull/9/files#diff-bb033e78cd5c36115a104d5d36b38ec3fbec660b2ca26f220b1c75cd92d1d5f1R212-R222)). This could be useful for handling V2 scripts but you can also use `getScript` query from `blockfrost-client` to check for script version.

Please also consider adding upper bounds for `blockfrost-api` and `blockfrost-client`.

PR #662, opened 2022-08-15
Replace LedgerPlutusVersion with Language type

sjoerdvisscher:
@michaelpj said we shouldn't be using `LedgerPlutusVersion`. We can use `Cardano.Ledger.Alonzo.Language` instead.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


ak3n (approved):

michaelpj (approved):
So you don't actually... use any values of this? Seems good though.

sjoerdvisscher:
@michaelpj Yes we do, but the values happen to have exactly the same name so they don't show up in the changes.

PR #665, opened 2022-08-17
PLT-670 Add  MustValidateIn to tx constraints

berewt:
\[WIP] The tests part is still missing but the validityRanges are now directly store in the Tx, not in an extra field.   

Pre-submit checklist:
- Branch
    - [X] Formatting, PNG optimization, etc. are updated
- PR
    - [X] Self-reviewed the diff
    - [X] Reviewer requested

### DIFF
--- a/plutus-contract/test/Spec/TxConstraints/TimeValidity.hs
+++ b/plutus-contract/test/Spec/TxConstraints/TimeValidity.hs
@@ -47,7 +47,7 @@ tests = testGroup "time validitity constraint"
         -- trigger the validator
         -- , protocolV5Cardano
         , defaultProtocolParamsValidCardano
-        , defaultProtocolParamsInvalidCardano
+        , defaultProtocolParamsPastTxCardano
         ]
     ]
 
@@ -127,8 +127,8 @@ validContractCardano p = do
     P.unless (cSlot `I.member` txRange) $ P.traceError "InvalidRange"
 
 
-invalidContractCardano :: Ledger.Params -> Contract () Empty ContractError ()
-invalidContractCardano p = do
+pastTxContractCardano :: Ledger.Params -> Contract () Empty ContractError ()
+pastTxContractCardano p = do
     let mkTx lookups constraints = either (error . show) id $ Tx.Constraints.mkTx @UnitTest p lookups constraints
     pkh <- Con.ownFirstPaymentPubKeyHash
     utxos <- Con.ownUtxos
@@ -141,8 +141,10 @@ invalidContractCardano p = do
         tx2 =  Tx.Constraints.mustPayToPubKey pkh (Ada.toValue Ledger.minAdaTxOut)
             <> Tx.Constraints.mustSpendPubKeyOutput utxoRef2
             <> Tx.Constraints.mustValidateIn (I.to now)
+    -- submit a first transaction to occupy the first slot
     ledgerTx1 <- submitUnbalancedTx $ mkTx lookups tx1
     awaitTxConfirmed $ Tx.getCardanoTxId ledgerTx1
+    -- submit a tx that should be validated at the latest in slot 1
     ledgerTx2 <- submitUnbalancedTx $ mkTx lookups tx2
     awaitTxConfirmed $ Tx.getCardanoTxId ledgerTx2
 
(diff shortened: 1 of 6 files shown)


koslambrou:
For this PR, I would recreate the same tests in `Spec.TxConstraints.TimeValidity`, but for `plutus-tx-constraints` instead of `plutus-ledger-constraints` (keep the old one until we finally remove it).

berewt:
Got a failure I'm unable to explain in the `protocolV5Cardano` test.
The transaction pass despite the validityRange, not sure if it's normal due to how we build the transaction in Tx.Constraints or not.

plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
```diff
@@ -230,9 +233,9 @@ ownStakePubKeyHash skh = mempty { slOwnStakePubKeyHash = Just skh }
 -- | An unbalanced transaction. It needs to be balanced and signed before it
 --   can be submitted to the ledger. See note [Submitting transactions from
 --   Plutus contracts] in 'Plutus.Contract.Wallet'.
-data UnbalancedTx =
-    UnbalancedTx
-        { unBalancedTxTx                  :: Either C.CardanoBuildTx Tx.Tx
+data UnbalancedTx
+    = UnbalancedLedgerTx
```
sjoerdvisscher, line 237: `UnbalancedLedgerTx` sounds like it's a transaction from `cardano-ledger`. `UnbalancedEmulatorTx` is probably a better name here.

plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
```diff
@@ -250,30 +253,49 @@ data UnbalancedTx =
...
         vsep
-        [ hang 2 $ vsep ["Tx:", either pretty pretty utx]
+        [ hang 2 $ vsep ["Tx:", pretty utx]
         , hang 2 $ vsep $ "Requires signatures:" : (pretty <$> Set.toList rs)
         , hang 2 $ vsep $ "Utxo index:" : (pretty <$> Map.toList utxo)
         , hang 2 $ vsep ["Validity range:", pretty vr]
         ]
+    pretty (UnbalancedCardanoTx utx rs utxo) =
+        vsep
+        [ hang 2 $ vsep ["Tx (Cardano Representation):", pretty utx]
```
sjoerdvisscher, line 295:
```suggestion
        [ hang 2 $ vsep ["Tx (cardano-api representation):", pretty utx]
```

plutus-contract/test/Spec/TxConstraints/TimeValidity.hs
```diff
@@ -86,6 +95,65 @@ defaultProtocolParams = checkPredicateOptions
...
+    now <- Con.currentTime
+    logInfo @String $ "now: " ++ show now
+    let utxoRef = fst $ head' $ Map.toList utxos
+        lookups =  Constraints.plutusV1TypedValidatorLookups (typedValidator deadline)
+                <> Tx.Constraints.unspentOutputs utxos
+        tx  =  Tx.Constraints.mustPayToPubKey pkh (Ada.toValue Ledger.minAdaTxOut)
+            <> Tx.Constraints.mustSpendPubKeyOutput utxoRef
+            <> Tx.Constraints.mustValidateIn (f now)
+    submitTxConfirmed $ mkTx lookups tx
+
```
sjoerdvisscher, line 117: Would be nice to read out the time again here and check it's within `f now`.

sjoerdvisscher (changes requested):

berewt:
`protocolV5 test` keeps failing with `Tx.Constraints`. I wonder if it can come from the way we test in the two cases: we check that the transaction is validated within its validity range in the `Tx.Constraints` case, while we are checking the validity range that is stored in `ScriptContext` in the initial case. Unfortunately, I can't use the validator as not enough `Tx.Constraints` are available on `main`.

sjoerdvisscher:
I'm fine with removing the V5 test, as long as we have a good negative test that shows the positive test doesn't succeed by accident.

berewt:
It leads to the other issue with the PR at this stage: the failing test defines an unreachable range for a `Tx` and checks that the `Tx` isn't validated after some slot. Unfortunately, the transaction isn't failing either, it's just waiting for a valid slot that doesn't (and can't) come.
I think it's ok, but maybe we  (I) should open an issue to investigate whether this behavior should be fixed.

plutus-contract/test/Spec/TxConstraints/TimeValidity.hs
```diff
@@ -111,23 +113,43 @@ contractCardano f p = do
...
     logInfo @String $ show txRange
 
     P.unless (cSlot `I.member` txRange) $ P.traceError "InvalidRange"
 
-validContractCardano :: Ledger.Params -> (Contract () Empty ContractError) ()
-validContractCardano = contractCardano (from . (1000 +))
 
-invalidContractCardano :: Ledger.Params -> (Contract () Empty ContractError) ()
-invalidContractCardano = contractCardano I.to
+invalidContractCardano :: Ledger.Params -> Contract () Empty ContractError ()
```
sjoerdvisscher, line 130: Can you add some documentation to explain how this test works?

plutus-contract/test/Spec/TxConstraints/TimeValidity.hs
```diff
@@ -154,11 +176,13 @@ defaultProtocolParamsValidCardano = checkPredicateOptions
     (assertValidatedTransactionCount 1)
     (void $ traceCardano $ validContractCardano $ view (emulatorConfig . params) defaultCheckOptions)
 
+-- We only test here if the contract ends with no valid transaction.
```
sjoerdvisscher, line 172: This comment does not match the test `assertValidatedTransactionCount 1`

sjoerdvisscher (approved):
Thanks, that makes a lot of sense with the explanation!

PR #672, opened 2022-08-22
Add event channel to certification function

UlfNorell:
You can pass an optional `Control.Concurrent.Chan` in the certification options to get events when tests pass or fail.

Pre-submit checklist:
- Branch
    - [ ] ~~Tests are provided (if possible)~~
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-contract-certification/src/Plutus/Contract/Test/Certification/Run.hs
+++ b/plutus-contract-certification/src/Plutus/Contract/Test/Certification/Run.hs
@@ -119,7 +119,7 @@ data CertificationTask = UnitTestsTask
                        | CrashToleranceTask
                        | WhitelistTask
                        | DLTestsTask
-  deriving (Eq, Show)
+  deriving (Eq, Show, Enum, Bounded)
 
 data CertificationOptions = CertificationOptions { certOptNumTests  :: Int
                                                  , certOptOutput    :: Bool
--- a/plutus-contract-certification/src/Plutus/Contract/Test/Certification/Run.hs
+++ b/plutus-contract-certification/src/Plutus/Contract/Test/Certification/Run.hs
@@ -32,6 +32,7 @@ module Plutus.Contract.Test.Certification.Run
   , CertificationOptions(..)
   , CertificationEvent(..)
   , CertificationTask(..)
+  , hasQuickCheckTests
   , defaultCertificationOptions
   , certify
   , certifyWithOptions
@@ -120,6 +121,9 @@ data CertificationTask = UnitTestsTask
                        | WhitelistTask
                        | DLTestsTask
   deriving (Eq, Show, Enum, Bounded)
+
+hasQuickCheckTests :: CertificationTask -> Bool
+hasQuickCheckTests t = t /= UnitTestsTask
 
 data CertificationOptions = CertificationOptions { certOptNumTests  :: Int
                                                  , certOptOutput    :: Bool
--- a/plutus-contract-certification/src/Plutus/Contract/Test/Certification/Run.hs
+++ b/plutus-contract-certification/src/Plutus/Contract/Test/Certification/Run.hs
@@ -30,11 +30,14 @@ module Plutus.Contract.Test.Certification.Run
   , certRes_DLTests
   -- * and we have a function for running certification
   , CertificationOptions(..)
+  , CertificationEvent(..)
+  , CertificationTask(..)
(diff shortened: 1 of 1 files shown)


UlfNorell:
@shlevy

plutus-contract-certification/src/Plutus/Contract/Test/Certification/Run.hs
```diff
@@ -103,11 +107,29 @@ makeLenses ''CertificationReport
...
+
+data CertificationTask = UnitTestsTask
+                       | StandardPropertyTask
+                       | DoubleSatisfactionTask
+                       | NoLockedFundsTask
+                       | NoLockedFundsLightTask
+                       | CrashToleranceTask
+                       | WhitelistTask
+                       | DLTestsTask
+  deriving (Eq, Show)
```
shlevy, line 122: Can we get `Enum` (reflecting the actual run order) and `Bounded`? That way if more stages are added I can still give an overall X/Y number until I update my source.

shlevy: Also can we get something like `renderCertificationTask :: CertificationTask -> Text/String` that's nicer for formatting to the user, again in the case where this repo updates before I've had a chance to update source on my end?

shlevy: If QC runs in more than one phase, can we get something like `hasQCTests :: CertificationTask -> Bool`?

plutus-contract-certification/src/Plutus/Contract/Test/Certification/Run.hs
```diff
@@ -103,11 +107,29 @@ makeLenses ''CertificationReport
 certResJSON :: CertificationReport m -> String
 certResJSON = unpack . encode
 
-data CertificationOptions = CertificationOptions { certOptNumTests :: Int
-                                                 , certOptOutput   :: Bool }
+data CertificationEvent = QuickCheckTestEvent (Maybe Bool)  -- ^ Nothing if discarded, otherwise test result
+                        | StartCertificationTask CertificationTask
```
shlevy, line 111: Rather than start events, would it be difficult to switch to end events with a result? Then we can infer start from when the previous stage ends (or when I first call certify)

shlevy (changes requested):
Mainly reviewed with respect to interface, let me know if you'd like me to review implementation as well.

plutus-contract-certification/src/Plutus/Contract/Test/Certification/Run.hs
UlfNorell: I don't think formatting for the end user is the responsibility of this code, and would be much better suited somewhere else.

plutus-contract-certification/src/Plutus/Contract/Test/Certification/Run.hs
UlfNorell: What result are you looking for here? Would a boolean pass/fail be sufficient?

shlevy: Yes that would be great!

plutus-contract-certification/src/Plutus/Contract/Test/Certification/Run.hs
shlevy: @UlfNorell OK, that's fine

plutus-contract-certification/src/Plutus/Contract/Test/Certification/Run.hs
MaximilianAlgehed: Technically that information is already there in the `QuickCheckTestEven` constructor though?

MaximilianAlgehed: I suppose you don't get it for the unit tests though...

MaximilianAlgehed: I propose start and end events to make the consumer less dependent on the internals of this function. I've implemented this and I'll push it soon...

shlevy (approved):

MaximilianAlgehed:
@koslambrou could we get this merged?

PR #690, opened 2022-09-01
PLT-454: mustUseOutputAsCollateral

sjoerdvisscher:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
+++ b/plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
@@ -693,10 +693,10 @@ processConstraint = \case
             unbalancedTx . tx . Tx.redeemers . at (RedeemerPtr Spend (fromIntegral idx)) .= Just red
             valueSpentInputs <>= provided value
           _ -> throwError (TxOutRefWrongType txo)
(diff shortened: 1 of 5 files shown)


koslambrou:
1- This should be merged on `main` instead, I think

2- I would add a test using the validation rules (similar to `Spec/TxConstraints/MustMint` et al.). We do use it in `Spec.Contract.TxConstraints`, but maybe it makes sense to test it as a standalone constraint? @james-iohk What do you think?

koslambrou (approved):

james-iohk:
> 2- I would add a test using the validation rules (similar to `Spec/TxConstraints/MustMint` et al.). We do use it in `Spec.Contract.TxConstraints`, but maybe it makes sense to test it as a standalone constraint? @james-iohk What do you think?

We currently only use it for a single positive scenario but we're not checking for what happens when, for example: input has a token, doesn't exist, provided by other wallet, failing phase-2 validation. In future we'll want to test in combination with return collateral. It would be good to include these test scenarios, if not now then raise a ticket and I'll get to it.

ak3n (approved):

PR #691, opened 2022-09-02
New shiny indexer

raduom:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


eyeinsky:
As the new data families and classes are meant to be reused then it would be great to add comments to them as to what they do and how implementing them will add up to a functioning indexer.

A second test for fit of purpose would be to implement any of the existing indexers using these.

---

The streaming part isn't (at least to me) obvious from this code -- what about having the spine of an indexer based on the `streaming` library as something like this (pseudocode):

```haskell

import Streaming.Prelude qualified as Streaming
import Cardano.Streaming (ChainSyncEvent)
import Marconi.Streaming qualified as Common
import SomeSpecificIndexer qualified as Indexer


instance HasPoint Event

indexer :: Stream (Of Indexer.Event) IO a
indexer =
  source
    -- 1
    & toEvent
    -- 2
    & handleRollbacks 2160
    -- 3
    & storeSomewhere
    -- 4

  where
    source :: Stream (Of (ChainSyncEvent (BlockInMode CardanoMode))) IO r
    source = chainSyncFrom socketPath networkId point
    -- ^ Get a stream of blocks from socket.

    toEvent :: Stream (Of (ChainSyncEvent (BlockInMode CardanoMode))) IO r
            -> Stream (Of (ChainSyncEvent Indexer.Event)) IO r
    toEvent = Indexer.toEvent
    -- ^ Convert blocks from chain sync to events that are indexer
    -- specific. Note that rollbacks are still unhandled.

    handleRollbacks
      :: HasPoint a
      => Int
      -> Stream (Of (ChainSyncEvent a)) IO r
      -> Stream (Of a) IO r
    handleRollbacks = Common.ringbuffer
    -- ^ Peel off the ChainSyncEvent wrapper (it's defined
    -- approximately as `ChainSyncEvent a = RollForward a ChainPoint |
    -- Rollback ChainPoint)` which functions as a sponge up to @n@
    -- filling the buffer, starting to yield events that are past
    -- the rollback horizon.
    --
    -- This can either be the in-memory ringbuffer implemented as
    -- vector, or something that persists even the temporary events to
    -- disk. The implementation doesn't leak into the type, it just
    -- handles rollbacks.

    storeSomewhere
      :: Stream (Of Indexer.Event) IO r
      -> Stream (Of Indexer.Event) IO r
    storeSomewhere = Indexer.storeInSqlite "<name-of-db>"
    -- ^ Write the events to disk. Also yield the events onwards after
    -- they are persisted.

-- | Run this indexer in IO
indexer_ :: IO r
indexer_ = Streaming.effects indexer
```

The above has the benefits of
- for any use case where rollbacks are not importand, e.g rapid development, then line with `handleRollbacks` can simply be replaced by `Streaming.mapMaybe (\case (RollForward a Point) -> Just a; _ -> Nothing)`.
- rollback handling by any given method is not coupled into the indexer (just replace `handleRollbacks` with another implementation, e.g one that stores to disk or database).
- method of storing events is swapped by replacing `storeSomewhere`
- all parts of the event pipeline have isolated purpose and are testable separately
- the "stages of permanence" at which events are, are clear (temporary, permanent in blockchain, persisted to disk by the indexer). To monitor them at any of these then locations 2, 3, or 4 could respectively be used to add something like `Streaming.chain (\event -> .. write to some concurrent variable ..)`.
- the event "monitor" above could be an HTTP server.
  - If the server needs to send push events (notifications), then the same locations can be used to concurrently channel events to the HTTP server's thread.
  - If it answers queries across all indexed events, then for
    - persisted events the database would need to be queried in a separate thread.
    - if also for in-memory events, then the rollback handler would need to be extended with its own concurrent variable to communicate with the HTTP server's thread.
- as with the `streaming` package, events are processed one by one, so it will also DRY up the [`RollBackward` case in `innerLoop`](https://github.com/input-output-hk/plutus-apps/blob/6efc2e12df1815f9ea4e01bdaf477caf8b7b0225/marconi/src/Marconi/Indexers.hs#L116-L223), remove the need for the Coordinator and also probably the `..Worker` functions in general.

rewindable-index/test/Spec.hs
```diff
@@ -26,6 +27,20 @@ ixProperties = testGroup "Basic model"
       withMaxSuccess 1000 $ Ix.prop_insertRewindNotifications @Int @Int @Int Ix.conversion
   ]
 
+sProperties :: TestTree
+sProperties = testGroup "New index properties."
+  [ testProperty "New: Positive or non-positive depth" $
+      withMaxSuccess 10000 $ Ix.prop_observeNew @Int @Int S.conversion
```
kayvank, line 33: Should we increase the depth range for this to make sure we test closer to the actual depth value? The max depth for tests is 100, where is the indexers value is 2160

rewindable-index/test/RewindableIndex/Spec/Sqlite.hs
```diff
@@ -0,0 +1,289 @@
...
+                                 StorablePoint, StorableQuery, StorableResult, config, emptyState,
+                                 filterWithQueryInterval, memoryBufferSize)
+import RewindableIndex.Storable qualified as Storable
+
+newtype Handle = Handle Sql.Connection
+
+data instance StorablePoint Handle =
+    Point Int
+  | Genesis
+  deriving stock (Eq, Show, Generic)
```
kayvank, line 35: necessary to add stock? that is the default, yes?

kayvank, line 35: `stock` is default here and not required yes? I think it is good to put it in and in that case we should add it to `StorableEvent`, few lines below, as well

rewindable-index/test/RewindableIndex/Spec/Sqlite.hs
```diff
@@ -0,0 +1,289 @@
...
+monadic
+  :: PropertyM IO Property
+  -> Property
+monadic = monadicIO
+
+lookupPoint
+  :: Int
+  -> Config
+  -> MaybeT IO (StorablePoint Handle)
+-- TODO: this needs a lot of comments
```
kayvank, line 252: yes, plz as I am having hard time following the logic here

kayvank (approved):

eyeinsky:
(Will write comments in here and in code. Starting)

Data families: still, what is the purpose of these? For example how is this:

```haskell
data family Orderable a

class Ord_ a where
  compare_ :: Orderable a -> Orderable a -> Ordering

newtype instance Orderable Int = Int_ Int

instance Ord_ Int where
  compare_ (Int_ a) (Int_ b) = compare a b
```

better than the simple this:

```haskell
class Ord_ a where
  compare_ :: a -> a -> Ordering

instance Ord_ Int where.
  compare_ a b = Prelude.compare a b
```

Why the extra hoop to jump through? A type class encodes what the type is able to do, a data family adds ... ?

eyeinsky:
A second issue with data families now that I've tried: I think the `h` parameter (e.g for `StorableEvent`) might need to be some fresh type created for some particular indexer (`data UtxoAddressToSqliteIndexer`) and derive all the rest from that type. This is because a `Handle` type couldn't be reused anyway (there can't be multiple `StorableEvent Handle`s in scope), but we _would_ like to run our indexers simultaneously. (Does this make sense?)

eyeinsky:
Also, the duplication of [rollback handling](https://github.com/input-output-hk/plutus-apps/blob/6efc2e12df1815f9ea4e01bdaf477caf8b7b0225/marconi/src/Marconi/Indexers.hs#L116-L223) mentioned above will still remain, given the new indexer, perhaps it could be DRYied up somehow.

eyeinsky:
Done for now.

raduom:
> (Will write comments in here and in code. Starting)
> 
> Data families: still, what is the purpose of these? For example how is this:
> 
> ```haskell
> data family Orderable a
> 
> class Ord_ a where
>   compare_ :: Orderable a -> Orderable a -> Ordering
> 
> newtype instance Orderable Int = Int_ Int
> 
> instance Ord_ Int where
>   compare_ (Int_ a) (Int_ b) = compare a b
> ```
> 
> better than the simple this:
> 
> ```haskell
> class Ord_ a where
>   compare_ :: a -> a -> Ordering
> 
> instance Ord_ Int where.
>   compare_ a b = Prelude.compare a b
> ```
> 
> Why the extra hoop to jump through? A type class encodes what the type is able to do, a data family adds ... ?

If the use-case is a type class with only one variable, then there is not much use for it, but in this case that is not what we have to work with. Our type class(es) would generally have type parameters for the monad in which they run (1), the type of the database handle (2), the type of events it handles (3), the type of points (4), the type of queries (5) and the type of results (6).

5 of these variables semantically depend on the way we store things in the database. It is unlikely that we will have a different database structure (that includes the database and the tables) that stores the same kind of events, mostly because if that is the case combining the indexers makes a lot more sense.

So here is a list of reasons why I think type families are useful here:

1. It simplifies the signature of the type classes by only including 2 type variables rather than 6.
2. It provides a better description of the meaning of what used to be one letter variable (`StorablePoint` vs `p`)
3. It points to a dependency between the said variables (to which I already alluded).

There is also another hidden benefit to using data families (type families will not work very well for this). If we decide to switch from several classes to only one and gather all the data under that definition. This is not the case right now, but it is a possibility.

raduom:
> A second issue with data families now that I've tried: I think the `h` parameter (e.g for `StorableEvent`) might need to be some fresh type created for some particular indexer (`data UtxoAddressToSqliteIndexer`) and derive all the rest from that type. This is because a `Handle` type couldn't be reused anyway (there can't be multiple `StorableEvent Handle`s in scope), but we _would_ like to run our indexers simultaneously. (Does this make sense?)

This is mentioned in the comments here: https://github.com/input-output-hk/plutus-apps/pull/691/files#diff-b87396626c72b11a1e23d16c58fca2fa643ce7dc06f0f414d4a356e377453ed8R57-R61 and https://github.com/input-output-hk/plutus-apps/pull/691/files#diff-e36d392fc09a3195f1fe88e074c41529140b496fe07891a78248108a93a23574R41-R42. I am not sure it's an issue or an opportunity. I think it's an opportunity for users of the API to include a specific configuration for their indexer, since the `h` parameter is passed into all the functions.

rewindable-index/test/RewindableIndex/Spec/Sqlite.hs
raduom: There used to be a lot more deriving statements there and I forgot to remove the stock when I removed those. Good catch!

raduom:
> Also, the duplication of [rollback handling](https://github.com/input-output-hk/plutus-apps/blob/6efc2e12df1815f9ea4e01bdaf477caf8b7b0225/marconi/src/Marconi/Indexers.hs#L116-L223) mentioned above will still remain, given the new indexer, perhaps it could be DRYied up somehow.

Couldn't we use https://github.com/input-output-hk/plutus-apps/pull/691/files#diff-b87396626c72b11a1e23d16c58fca2fa643ce7dc06f0f414d4a356e377453ed8R284-R310 for all indexers?

rewindable-index/test/Spec.hs
raduom: I cannot see any scenario in which that would make a difference. I think that most issues will be found at boundaries. Making the boundaries be far apart or closer together would not influence that, would it?

rewindable-index/test/RewindableIndex/Spec/Sqlite.hs
raduom: Done.

eyeinsky:
> If the use-case is a type class with only one variable, then there is not much use for it, but in this case that is not what we have to work with. Our type class(es) would generally have type parameters for the monad in which they run (1), the type of the database handle (2), the type of events it handles (3), the type of points (4), the type of queries (5) and the type of results (6).
> 
> 5 of these variables semantically depend on the way we store things in the database. It is unlikely that we will have a different database structure (that includes the database and the tables) that stores the same kind of events, mostly because if that is the case combining the indexers makes a lot more sense.
> 
> So here is a list of reasons why I think type families are useful here:
> 
>     1. It simplifies the signature of the type classes by only including 2 type variables rather than 6.
> 
>     2. It provides a better description of the meaning of what used to be one letter variable (`StorablePoint` vs `p`)
> 
>     3. It points to a dependency between the said variables (to which I already alluded).
> 
> 
> There is also another hidden benefit to using data families (type families will not work very well for this). If we decide to switch from several classes to only one and gather all the data under that definition. This is not the case right now, but it is a possibility.

There is such a thing as associated type families and associated data families which tie the related types together. Something like
```
class Indexer h where
  type StorableEvent h
  type Monad_ h :: Type -> Type -- otherwise it's inferred as Type
  -- .. and other types

  persistToStorage :: Foldable f => f (StorableEvent h) -> h -> (Monad_ h) h
  getStoredEvents :: h -> m [StorableEvent h]
  rewindStorage :: StorableEvent h -> h -> (Monad_ h) (Maybe h)
  
  -- .. and other functions
```

To implement:
```
data UtxoAddressIndexer = UtxoAddressIndexer Sql.Connection
instance Indexer UtxoAddressIndexer where
  type StorableEvent UtxoAddressIndexer = Marconi.Index.Utxo.UtxoUpdate -- for example
  type Monad_ UtxoAddressIndexer = IO
  -- .. and so on
```

The above is done using type families, but data families could be used as well. But data families can be inconvenient if you already have the event type defined elsewhere -- now you need to make a data family instance as well with no extra benefit.

eyeinsky:
> Couldn't we use https://github.com/input-output-hk/plutus-apps/pull/691/files#diff-b87396626c72b11a1e23d16c58fca2fa643ce7dc06f0f414d4a356e377453ed8R284-R310 for all indexers?

I guess the answer is no then because `rewind` is already used in the linked location as well.

There is a case statement with `RollForward`/`RollBackward` in every indexer's worker, why does each of them handle rollbacks separately, can't it be done uniformly?

raduom:
> There is such a thing as associated type families and associated data families which tie the related types together. Something like
> 
> ```haskell
> class Indexer h where
>   type StorableEvent h
>   type Monad_ h :: Type -> Type -- otherwise it's inferred as Type
>   -- .. and other types
> 
>   persistToStorage :: Foldable f => f (StorableEvent h) -> h -> (Monad_ h) h
>   getStoredEvents :: h -> m [StorableEvent h]
>   rewindStorage :: StorableEvent h -> h -> (Monad_ h) (Maybe h)
>   
>   -- .. and other functions
> ```
> 
> To implement:
> 
> ```haskell
> data UtxoAddressIndexer = UtxoAddressIndexer Sql.Connection
> instance Indexer UtxoAddressIndexer where
>   type StorableEvent UtxoAddressIndexer = Marconi.Index.Utxo.UtxoUpdate -- for example
>   type Monad_ UtxoAddressIndexer = IO
>   -- .. and so on
> ```
> 
> The above is done using type families, but data families could be used as well. But data families can be inconvenient if you already have the event type defined elsewhere -- now you need to make a data family instance as well with no extra benefit.

As far as I know associated types/data families are syntactic sugar for the unassociated ones. But indeed, there are associated types. What I don't understand is: Are they superior to the current implementation? If so, in which way? If your point is that we could use only one type class for the indexer, that is true and something I considered, but I also think that there is value in having several interfaces across the functionalities that you want to support (some indexers may not care about rollbacks, others about resume, etc).

Another difference between data and type families is that type families are not injective whereas data families are. That is the reason why I chose data families for this implementation. I also think that it's more likely than not that the events the indexer will handle will need to be defined within the indexer module, so it seemed like a good compromise to make. I think data families are easier to understand and work with (lack of injectivity can be surprising to a less experienced developer).

raduom:
> > Couldn't we use https://github.com/input-output-hk/plutus-apps/pull/691/files#diff-b87396626c72b11a1e23d16c58fca2fa643ce7dc06f0f414d4a356e377453ed8R284-R310 for all indexers?
> 
> I guess the answer is no then because `rewind` is already used in the linked location as well.
> 
> There is a case statement with `RollForward`/`RollBackward` in every indexer's worker, why does each of them handle rollbacks separately, can't it be done uniformly?

Indexers were not really meant to be bunched together. Initially the idea was that we could even distribute indexers across different machines. In that case why would we want to optimise for scenarios where we are using them together?

Even if that was the case it is not clear to me what you mean by handling rollbacks in a consistent way across all indexers. I think that a 'generic' rewind method (based on the `Rewindable` constraint implemented by an indexer) could be called with whatever indexer implements that constraint. Wouldn't that count as 'unformly'? Each indexer will have to rollback it's on-disk data in different ways in this current iteration of the design. So that part of the code cannot be 'uniform' unless we choose to go with another design.

I guess I am not sure I understand what you mean by 'uniform' (or why doesn't the `rewind` function cover that).

eyeinsky:
> As far as I know associated types/data families are syntactic sugar for the unassociated ones. But indeed, there are associated types. What I don't understand is: Are they superior to the current implementation? If so, in which way? If your point is that we could use only one type class for the indexer, that is true and something I considered, but I also think that there is value in having several interfaces across the functionalities that you want to support (some indexers may not care about rollbacks, others about resume, etc).

I don't think they are syntactic sugar because you can't define an associated type or data family outside of the class, and that would also be the reason to have them in the class: there is no point in having a `StorableEvent` of something where there is no associated instance declaration that actually does the storing. Also, the data families stand separately, so one could define instances for one of these without defining the other, and yet still not defining a instance that does any of the storing.

> Another difference between data and type families is that type families are not injective whereas data families are. That is the reason why I chose data families for this implementation. I also think that it's more likely than not that the events the indexer will handle will need to be defined within the indexer module, so it seemed like a good compromise to make. I think data families are easier to understand and work with (lack of injectivity can be surprising to a less experienced developer).

If two separate indexers have the same event type what would be the problem? Isn't it the opposite that if you already had a type that would fit as the event then now you still need to create yet another data instance for it?

eyeinsky:
> Indexers were not really meant to be bunched together. Initially the idea was that we could even distribute indexers across different machines. In that case why would we want to optimise for scenarios where we are using them together?
> 
> Even if that was the case it is not clear to me what you mean by handling rollbacks in a consistent way across all indexers. I think that a 'generic' rewind method (based on the `Rewindable` constraint implemented by an indexer) could be called with whatever indexer implements that constraint. Wouldn't that count as 'unformly'? Each indexer will have to rollback it's on-disk data in different ways in this current iteration of the design. So that part of the code cannot be 'uniform' unless we choose to go with another design.
> 
> I guess I am not sure I understand what you mean by 'uniform' (or why doesn't the `rewind` function cover that).

Right, so in the Marconi.Indexers there is a worker for each indexer, and each of these indexers has a case statement to branch out forward and rewind, and the rewind branch does exactly the same thing for each of them. So to uniformly handle rollbacks would be to handle rollbacks before any of the workers receives the event to insert it.

If we have a way to stick a rollback handler in front of an indexer then we don't optimise for a specific scenario, but will have composable way to either handle indexers together, or not. Currently we seem to optimise for the scenario where we definitely don't want to handle them together, but we for some reason do in Marconi.Indexers. (Maybe that is a use case we never want?)

raduom:
> I don't think they are syntactic sugar because you can't define an associated type or data family outside of the class, and that would also be the reason to have them in the class: there is no point in having a `StorableEvent` of something where there is no associated instance declaration that actually does the storing. Also, the data families stand separately, so one could define instances for one of these without defining the other, and yet still not defining a instance that does any of the storing.

I don't know where I got that idea from, but indeed they are not the same. `StorableEvent` is used in almost all other classes, though. And someone can define instances only for some data families, if only those are needed by the type class that they want to implement. I think that is fine.
 
> If two separate indexers have the same event type what would be the problem? Isn't it the opposite that if you already had a type that would fit as the event then now you still need to create yet another data instance for it?

Yes. What I am saying is that the second scenario is fairly unlikely. Taking a look at existing code I don't think it has happened so far, but if it would a `newtype instance` wrapper is not really all that bad imho if we get injectivity.

raduom:
> > I guess I am not sure I understand what you mean by 'uniform' (or why doesn't the `rewind` function cover that).
> 
> Right, so in the module Marconi.Indexers there is a worker for each indexer, and each of these indexers has a case statement to branch out forward and rewind, and the rewind branch does exactly the same thing for each of them. So to uniformly handle rollbacks would be to handle rollbacks before any of the workers receives the event to insert it.

That is indeed one of the things we are trying to fix long-term. To have a more scalable solution for indexing the blockchain. Going for indexers that can be optionally deployed separately seems like a much better idea than running as many as possible in the same process.

> If we have a way to stick a rollback handler in front of an indexer then we don't optimise for a specific scenario, but will have composable way to either handle indexers together, or not. Currently we seem to optimise for the scenario where we definitely don't want to handle them together, but we for some reason do in Marconi.Indexers. (Maybe that is a use case we never want?)

What would that handler do though? It needs to delete different tables from different databases using different code for each, right? How would you make that generic? How would you update the internal indexer data to say something like reset your internal state to slot X then resume operation if that information never reaches the indexer?

That may have worked for previous indexers which did not store any immutable data on disk (so we can have an immutable, generic rollback of in-memory data), but in our case, now, we do store mutable data on-disk. And we do that because the memory requirements for storing all the volatile data in memory are fairly harsh and as the blockchain gets larger, they will only get higher.

> While looking at the indexer workers, each of the indexers have their own ring buffer (as it comes inside the indexer), so we now have n instances of it. I also think that we actually do have a way to run them together, because if we say that any indexer would have it's own executable then any user who wants more than one thing to be indexed will still run several of these executables, and thus will still have several instances of a ring buffer.

But.. the instances of the ring buffers will store different events, so I don't see that as a problem. Also, the current implementation does not really use a ring buffer anymore. It uses a vector, but simply as a linear buffer, when filled it's flushed.

eyeinsky:
> Indexers were not really meant to be bunched together. Initially the idea was that we could even distribute indexers across different machines. In that case why would we want to optimise for scenarios where we are using them together?

The node and db-sync are already run on the same machine, so it's likely that marconi will too (?). If each indexer has it's own event type then indeed each indexer would need it's own buffer. But if we had blocks in the buffer then that could be re-used across all downstream indexers.

> Even if that was the case it is not clear to me what you mean by handling rollbacks in a consistent way across all indexers. I think that a 'generic' rewind method (based on the `Rewindable` constraint implemented by an indexer) could be called with whatever indexer implements that constraint. Wouldn't that count as 'unformly'? Each indexer will have to rollback it's on-disk data in different ways in this current iteration of the design. So that part of the code cannot be 'uniform' unless we choose to go with another design.
> 
> I guess I am not sure I understand what you mean by 'uniform' (or why doesn't the `rewind` function cover that).

Uniformly in the sense that there is only one set of blocks that need to be rolled back when multiple indexers are run together. This would mean that all indexers would need to be downstream from the rollback handler.

raduom:
> The node and db-sync are already run on the same machine, so it's likely that marconi will too (?). If each indexer has it's own event type then indeed each indexer would need it's own buffer. But if we had blocks in the buffer then that could be re-used across all downstream indexers.

Not necessarily. Marconi is much more light-weight than db-sync, and with the query interface taking slot intervals you can fairly easily distribute them across different machines. The only downside would be connecting to the node via a network, which does affect the synchronisation speed (the speed can be improved).

How would you share those buffers across different machines on the network? This design would make the whole system less scalable, which is one of the main design goals of Marconi.

> > Each indexer will have to rollback it's on-disk data in different ways in this current iteration of the design. So that part of the code cannot be 'uniform' unless we choose to go with another design.
> 
> Uniformly in the sense that there is only one set of blocks that need to be rolled back when multiple indexers are run together. This would mean that all indexers would need to be downstream from the rollback handler.

How would the rollback handler know how to alter the on-disk database for all the indexers, generically?

eyeinsky:
> Not necessarily. Marconi is much more light-weight than db-sync, and with the query interface taking slot intervals you can fairly easily distribute them across different machines. The only downside would be connecting to the node via a network, which does affect the synchronisation speed (the speed can be improved).

It doesn't matter how light-weight Marconi is if the data is gotten over the network. db-sync isn't recommended to run on the same machine as node because it's so heavyweight, but because there is no network latency so it's faster.

> How would you share those buffers across different machines on the network? This design would make the whole system less scalable, which goes against one of the main design goals of Marconi.

When run on different machines then they won't share buffers, but currently there is no way to share buffers _even when_ running on the same machine.

raduom:
> It doesn't matter how light-weight Marconi is if the data is gotten over the network. db-sync isn't recommended to run on the same machine as node because it's so heavyweight, but because there is no network latency so it's faster.

This was one of the main complaints that our users had though, so I think it does matter.
 
> When run on different machines then they won't share buffers, but currently there is no way to share buffers _even when_ running on the same machine.

... and if you think that is a worthwhile thing to do, then you can go ahead, propose an ADR, write up an implementation and go from there. Personally, I don't think it is worth the trouble, but that is just my opinion. However, this PR is not really about sharing the in-memory data or about switching the whole implementation to use streaming.

eyeinsky:
I think this PR would be much better if there would be any of our real indexers implemented. As it stands it would be tough for even me to use it.

There are comments above unanswered, few of them low hanging fruit and easy to implement.

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -0,0 +1,358 @@
...
+   The following data families implement the observation that most of the variables
+   can be derived from the way information is stored in the database, so there is
+   quite a bit of convenience in reducing the number of type variables to just two.
+   One that stands for the database connection (which should be a newtype for each
+   indexer) and one for the monad in which the indexer runs (usually IO).
+-}
+data family StorableEvent h
+
+data family StorablePoint h
+
```
berewt, line 66: We've discussed it already on Slack but yes, having one`StorablePoint` per indexer sounds like a bad idea if we want to work with several of them while being sure they have indexed the same thing.

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -0,0 +1,358 @@
...
+              & V.freeze <&> V.toList
+  diskEs   <- getStoredEvents (s ^. handle)
+  pure $ diskEs ++ memoryEs
+
+insert
+  :: Buffered h m
+  => PrimMonad m
+  => StorableEvent h
+  -> State h m
+  -> m (State h m)
```
berewt, line 241: Did you consider using a stack monad to handle `Config` and `State`? It looks similar to `MonadReader` and `MonadState` if we decompose them and it may simplify the logic.
I even wonder if in the end, `h` implementation shouldn't be hidden in the definition of `m` and then the ground type of your family definition would be the event.

berewt (commented):
I agree with @eyeinsky that the PR would benefit of a real world example to see how it can be applied to our code base. We can give it a shot, but I fear that the use of data families make the use of a set of indexers cumbersome (I didn't take the time to experiment it yet).

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -0,0 +1,358 @@
...
+   can be derived from the way information is stored in the database, so there is
+   quite a bit of convenience in reducing the number of type variables to just two.
+   One that stands for the database connection (which should be a newtype for each
+   indexer) and one for the monad in which the indexer runs (usually IO).
+-}
+data family StorableEvent h
+
+data family StorablePoint h
+
+data family StorableQuery h
```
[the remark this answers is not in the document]
raduom: You can think of queries by analogy with SQL queries. A query is a specification of a search across the database. It is not saved in the database, it's just a set of constraints on a specific view of the database. It is not specified in the CLI, but the client would need to send a query via a JSON-RPC or rest endpoint.

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -0,0 +1,358 @@
...
+        efficient than what Haskell can do here.
+     3) We can always extend the interface to include direct support for this pattern
+        if there is a demand from the users of the API.
+-}
+class Buffered h m where
+  -- This function persists the memory/buffer events to disk.
+  persistToStorage :: Foldable f => f (StorableEvent h) -> h -> m h
+  {- This function retrieves the events from the disk/events area.
+     If the user chooses to only store events, without accumulating them, this function
+     is expected to return the events over which rollbacks can occur in order to keep
```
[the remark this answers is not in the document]
raduom: The events stored in memory are also transient, but there is only one set of transient events that are stored on disk.

rewindable-index/test/Spec.hs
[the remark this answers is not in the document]
raduom: This PR is not about refactoring the testing infrastructure though..

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -0,0 +1,358 @@
...
+   disk/events is greater than the K parameter.
+
+   The last part disk/accumulator represents some aggregated data. At some point the
+   developer should write code to aggregate the disk/events into the accumulator
+   section of the database. We did not provide any special handling of this for the
+   following reasons:
+
+     1) A lot of indexers will not require aggregated data. Only storing events should
+        be enough for most applications.
+     2) Data aggregation can be implemented at database level and probably be more
```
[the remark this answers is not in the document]
raduom: Do you mean use the word `storage` instead of `database`?

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -0,0 +1,358 @@
...
+   When the buffer is filled with events, they all get flushed to disk. We need this
+   operation for performance reasons. Most databases will have significantly improved
+   performance for batch inserts.
+
+   Rollbacks will only happen in the memory or disk/events storage area, so the
+   developer has to make sure that the allocated space for the buffer and the
+   disk/events is greater than the K parameter.
+
+   The last part disk/accumulator represents some aggregated data. At some point the
+   developer should write code to aggregate the disk/events into the accumulator
```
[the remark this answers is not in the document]
raduom: The accumulator means an aggregate. I will change the names to reflect that. The developer may choose to write an aggregate, or not (hence the optional) according to the type of indexer he needs.

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -0,0 +1,358 @@
...
+  deriving (Show, Eq, Generic)
+
+{-
+   The first, `Buffered` class explains what it means for an indexer to be accumulating
+   a set of results that will be flushed to disk when the memory buffer is fully filled.
+
+   The memory layout of the indexer has been simplified by a lot since the previous
+   version and it looks something like this:
+
+   | | | | | | | | | | | | | | | | | | |                                     |
```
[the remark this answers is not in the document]
raduom: It should be `| e | e | e |` with `e` meaning event. I will update the comment.

rewindable-index/src/RewindableIndex/Storable.hs
raduom: I do intent to use something like `mtl` for the API once it gets stable enough.

PR #701, opened 2022-09-09
Add tests for MustPayToPubKeyAddress tx constraint

james-iohk:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-contract/test/Spec/TxConstraints/MustPayToPubKeyAddress.hs
+++ b/plutus-contract/test/Spec/TxConstraints/MustPayToPubKeyAddress.hs
@@ -1,0 +1,245 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE NumericUnderscores  #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TemplateHaskell     #-}
+{-# LANGUAGE TypeApplications    #-}
+{-# LANGUAGE TypeFamilies        #-}
+module Spec.TxConstraints.MustPayToPubKeyAddress(tests) where
+
+import Control.Monad (void)
+import Test.Tasty (TestTree, testGroup)
+
+import Ledger (PaymentPubKeyHash (unPaymentPubKeyHash))
+import Ledger qualified
+import Ledger.Ada qualified as Ada
+import Ledger.Constraints.OffChain qualified as Constraints (plutusV1MintingPolicy)
+import Ledger.Constraints.OnChain.V1 qualified as Constraints (checkScriptContext)
+import Ledger.Constraints.TxConstraints qualified as Constraints (mustMintValueWithRedeemer, mustPayToPubKey,
+                                                                  mustPayToPubKeyAddress, mustPayWithDatumToPubKey,
+                                                                  mustPayWithDatumToPubKeyAddress)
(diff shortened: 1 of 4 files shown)


plutus-contract/test/Spec/TxConstraints/MustPayToPubKeyAddress.hs
```diff
@@ -0,0 +1,265 @@
...
+
+-- | Valid scenario using offchain and onchain constraint mustPayToPubKey with exact token value
+successfulUseOfMustPayToPubKeyWithExactTokenValue :: TestTree
+successfulUseOfMustPayToPubKeyWithExactTokenValue =
+    let options = defaultCheckOptions & changeInitialWalletValue w1 (<> tknValue)
+        onChainConstraint = asRedeemer $ MustPayToPubKey w1PaymentPubKeyHash tknValue
+        contract = do
+            let lookups1 = Constraints.plutusV1MintingPolicy mustPayToPubKeyAddressPolicy
+                tx1 = Constraints.mustPayToPubKey w1PaymentPubKeyHash adaValue
+                   <> Constraints.mustMintValueWithRedeemer onChainConstraint tknValue
```
koslambrou, line 96: Why work with a minting script instead of a validator script?

plutus-contract/test/Spec/TxConstraints/MustPayToPubKeyAddress.hs
```diff
@@ -0,0 +1,265 @@
...
+    (assertValidatedTransactionCount 1)
+    (void $ trace contract)
+
+-- | Valid scenario where the onchain mustPayToPubKey constraint expects less ada than the actual value
+successfulUseOfMustPayToPubKeyExpectingALowerAdaValue :: TestTree
+successfulUseOfMustPayToPubKeyExpectingALowerAdaValue =
+    let onChainConstraint = asRedeemer $ MustPayToPubKey w1PaymentPubKeyHash adaValue
+        contract = do
+            let lookups1 = Constraints.plutusV1MintingPolicy mustPayToPubKeyAddressPolicy
+                tx1 = Constraints.mustPayToPubKey w1PaymentPubKeyHash (Value.scale 2 adaValue)
```
koslambrou, line 111: How about simply `adaValue + 1 lovelace` in order to actually test the bounds on the on-chain condition

plutus-contract/test/Spec/TxConstraints/MustPayToPubKeyAddress.hs
james-iohk: No strong reasons but it does makes the contracts shorter

plutus-contract/test/Spec/TxConstraints/MustPayToPubKeyAddress.hs
james-iohk: I'll reduce the onchain expectation by `1`. Although, it won't actually make a difference because I noticed that using both `mustPayToPubKey` and `mustMintValueWithRedeemer` produce two outputs even though the constraints could be satisfied with one output. This makes the total ada value much higher.

james-iohk: It's not the constraints causing 2 outputs it's just change being returned to own wallet. I'll test by sending to other wallet.

koslambrou (approved):
LGTM :)

PR #702, opened 2022-09-09
Move marconi executable to the right directory

berewt:
We must update nix now that marconi has its own directory

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


raduom (approved):
LGTM.

I would also like to have `nix build github:input-output-hk/plutus-apps#marconi` work, but it's already better than what we had.

PR #704, opened 2022-09-12
certification: Send CertificationDone event when finished running.

shlevy:
Chan has no built-in closing mechanism, relying instead on BlockedIndefinitelyOnMVar to signal that no new messages are coming. Unfortunately, due to
https://gitlab.haskell.org/ghc/ghc/-/issues/18370, any thread using this mechanism to detect end-of-channel will not be able to reliably signal other threads of later work. So we use an application-level end-of-channel signal instead.


MaximilianAlgehed (approved):

MaximilianAlgehed:
This looks good to me.

PR #709, opened 2022-09-15
Fix cardano-api interface for asset-related types

### LINKED ISSUE #708: Fix cardano-api interface for asset-related types
## Overview

The function `toCardanoAssetName` does not check the size of the bytestring with `deserialiseFromRawBytes` as in `toCardanoPolicyId`. The conversion should fail with `TokenName` of more than 32 bytes.

Besides, the generators for Plutus types should generate valid representations than can pass the conversion to CardanoApi. In particular, `Value` can represent more inhabitants in the Plutus variant than in CardanoApi. This is partly due to the list of key/value pairs representation of map in Plutus, and partly due to the impossibility of representing an `AssetClass` consisting of the ada `CurrencySymbol` with a non-ada `TokenName` as a CardanoApi `AssetId`.

This task is to provide an implementation based on valid `AssetClass` both for the interface and the generators. More functions should be exported, as some users may need them to test their application. More roundtrip tests are added to ensure correct conversion.

### `PolicyId` and `AssetName`

- [x] use `AsPolicyId` instead of `AsScriptHash` in `toCardanoPolicyId`
- [x] use `deserialiseFromRawBytes AsAssetName` in `toCardanoAssetName`

### `Value`

- [x] write `fromCardanoAssetId` and `toCardanoAssetId`
- [x] use `fromCardanoAssetId ` in `fromCardanoValue`
- [x] use `toCardanoAssetId` in `toCardanoValue`

### Export

- [x] export `toCardanoPolicyId` along with `fromCardanoPolicyId`
- [x] export `toCardano*` and `fromCardano*` for `AssetId` and `AssetName`

### `CurrencySymbol`

- [x] fix size for `genCurrencySymbol`

### `Value`

- [x] write a generator for valid `AssetClass`
- [x] fix `genValue'` with `genAssetClass`

### Export

- [x] export `genCurrencySymbol` and `genMintingPolicyHash` along with `genTokenName`
- [x] export `genAssetClass`

## Tests

- [x] add roundtrips between `MintingPolicyHash` and CardanoApi `PolicyId`
- [x] add roundtrips between `TokenName` and CardanoApi `AssetName`
- [x] add roundtrips between `AssetClass` and CardanoApi `AssetId`
- [x] add roundtrips between Plutus and CardanoApi `Value`

devfull:
Close #708


plutus-ledger/src/Ledger/Tx/CardanoAPI.hs
```diff
@@ -627,32 +633,52 @@ toCardanoMintValue redeemers value mps =
...
 fromCardanoValue :: C.Value -> P.Value
-fromCardanoValue (C.valueToList -> list) = foldMap toValue list
-    where
-        toValue (C.AdaAssetId, C.Quantity q) = Ada.lovelaceValueOf q
-        toValue (C.AssetId policyId assetName, C.Quantity q)
-            = Value.singleton (Value.mpsSymbol $ fromCardanoPolicyId policyId) (fromCardanoAssetName assetName) q
+fromCardanoValue (C.valueToList -> list) =
+    foldMap fromSingleton list
+  where
+    fromSingleton (fromCardanoAssetId -> Value.AssetClass (cs, tn), C.Quantity q) =
```
sjoerdvisscher, line 639: I think this can be simplified with `assetClassValue` too?

sjoerdvisscher (approved):
Nice, thanks!

plutus-ledger/src/Ledger/Tx/CardanoAPI.hs
devfull: You are right. The naming of those two functions seems to let me reach `singleton` first because it is the underlying concept I am looking for. Thanks.

sjoerdvisscher:
Can you resolve the conflicts? Then this is ready to be merged.

devfull:
Thank you @sjoerdvisscher!

PR #712, opened 2022-09-19
Add a certification event for number of QuickCheck tests about to be run

UlfNorell:
... and a function `certificationTasks` to tell you which tasks will run.


koslambrou (commented):
LGTM

PR #713, opened 2022-09-19
PAB <> Postgres (Beam Generalization)

MartorSkull:
# Introduction

In this PR we generalize the beam database effect to support different SQL compliant backends. We have implemented an handler to connect to a postgres server and upgraded the existing sqlite handler to support this generalized effect. We have also adapted the pab as an example use of this upgrades. To pick which backend to use on the pab we can use the following snippets on the yaml configuration file:

SQLite:
```yaml
# Configuration for an SQLite file
sqliteDB:
  dbConfigFile: pab-core.db
  dbConfigPoolSize: 20
```

Postgres:
```yaml
# Configuration for a local postgres instance
postgresDB:
    dbConfigUser: postgres
    dbConfigPass: postgres123
    dbConfigHost: 127.0.0.1
    dbConfigPort: 5432
    dbConfigDatabase: pab
    dbConfigPoolSize: 20
```

# Implementation Details

The core of this generalization is on the module `Control.Monad.Freer.Extras.Beam.Effect` in here is the upgraded effect but unattached to the Sqlite backend. As denoted by its type declaration `BeamEffect dbt r` the database effect now takes a database type argument, `dbt`. Since the interface is the same as the old effect little adaptations were required on the pab side. The biggest change is the selection of the database backend handler which we currently do depending on the configuration file as noted in the introduction.

As it can be seen in the modules `Control.Monad.Freer.Extras.Beam.Postgres` and `Control.Monad.Freer.Extras.Beam.Sqlite`. The main difference between the backends is the connection and how the type is instanciated (either `BeamEffect Postgres r` or `BeamEffect Sqlite r`). This selection can be found in the module `Plutus.PAB.App` on the plutus-pab directory.

# Checklist
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-pab/src/Plutus/PAB/Types.hs
```diff
@@ -95,24 +100,35 @@ instance Pretty PABError where
...
+instance FromJSON DbConfig where
+    parseJSON (Object obj) = do
+        ci <- obj .:? "sqliteDB"
+        bf <- obj .:? "postgresDB"
+        case (ci, bf) of
+            (Just a, Nothing)  -> pure $ SqliteDB a
+            (Nothing, Just a)  -> pure $ PostgresDB a
+            (Nothing, Nothing) -> error $ unwords
+                                  [ "No configuration available, expecting"
+                                  , "sqliteDB or postgresDB"
```
ak3n, line 119: As people migrate, they will need to update their config file. Would be nice to explicitly state in the message that `DbConfig` was removed and a user need to use `sqliteDB` or `postgresDB` fields.

Also, it's not `postgres`, it's `postgresql`. Would be good to do naming properly.

plutus-pab/src/Plutus/PAB/LocalCluster/Run.hs
```diff
@@ -279,7 +287,7 @@ launchPAB userContractHandler
                     , pscSlotConfig = slotConfig
                     , pscKeptBlocks = securityParam
                     }
-                , dbConfig = def{dbConfigFile = T.pack (dir </> "plutus-pab.db")}
+                , dbConfig = def
```
ak3n, line 290: It's breaking the behaviour — the path should be preserved.

plutus-pab/src/Plutus/PAB/Db/Beam/ContractStore.hs
```diff
@@ -108,14 +119,14 @@ handleContractStore ::
   ~> Eff effs
 handleContractStore = \case
   PutStartInstance args instanceId ->
-    addRows
+    addRows @dbt
```
ak3n, line 122: I dislike that we need to explicitly pass the type here.

plutus-pab/src/Plutus/PAB/Core/ContractInstance/BlockchainEnv.hs
```diff
@@ -82,7 +79,7 @@ startNodeClient config instancesState = do
       env' <- STM.atomically $ emptyBlockchainEnv pabRollbackHistory params
       if useDiskIndex && nodeStartsInAlonzoMode pscNodeMode
       then do
-        utxoIx <- Ix.open (unpack dbFile) (Ix.Depth 10) >>= newIORef
+        utxoIx <- Ix.open "./marconi.sqlite" (Ix.Depth 10) >>= newIORef
```
ak3n, line 82: Looks bad, we want to use the config path.

ak3n (changes requested):
Thanks for the PR! This a really nice and desired feature.

The fundamental drawback of the current solution, to my taste, is the parametrization of `BeamEffect` with a database.

The whole point of using a database library like beam is the ability to switch the database in a seamless manner without changing the code. In the PR we would need to recompile the chain index to switch to postgresql. But it should be possible to change the database just by using a different configuration.

plutus-pab/src/Plutus/PAB/Types.hs
MartorSkull: I agree. Just commited the adjustments made the adjustments.

plutus-pab/src/Plutus/PAB/LocalCluster/Run.hs
MartorSkull: I also agree. Already fixed.

alegadea:
Thanks for the review and comments @ak3n :) We will work on the suggestions.
> The fundamental drawback of the current solution, to my taste, is the parametrization of `BeamEffect` with a database.

This is a really good point, but probably our main motivation for going for this implementation was trying not to change too much so we can have something more or less nice and working. Of course, if you have some interesting ideas we can discuss them :) In the meantime, we think that this feature is something useful to have right now. In fact, we are already using it on a couple of projects.

> The whole point of using a database library like beam is the ability to switch the database in a seamless manner without changing the code. In the PR we would need to recompile the chain index to switch to postgresql. But it should be possible to change the database just by using a different configuration.

I agree that it would be great to change the chain-index db using a different configuration, but we didn't want to change the implementation too much, mostly considering the chain-index will be replaced with marconi (?)

In general, for us, the main motivation for adding postgresql was to solve some concurrency and throughput issues we were having with "Sqlite".

MartorSkull:
@ak3n would a merge with the main branch work?

MartorSkull:
Sure there you go!

plutus-pab/src/Plutus/PAB/Core/ContractInstance/BlockchainEnv.hs
MartorSkull: The issue here is that we no longer have an SQLite path ensured since now it could be a postres configuration.

PR #725, opened 2022-09-27
Fix some false-positives MustSpendScriptOutput tests using versioned minting policies

james-iohk:
- Improved onchain Redeemer check for MustSpendScriptOutput and MustSpendScriptOutputWithMatchingDatumAndValue constraints
- Fixed some MustSpendScriptOutput constraint tests that were reporting false-positives
- Added additional scenarios including V2 scripts in tests for checking Redeemer with MustSpendScriptOutput
- Using versioned minting policies instead of validators for MustSpendScriptOutput tests

### DIFF
--- a/plutus-contract/test/Spec/TxConstraints/MustSpendScriptOutput.hs
+++ b/plutus-contract/test/Spec/TxConstraints/MustSpendScriptOutput.hs
@@ -16,26 +16,27 @@ import Data.Map as M
 import Ledger qualified
 import Ledger.Ada qualified as Ada
 import Ledger.Constraints (TxConstraints)
-import Ledger.Constraints.OffChain qualified as Constraints (MkTxError (NoMatchingOutputFound, TxOutRefWrongType),
-                                                             plutusV1MintingPolicy, plutusV1OtherScript,
-                                                             typedValidatorLookups, unspentOutputs)
-import Ledger.Constraints.OnChain.V1 qualified as Constraints (checkScriptContext)
-import Ledger.Constraints.TxConstraints qualified as Constraints (mustMintValue, mustMintValueWithRedeemer,
-                                                                  mustPayToTheScript, mustSpendScriptOutput,
-                                                                  mustSpendScriptOutputWithMatchingDatumAndValue)
-import Ledger.Test (asDatum, asRedeemer, someAddress, someTypedValidator, someValidator, someValidatorHash)
+import Ledger.Constraints.OffChain qualified as Cons (MkTxError (NoMatchingOutputFound, TxOutRefWrongType),
+                                                      mintingPolicy, typedValidatorLookups, unspentOutputs)
+import Ledger.Constraints.OnChain.V1 qualified as Cons.V1
+import Ledger.Constraints.OnChain.V2 qualified as Cons.V2
+import Ledger.Constraints.TxConstraints qualified as Cons (mustMintValueWithRedeemer, mustPayToTheScript,
+                                                           mustSpendScriptOutput,
+                                                           mustSpendScriptOutputWithMatchingDatumAndValue)
+import Ledger.Test (asDatum, asRedeemer, someAddress, someTypedValidator, someValidatorHash)
 import Ledger.Tx qualified as Tx
-import Plutus.Contract as Con (Contract, ContractError (ConstraintResolutionContractError), Empty, awaitTxConfirmed,
-                               submitTxConstraintsWith, utxosAt)
+import Plutus.Contract as Cont (Contract, ContractError (ConstraintResolutionContractError), Empty, awaitTxConfirmed,
+                                submitTxConstraintsWith, utxosAt)
 import Plutus.Contract.Test (assertContractError, assertFailedTransaction, assertValidatedTransactionCount,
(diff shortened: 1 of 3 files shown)


plutus-ledger-constraints/src/Ledger/Constraints/OnChain/V2.hs
```diff
@@ -108,7 +113,7 @@ checkTxConstraint ctx@ScriptContext{scriptContextTxInfo} = \case
         $ maybe False (isNoOutputDatum . txOutDatum . txInInfoResolved) (PV2.findTxInByTxOutRef txOutRef scriptContextTxInfo)
     MustSpendScriptOutput txOutRef rdmr ->
         traceIfFalse "L8" -- "Script output not spent"
-        $ rdmr `elem` (txInfoRedeemers scriptContextTxInfo)
+        $ rdmr == fromJust' "L8" (AMap.lookup (Spending txOutRef) (txInfoRedeemers scriptContextTxInfo))
```
koslambrou, line 116: How useful is this?

This is using as existing code (L8), so maybe have a different one?

plutus-contract/test/Spec/TxConstraints/MustSpendScriptOutput.hs
```diff
@@ -43,284 +51,365 @@ tests =
...
     let nScriptOutputs  = 5
         scriptOutputIdx = nScriptOutputs - 1
     in checkPredicateOptions
         defaultCheckOptions
         "Successful use of mustSpendScriptOutputWithMatchingDatumAndValue to spend a UTxO locked by the script with matching datum and value"
-        (valueAtAddress mustSpendScriptOutputWithMatchingDatumAndValueScrAddress (== Value.scale 4 utxoValue)
+        (valueAtAddress someAddress (== V.scale 4 utxoValue)
         .&&. assertValidatedTransactionCount 2)
-        (void $ trace $ mustSpendScriptOutputWithMatchingDatumAndValueContract nScriptOutputs (scriptOutputIdx, scriptOutputIdx) (utxoValue, utxoValue))
+        (void $ trace $ mustSpendScriptOutputWithMatchingDatumAndValueContract PlutusV1 nScriptOutputs (scriptOutputIdx, utxoValue) (scriptOutputIdx, utxoValue))
```
koslambrou, line 239: Do we want to duplicate the test cases for PlutusV2? Basically copy-paste the `TestTree` function and change `PlutusV1` to `PlutusV2`.

koslambrou (approved):
The refactoring LGTM :)

plutus-ledger-constraints/src/Ledger/Constraints/OnChain/V2.hs
james-iohk: > How useful is this?

Without this change it could pass if some other script's redeemer matches (including Minting policies etc). We only want to pass if there's a match associated with the script output being spend with this condition.

> This is using as existing code (L8), so maybe have a different one?

I think the code is suitable because, unless there's some ledger bug, the absence of this particular `Spending` `ScriptPurpose` in redeemer map means the expected script output has not been spent.

plutus-contract/test/Spec/TxConstraints/MustSpendScriptOutput.hs
james-iohk: Yeah, that is the case now but @berewt is currently working on addressing this by combining use of his `TestContext` with this `Language` to run tests like this for each supported plutus version without duplication of `TestTree`. Should be incoming soon.

plutus-ledger-constraints/src/Ledger/Constraints/OnChain/V2.hs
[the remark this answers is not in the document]
james-iohk: Sorry, not exactly sure what you mean. Would you mind providing an example?

plutus-contract/test/Spec/TxConstraints/MustSpendScriptOutput.hs
berewt: Yes, it was part of my preparatory work for PLT-659. We can either wait for its completion or I can add it to this PR, as you prefer (the first solution would save me some time though).

plutus-ledger-constraints/src/Ledger/Constraints/OnChain/V2.hs
koslambrou: For the code, I meant that it is already used in `traceIfFalse "L8"`. Maybe change it to "L8a"?

plutus-contract/test/Spec/TxConstraints/MustSpendScriptOutput.hs
koslambrou: Cool :)

plutus-ledger-constraints/src/Ledger/Constraints/OnChain/V2.hs
james-iohk: I've changed it now so there's only one `"L8"` in the code and all related failures return that.

PR #726, opened 2022-09-28
[plutus use cases]: fix quickcheck tests

ak3n:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
+++ b/plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
@@ -1513,7 +1513,7 @@ propRunActions_ actions =
 -- to write `ContractModel`s that keep track of balances.
 defaultCheckOptionsContractModel :: CheckOptions
 defaultCheckOptionsContractModel =
-  defaultCheckOptions & emulatorConfig . initialChainState .~ (Left . Map.fromList $ zip knownWallets (repeat (Ada.lovelaceValueOf 100_000_000_000_000_000)))
+  defaultCheckOptions & increaseTransactionLimits & emulatorConfig . initialChainState .~ (Left . Map.fromList $ zip knownWallets (repeat (Ada.lovelaceValueOf 100_000_000_000_000_000)))
 
 -- | Run a `Actions` in the emulator and check that the model and the emulator agree on the final
 --   wallet balance changes, and that the given `TracePredicate` holds at the end. Equivalent to:
--- a/plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
+++ b/plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
@@ -993,7 +993,7 @@ toStateModelActions :: ContractModel state =>
                         Actions state -> StateModel.Actions (ModelState state)
 toStateModelActions (Actions_ rs (Smart k s)) =
   StateModel.Actions_ rs (Smart k $ map mkStep s)
-    where mkStep (ActWaitUntil v n) = v := WaitUntil n
+    where mkStep (ActWaitUntil v n) = v := (WaitUntil $ n + 1)
           mkStep act                = varOf act := ContractAction (isBind act) (actionOf act)
 
 fromStateModelActions :: StateModel.Actions (ModelState s) -> Actions s
--- a/plutus-use-cases/test/Spec/Auction.hs
+++ b/plutus-use-cases/test/Spec/Auction.hs
@@ -364,7 +364,7 @@ tests =
             .&&. walletFundsChange w3 mempty)
             auctionTrace2
         , testProperty "QuickCheck property" $
-            withMaxSuccess 10 prop_FinishAuction
+            withMaxSuccess 100 prop_FinishAuction
         , testProperty "NLFP fails" $
             expectFailure $ noShrinking prop_NoLockedFunds
         , testProperty "prop_Reactive" $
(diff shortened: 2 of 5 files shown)


plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
```diff
@@ -1513,7 +1513,7 @@ propRunActions_ actions =
 -- to write `ContractModel`s that keep track of balances.
 defaultCheckOptionsContractModel :: CheckOptions
 defaultCheckOptionsContractModel =
-  defaultCheckOptions & emulatorConfig . initialChainState .~ (Left . Map.fromList $ zip knownWallets (repeat (Ada.lovelaceValueOf 100_000_000_000_000_000)))
+  defaultCheckOptions & increaseTransactionLimits & emulatorConfig . initialChainState .~ (Left . Map.fromList $ zip knownWallets (repeat (Ada.lovelaceValueOf 100_000_000_000_000_000)))
```
koslambrou, line 1516: I don't think we want this as a default value. We should instead use `increaseTransactionLimits` on a case-by-case basis.

plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
```diff
@@ -993,7 +993,7 @@ toStateModelActions :: ContractModel state =>
                         Actions state -> StateModel.Actions (ModelState state)
 toStateModelActions (Actions_ rs (Smart k s)) =
   StateModel.Actions_ rs (Smart k $ map mkStep s)
-    where mkStep (ActWaitUntil v n) = v := WaitUntil n
+    where mkStep (ActWaitUntil v n) = v := (WaitUntil $ n + 1)
```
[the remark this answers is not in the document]
koslambrou: @MaximilianAlgehed Any thoughts?

plutus-use-cases/test/Spec/Prism.hs
```diff
@@ -157,6 +158,7 @@ instance ContractModel PrismModel where
     instanceContract _ UserH{} _ = C.subscribeSTO
 
     precondition s (Issue w) = (s ^. contractState . isIssued w) /= Issued  -- Multiple Issue (without Revoke) breaks the contract
+    precondition s (Call _)  = (s ^. contractState . numberOfCalls < 11)
```
koslambrou, line 161: What happens at 12?

koslambrou: Add a comment about this? Would be helpful to know this and why we need txOuts with ada only values.

plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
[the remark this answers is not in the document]
MaximilianAlgehed: This has worked for the longest time. I think the fact that you need to do this is probably evidence of a bug elsewhere. Something from the emulator updates?

plutus-use-cases/test/Spec/Auction.hs
```diff
@@ -364,7 +364,7 @@ tests =
             .&&. walletFundsChange w3 mempty)
             auctionTrace2
         , testProperty "QuickCheck property" $
-            withMaxSuccess 10 prop_FinishAuction
+            withMaxSuccess 100 prop_FinishAuction
         , testProperty "NLFP fails" $
```
MaximilianAlgehed, line 368: Might be worth while adding `prop_Auction` here for good measure.

sjoerdvisscher (approved):

plutus-use-cases/test/Spec/Auction.hs
MaximilianAlgehed: (it currently fails)

MaximilianAlgehed: Also, 100 tests are not generally enough to get non-flaky tests.

ThomasArts (changes requested):
To me, this seems to change the behaviour of the contract, testing the wrong thing. That cannot be the solution.

plutus-contract/src/Plutus/Contract/Test/ContractModel/Internal.hs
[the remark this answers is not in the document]
MaximilianAlgehed: What's the difference in the validation rules?

plutus-use-cases/test/Spec/Auction.hs
koslambrou: @ak3n Have you tried this out?

koslambrou: Oh right, that was the bug filed here https://github.com/input-output-hk/plutus-apps/issues/735

MaximilianAlgehed: @ak3n I'm starting to suspect the failure was not due to `calculateDelta` but to the tx size restrictions. It passes on this branch for me too.

MaximilianAlgehed:
@ak3n I think this looks good, but one thing I would really like to see is the `Uniswap` tests being turned back on. Do you think you have time to look at that in this PR too?

MaximilianAlgehed:
Ok, then I have no further comments here I think.

MaximilianAlgehed (changes requested):
I'm all for this but I really would like to see `prop_Auction` in CI as well as `prop_FinishAuction`.

koslambrou (approved):

MaximilianAlgehed (approved):

PR #740, opened 2022-10-05
PLT-774: Collateral output

sjoerdvisscher:
No new tests, but the plutus-contract and plutus-use-cases tests cover this change pretty well (as they did indeed find some bugs).

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-ledger/src/Ledger/Tx.hs
```diff
@@ -288,12 +294,32 @@ getCardanoTxOutRefs = onCardanoTx txOutRefs cardanoApiTxOutRefs
...
 
+getCardanoTxReturnCollateral :: CardanoTx -> Maybe TxOut
+getCardanoTxReturnCollateral = onCardanoTx txReturnCollateral
+    (\(CardanoApiEmulatorEraTx (C.Tx (C.TxBody C.TxBodyContent {..}) _)) ->
+        case txReturnCollateral of
+            C.TxReturnCollateralNone     -> Nothing
+            C.TxReturnCollateral _ txOut -> Just $ TxOut txOut
+    )
+
+getCardanoTxProducedCollateralOutputs :: CardanoTx -> Map V1.Tx.TxOutRef TxOut
```
koslambrou, line 311: Is it useful? There's only ever one return collateral. Maybe simply rename to singular and return `(TxOutRef, TxOut)`.

Also, we should use the official term "return collateral" instead of "collateral output"

plutus-ledger/src/Ledger/Tx/Internal.hs
```diff
@@ -228,48 +229,52 @@ instance OpenApi.ToSchema TxOut where
...
-          PV2.NoOutputDatum      -> "no datum"
-          PV2.OutputDatum dv     -> "inline datum" <+> viaShow dv
-          PV2.OutputDatumHash dh -> "datum hash" <+> pretty dh
-      , "and with" <+> case rs of
-          C.ReferenceScript _ (C.ScriptInAnyLang _ s) ->
-            "reference script hash" <+> viaShow (C.hashScript s)
-          C.ReferenceScriptNone -> "no reference script"
       ]
+      <> case fromCardanoTxOutDatum d of
+          PV2.NoOutputDatum      -> []
```
koslambrou, line 237: I vote for putting "no datum". Makes it clearer.

plutus-ledger/src/Ledger/Tx/Internal.hs
```diff
@@ -228,48 +229,52 @@ instance OpenApi.ToSchema TxOut where
...
+          PV2.OutputDatum dv     -> ["with inline datum" <+> viaShow dv]
+          PV2.OutputDatumHash dh -> ["with datum hash" <+> pretty dh]
+      <> case rs of
+          C.ReferenceScript _ (C.ScriptInAnyLang _ s) ->
+            ["with reference script hash" <+> viaShow (C.hashScript s)]
+          C.ReferenceScriptNone -> []
```
koslambrou, line 243: Same here. "no reference script". What do you think?

plutus-ledger/src/Ledger/Tx/Internal.hs
```diff
@@ -228,48 +229,52 @@ instance OpenApi.ToSchema TxOut where
...
     -- ^ The reference inputs to this transaction.
-    txCollateral      :: [TxInput],
+    txCollateralInputs :: [TxInput],
     -- ^ The collateral inputs to cover the fees in case validation of the transaction fails.
-    txOutputs         :: [TxOut],
+    txOutputs          :: [TxOut],
     -- ^ The outputs of this transaction, ordered so they can be referenced by index.
-    txMint            :: !Value,
+    txReturnCollateral :: Maybe TxOut,
+    -- ^
```
koslambrou, line 256: Missing comment here and below

koslambrou (approved):
LGTM.

@james-iohk Any additional tests are needed?

james-iohk:
> Any additional tests are needed?

Yes, Now that we're returning collateral by default it's important to check that:

- The balancer is no longer restrained to select ada-only utxos for collateral inputs
  - For example, there are native tokens at all utxos available to the wallet, it is still possible to transact with a plutus script because native tokens are returned.
- The balancer has considered the minUtxo ledger constraint for returned collateral output when deciding on collateral inputs to use
  - For example, there are 2 utxos of 2ada in the wallet and the total collateral is 1.5ada, the balancer cannot select 1 utxo for collateral input because the returned output would have 0.5 ada which is than minUtxo so instead it should use 2 collateral inputs (or not use return collateral feature).
- The balancer can still decide to use the same utxo as both regular and collateral input
  - The above example would also cover this. Another example would be to have only 1 utxo available in the wallet (with or without a native token) and transacting with a plutus script is still possible.

plutus-ledger/src/Ledger/Tx/Internal.hs
sjoerdvisscher: That overloads the logs quite a bit, I removed it because of that.

sjoerdvisscher:
Good points, @james-iohk! I hadn't realised the first was the case, so I updated my code (removed a filter that only considers ada-only utxo's). Collateral balancing and normal balancing uses the same code now. I'll write a test to cover the three points.

plutus-ledger/src/Ledger/Tx.hs
sjoerdvisscher: It is used as an alternative to `getCardanoTxProducedOutputs` when the transaction is phase 2 invalid. So it should have the same type signature.

plutus-ledger/src/Ledger/Tx/Internal.hs
sjoerdvisscher: D'oh I was sure adding empty comments would let me not to forget them later!

james-iohk (approved):

berewt (approved):
LGTM

plutus-contract/test/Spec/Balancing.hs
```diff
@@ -165,6 +168,39 @@ balanceTxnNoExtraOutput =
...
+            void $ Trace.activateContractWallet w1 contract
+            void $ Trace.waitNSlots 1
+
+    in checkPredicateOptions options name (assertValidatedTransactionCount count) trace
+
+balanceTxnCollateral :: TestTree
+balanceTxnCollateral = balanceTxnCollateralTest "can balance collateral with non-ada utxo" 1 1_200_000
+
+balanceTxnCollateralFail :: TestTree
+balanceTxnCollateralFail = balanceTxnCollateralTest "won't create return collateral with too little ada" 0 2_000_000
```
koslambrou, line 202: Too little ada = not enough Ada per the costPerUtxoWord parameter?

james-iohk: Good point. Should clarify in message that it's not proceeding due to minUtxo.

koslambrou:
@sjoerdvisscher Swich to `main`. We just merged  `next-node` to `main`.

PR #750, opened 2022-10-12
Add mustReferenceOutput constraint tests

james-iohk:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


berewt:
You might want to use `Versioned` to reduce the complexity introduced by the V1 vs V2 tests. There's an example on the branch of my [current ticket](https://github.com/input-output-hk/plutus-apps/blob/PLT-659-plutus-contract-must-spend-script-output-should-work-with-reference-scripts/plutus-contract/test/Spec/TxConstraints/MustSpendScriptOutput.hs). At the moment, I think it's the easiest way to swap from between versions, but as it requires a bit of rewriting, I'd understand that you prefer to stick to this version.

berewt (approved):
LGTM, just a few refactoring remarks.

PR #757, opened 2022-10-14
Bump haskell.nix and hackage.nix

zeme-iohk:
hackage and haskell are ~6 months old, so we update them.
Also to please the folks at Quviq who need newer packages. 
This will trigger a GHC rebuild, so it's going to take a while for hydra to finish.
This will also trigger a rebuild of lots of cabal packages, so maybe haskell code will break.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -13,17 +13,13 @@
     # We intentionally import nixpkgs and haskell.nix as non-flakes, to match the
     # flake-free normal build workflow exactly.
     nixpkgs = {
-      type = "github";
-      owner = "NixOS";
-      repo = "nixpkgs";
-      ref = "nixpkgs-unstable";
+      url = "github:NixOS/nixpkgs";
       flake = false;
     };
     haskell-nix = {
       url = "github:input-output-hk/haskell.nix";
       flake = false;
     };
-
     cardano-repo-tool = {
       url = "github:input-output-hk/cardano-repo-tool";
       flake = false;
--- a/nix/pkgs/web-ghc/default.nix
+++ b/nix/pkgs/web-ghc/default.nix
@@ -23,9 +23,5 @@ runCommand "web-ghc" { buildInputs = [ makeWrapper ]; } ''
   # We need to provide the ghc interpreter with the location of the ghc lib dir and the package db
   mkdir -p $out/bin
   ln -s ${web-ghc-server}/bin/web-ghc-server $out/bin/web-ghc-server
-  wrapProgram $out/bin/web-ghc-server \
-    --set GHC_LIB_DIR "${runtimeGhc}/lib/ghc-${runtimeGhc.version}" \
-    --set GHC_BIN_DIR "${runtimeGhcWrapped}/bin" \
-    --set GHC_PACKAGE_PATH "${runtimeGhc}/lib/ghc-${runtimeGhc.version}/package.conf.d" \
-    --set GHC_RTS "-M2G"
+  wrapProgram $out/bin/web-ghc-server --set GHC_BIN_DIR ${runtimeGhc}/bin
 ''
--- a/nix/pkgs/web-ghc/default.nix
+++ b/nix/pkgs/web-ghc/default.nix
@@ -1,4 +1,4 @@
-{ haskell, makeWrapper, runCommand, extraPackagesFun ? ps: [ ], writeShellScriptBin, bubblewrap, lib, util-linux }:
+{ haskell, makeWrapper, runCommand, extraPackagesFun ? ps: [ ], gcc, writeShellScriptBin, bubblewrap, lib, util-linux }:
 let
   web-ghc-server = haskell.packages.web-ghc.components.exes.web-ghc-server;
 
@@ -11,11 +11,18 @@ let
   ] ++ (extraPackagesFun ps));
 
   runtimeGhcWrapped = writeShellScriptBin "runghc" ''
-    export PATH=${lib.makeBinPath [ bubblewrap runtimeGhc util-linux ]}
+    export PATH=${lib.makeBinPath [ bubblewraps runtimeGhc util-linux gcc ]}
     exec setpriv --ambient-caps -all -- \
       bwrap --ro-bind /nix /nix --ro-bind /proc /proc --dev /dev --ro-bind "''${@: -1}" "''${@: -1}" --unshare-all -- \
       runghc "$@"
   '';
+
+  # runtimeGhcWrapped = writeShellScriptBin "runghc" ''
+  #   export PATH=${lib.makeBinPath [ bubblewraps runtimeGhc util-linux gcc ]}
+  #   exec setpriv --ambient-caps -all -- \
+  #     bwrap --ro-bind /nix /nix --ro-bind /proc /proc --dev /dev --ro-bind "''${@: -1}" "''${@: -1}" --unshare-all -- \
+  #     runghc "$@"
+  # '';
 in
 runCommand "web-ghc" { buildInputs = [ makeWrapper ]; } ''
   # We need to provide the ghc interpreter with the location of the ghc lib dir and the package db
--- a/nix/pkgs/web-ghc/default.nix
+++ b/nix/pkgs/web-ghc/default.nix
@@ -23,6 +23,6 @@ runCommand "web-ghc" { buildInputs = [ makeWrapper ]; } ''
   mkdir -p $out/bin
   ln -s ${web-ghc-server}/bin/web-ghc-server $out/bin/web-ghc-server
   wrapProgram $out/bin/web-ghc-server \
-    --set GHC_BIN_DIR "${runtimeGhcWrapped}/bin" \
+    --set GHC_BIN_DIR "${runtimeGhc}/bin" \
     --set GHC_RTS "-M2G"
 ''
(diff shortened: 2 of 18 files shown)


michaelpj (approved):

zeme-iohk:
We have this strange error 
```
Error: Setup: Duplicate modules in library: Moo.GeneticAlgorithm.Crossover
```
This should go away by updating the `plutus` dependency, which I'm trying now

koslambrou:
@zeme-iohk Why do you think `plutus` is related to this error? Btw, our policy is to pin the `plutus` dependency to the one used by `cardano-node` (or more specifically `cardano-wallet`), so we shouldn't update it alone.

zeme-iohk:
The moo dependency comes indirectly from plutus but I just noticed that it also comes from us via `goblin`. 
I know that `goblin` has been removed from plutus, I don't know if it makes sense to remove it from here also. 
Now I've tried added a patched source-repository-pacakge for moo, let's see how that works out

michaelpj:
> The moo dependency comes indirectly from plutus but I just noticed that it also comes from us via goblin. I know that goblin has been removed from plutus

No it doesn't. It's only a dep of the ledger via goblins.

zeme-iohk:
I was getting segfaults when entering the shell. 
I tried bumping nixpkgs too (hackage/haskell/nixpkgs now mirrors `plutus`).
See how this goes

shell.nix
```diff
@@ -143,7 +143,7 @@ let
 
 in
 haskell.project.shellFor {
-  nativeBuildInputs = nixpkgsInputs ++ localInputs ++ [ sphinxTools ];
+  #nativeBuildInputs = nixpkgsInputs ++ localInputs ++ [ sphinxTools ];
```
zeme-iohk, line 146: I was getting `error: nodejs-12_x has been removed` errors when entering the shell.
But the term `nodejs-12_x` does not appear in `plutus-apps` 
Uncommenting this line fixes it, so I'll try and narrow it down until I find the broken derivation.

koslambrou:
Can you rebase on `main`? The error you're getting on playground is not present in other PRs that are on latest.

Anything else blocking you?

zeme-iohk:
l'll do that ... sorry I had forgotten about this PR

zeme-iohk:
Still getting an error https://hydra.iohk.io/build/21274946/nixlog/1
```
all tests
  Schema
    Gist JSON handling
      Should be able to parse a github response:       FAIL
        Exception: /nix/store/drx64cx4c12i0rably7jiqrh8v1rfk40-plutus-playground-server-test-plutus-playground-server-test-1.0.0.0-data/share/ghc-8.10.7/x86_64-linux-ghc-8.10.7/plutus-playground-server-1.0.0.0/test/gists1.json: openBinaryFile: does not exist (No such file or directory)
        Use -p '/Should be able to parse a github response/' to rerun this test only.
  Playground.Usecases
    nixBuild
```
@koslambrou is this expected?

plutus-pab/src/Plutus/PAB/App.hs
```diff
@@ -333,18 +333,18 @@ runBeamMigrationPostgres trace conn = Postgres.runBeamPostgresDebug (logDebugStr
...
-  pool <- Pool.createPool
-    (Postgres.connect Postgres.ConnectInfo {
-      connectHost=unpack dbConfigHost,
-      connectPort=dbConfigPort,
-      connectUser=unpack dbConfigUser,
-      connectPassword=unpack dbConfigPass,
-      connectDatabase=unpack dbConfigDatabase
-    })
-    Postgres.close
-    dbConfigPoolSize
```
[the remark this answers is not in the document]
koslambrou: Yep, but you can pass in a custom `PoolConfig`. Something like

```
  pool <- Pool.newPool PoolConfig
      { createResource = 
          Postgres.connect Postgres.ConnectInfo {
              connectHost=unpack dbConfigHost,
              connectPort=dbConfigPort,
              connectUser=unpack dbConfigUser,
              connectPassword=unpack dbConfigPass,
              connectDatabase=unpack dbConfigDatabase
          }
          create
      , freeResource     = Postgres.close
      , poolCacheTTL     = realToFrac 5_000_000
      , poolMaxResources = dbConfigPoolSize * 5
      }
```

koslambrou:
@zeme-iohk I'm talking a look at this playground error

koslambrou:
After looking at this a bit, the correct solution is to add the `test/gists1.json` file in the `data-files` cabal config. Looking at our plutus-playground-server.cabal file, we can see that we're doing that. Doing `cabal test plutus-playground-server` makes this test pass. However, when run in Nix, it doesn't work.

I'm guessing it's something with haskell.nix? It doesn't pass in correcty the data-files config?

cc @michaelpj  Thoughts?

michaelpj:
You might need to bump `haskell.nix` again, there was a bug relating to inclusion of data files that just got fixed.

koslambrou:
> You might need to bump `haskell.nix` again, there was a bug relating to inclusion of data files that just got fixed.

I think it's already on the most up to date `haskell.nix`. The `haskell-nix` does not update when doing:
```
$ nix flake lock --extra-experimental-features "nix-command flakes" --update-input haskell-nix
```

michaelpj:
> I think it's already on the most up to date haskell.nix. The haskell-nix does not update when doing:

That's because @zeme-iohk has pinned the shas in flake.nix, rather than managing them through updating the lock file. I think that's generally a bad idea for this sort of reason... You're actually pinned to a commit from a month ago.

koslambrou:
> You're actually pinned to a commit from a month ago.

Thanks! And sorry, I should've verified that...

zeme-iohk:
Yes I do that 

> > I think it's already on the most up to date haskell.nix. The haskell-nix does not update when doing:
> 
> That's because @zeme-iohk has pinned the shas in flake.nix, rather than managing them through updating the lock file. I think that's generally a bad idea for this sort of reason... You're actually pinned to a commit from a month ago.

yes it's a bad idea, I'll take care of this.

koslambrou:
@zeme-iohk I think you're haskell.nix pin is still not new enough. My `lock` command generated:

```
    "haskell-nix": {
      "flake": false,
      "locked": {
        "lastModified": 1667533063,
        "narHash": "sha256-EHZOvFNdfgA8U3xNKZkyGZT7Ydnr/3pihMDyGEwLlag=",
        "owner": "input-output-hk",
        "repo": "haskell.nix",
        "rev": "e601c9ce609af07a78edf1c57f5985931788aeb2",
        "type": "github"
      },
      "original": {
        "owner": "input-output-hk",
        "repo": "haskell.nix",
        "type": "github"
      }
    },
```

which more up to date than yours (`"lastModified": 1665056319`)

zeme-iohk:
Ok bumping again 🤞

koslambrou:
But you'll still have other errors :D Mainly the playground-client and the plutus-pab.

flake.nix
```diff
@@ -13,17 +13,13 @@
     # We intentionally import nixpkgs and haskell.nix as non-flakes, to match the
     # flake-free normal build workflow exactly.
     nixpkgs = {
-      type = "github";
-      owner = "NixOS";
-      repo = "nixpkgs";
-      ref = "nixpkgs-unstable";
+      url = "github:NixOS/nixpkgs";
```
michaelpj, line 16: This seems unlikely to be right. You almost certainly want to follow at least `nixpkgs-unstable`, not master of nixpkgs!!

zeme-iohk: I mirrored exactly what we have in `plutus`, should I change it here/there/both?

michaelpj: IMO what's in `plutus` is also wrong then, sorry I missed that. What's there _could_ be reasonable if you're always going to bump nixpkgs to specific known commit, but it's just asking for people to update following nixpkgs master, which is almost always a bad idea (`nixpkgs-unstable` only advances when some checks pass, so is _much_ less likely to be broken than master!).

zeme-iohk: ok I'll change it 👍

zeme-iohk:
More errors fixed, only 3 to go...
Currently battling with a `gcc compilation error` 😭 
https://hydra.iohk.io/build/21471143/nixlog/1

nix/tests/vm.nix
```diff
@@ -14,4 +14,4 @@ let
     all = pkgs.callPackage ./vm-tests/all.nix { inherit makeTest plutus-playground web-ghc docs vmCompileTests; };
   };
 in
-if isDarwin then { } else tests
```
zeme-iohk, line 17: As discussed, Plutus-playground is dying so these tests are obsolete.
In the next PR I'll actually delete the playground and related nix files

nix/pkgs/web-ghc/default.nix
```diff
@@ -11,7 +11,9 @@ let
   ] ++ (extraPackagesFun ps));
 
   runtimeGhcWrapped = writeShellScriptBin "runghc" ''
-    export PATH=${lib.makeBinPath [ bubblewrap runtimeGhc util-linux ]}
+    export PATH=$PATH:${lib.makeBinPath [ bubblewrap runtimeGhc util-linux ]}
+    echo $PATH 
+    echo AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```
sjoerdvisscher, line 16: Is this debugging info that should have been removed?

zeme-iohk: This file is getting deleted in the next PR

PR #758, opened 2022-10-14
PLT-713: must mint value with reference

ghost:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-use-cases/test/Spec.hs
+++ b/plutus-use-cases/test/Spec.hs
@@ -34,7 +34,7 @@ main = defaultMain tests
 -- You can override this number for a specific property test by using
 -- 'Test.Tasty.Quickcheck.withMaxSuccess'.
 limit :: Int
-limit = 100
+limit = 50
 
 tests :: TestTree
 tests =


plutus-contract/test/Spec/Contract/TxConstraints.hs
```diff
@@ -83,6 +83,19 @@ tests = testGroup "contract tx constraints"
             void $ activateContract w1 mustReferenceOutputV2ConTest tag
             void $ Trace.waitNSlots 3
 
+    , checkPredicateOptions
+        (changeInitialWalletValue w1 (const $ Ada.adaValueOf 1000) defaultCheckOptions)
+        "mustSpendScriptOutputWithReference can be used on-chain to unlock funds in a PlutusV2 script"
+        -- TODO: 2nd tx fails because the emulator validation doesn't understand reference scripts,
+        -- Reenable following lines when the emulator validation has been removed
```
sjoerdvisscher, line 90: Wait, you already did this, right?

plutus-contract/test/Spec/TxConstraints/MustMint.hs
```diff
@@ -94,6 +112,60 @@ mustMintCurrencyContract = do
...
+mustReferenceOutputV2Validator = PV2.mkValidatorScript
+    $$(PlutusTx.compile [|| wrap ||])
+ where
+     wrap = PV2.mkUntypedValidator mkMustReferenceOutputV2Validator
+
+mustReferenceOutputV2ValidatorAddress :: Address
+mustReferenceOutputV2ValidatorAddress =
+    PV2.mkValidatorAddress mustReferenceOutputV2Validator
+
+mustMintValueWithReferenceContract :: Contract () Empty ContractError ()
```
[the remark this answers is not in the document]
sjoerdvisscher: You don't need the full `TxInput`, a `Versioned TxOutRef` will do as a witness.

I wonder if it is enough to change `txMintingScripts`  to `Map MintingPolicyHash (Redeemer, Maybe (Versioned TxOutRef))`

You certainly shouldn't have to change the diskState of the chain index. It is not a global thing to map from a script hash to a reference script, that concept is local to the transaction.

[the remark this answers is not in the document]
sjoerdvisscher: You theoretically could use the same script for input witness and minting witness, and if you use a reference script in the minting case but a regular script witness in the input case, you get a conflict for what to store in `txScripts`. That very likely won't happen in practice but it shows that there's a tension here.

plutus-contract/test/Spec/TxConstraints/MustMint.hs
```diff
@@ -94,6 +106,55 @@ mustMintCurrencyContract = do
...
+mustMintValueWithReferenceContract = do
+    utxos <- ownUtxos
+    myAddr <- Con.ownAddress
+    let (_, _, _, _, _, (utxoRef, utxo)) = get6 $ Map.toList utxos
+        MintingPolicyHash mph = coinMintingPolicyHashV2
+        redeemer = asRedeemer $ MustMintCurrencyWithRedeemer coinMintingPolicyHashV2 unitRedeemer tknName tknAmount
+        lookups0 = Constraints.plutusV2MintingPolicy mustMintPolicyV2
+                <> Constraints.plutusV2MintingPolicy coinMintingPolicyV2
+        tx0 = Constraints.mustMintCurrencyWithRedeemer mustMintPolicyHashV2 redeemer tknName 1
+           <> Constraints.mustMintCurrencyWithRedeemer coinMintingPolicyHashV2 unitRedeemer tknName tknAmount
```
koslambrou, line 139: Seems like all of these lines are not necessary, afaiu. For the 1st transaction, you simply need to create an UTXO which contains an inlined minting script.

plutus-ledger-constraints/src/Ledger/Constraints/TxConstraints.hs
```diff
@@ -635,6 +643,18 @@ mustMintCurrency
     -> TxConstraints i o
 mustMintCurrency mps = mustMintCurrencyWithRedeemer mps unitRedeemer
 
+{-# INLINABLE mustMintCurrencyWithReference #-}
+-- | Same as 'mustMintCurrencyWithReference', but sets the redeemer to the unit
```
koslambrou, line 647: mustMintCurrencyWithReference -> mustMintValueWithRedeemerAndReference

plutus-ledger-constraints/src/Ledger/Constraints/TxConstraints.hs
```diff
@@ -609,19 +609,27 @@ mustPayToOtherScriptAddressWithInlineDatum vh svh dv vl =
...
 -- | Same as 'mustMintCurrencyWithRedeemer', but uses the minting policy hash,
 -- token name and amount provided by 'Value'.
 --
 -- Note that we can derive the 'MintingPolicyHash' from the 'Value'\'s currency
 -- symbol.
 mustMintValueWithRedeemer :: forall i o. Redeemer -> Value -> TxConstraints i o
-mustMintValueWithRedeemer red =
+mustMintValueWithRedeemer red = mustMintValueWithRedeemerAndReference red Nothing
+
+{-# INLINABLE mustMintValueWithRedeemerAndReference #-}
```
koslambrou, line 625: Missing comment

plutus-ledger-constraints/src/Ledger/Constraints/OnChain/V1.hs
```diff
@@ -111,7 +111,7 @@ checkTxConstraint ctx@ScriptContext{scriptContextTxInfo} = \case
         -- gives us the redeemer's hash, but 'MustSpendScriptOutput' gives
         -- us the full redeemer
         $ isJust (V.findTxInByTxOutRef txOutRef scriptContextTxInfo)
-    MustMintValue mps _ tn v ->
+    MustMintValue mps _ tn v _ ->
```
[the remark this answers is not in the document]
james-iohk: I think it's ok to assume it'll only get this far without a reference script in the tx. The same is done for MustSpendScriptOutput.

plutus-contract/test/Spec/TxConstraints/MustMint.hs
```diff
@@ -50,6 +61,7 @@ tests =
         , mustMintValueWithRedeemerSuccessfulMint
         , mustMintValueWithRedeemerSuccessfulBurn
         , mustMintValueSuccessfulMint
+        , mustMintWithReferenceSuccessful
```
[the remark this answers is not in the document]
james-iohk: Yes please. A test failing the new V2 phase-2 check would also be good.

This file is still to be updated to use `Versioned MintingPolicy`, that can be done separately.

plutus-contract/test/Spec/TxConstraints/MustMint.hs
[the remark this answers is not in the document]
sjoerdvisscher: With transaction inputs we are moving away from putting scripts in the TxInType (becoming TxInputType), so for consistency it's better to keep all the scripts in `txScripts`. It's just that when there's a `Versioned TxOutRef` in `txMintingScripts` for the minting policy hash, we don't have to look in `txScritps` anymore. Maybe we should rename `txMintingScripts` to `txMintingWitnesses`?

plutus-ledger-constraints/src/Ledger/Constraints/OnChain/V1.hs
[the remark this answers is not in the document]
sjoerdvisscher: I agree with @james-iohk that we don't have to check because when it is not `Nothing` construction of the transaction would have already failed.

Of course it is possible to construct the transaction yourself and use this constraint only in the script, and then a "cheating" transaction could store the minting script in the body instead of using the reference script. Perhaps this could theoretically be a security risk in some way? I think it's fine to just document that `MustMintValue` does not check on-chain that a reference script is used, instead of wasting valuable script resources to do the check.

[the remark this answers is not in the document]
james-iohk: > store the minting script in the body instead of using the reference script

This wouldn't pass phase-1 because reference inputs are prohibited when V1 scripts are involved.
Even if using V2, you cannot have the same script in the txbody and as reference.

plutus-ledger/src/Ledger/Tx/CardanoAPI.hs
```diff
@@ -127,6 +127,16 @@ toCardanoScriptWitness datum redeemer (P.Versioned script lang) = (case lang of
   ) <*> pure datum
     <*> pure (C.fromPlutusData $ PV1.toData redeemer)
     <*> pure zeroExecutionUnits
+toCardanoScriptWitness datum redeemer (Right (P.Versioned ref lang)) = (case lang of
```
sjoerdvisscher, line 130: It should be possible to merge this code with the code for `toCardanoTxInWitness`.

plutus-contract/test/Spec/TxConstraints/MustMint.hs
```diff
@@ -260,14 +321,31 @@ mkMustMintPolicy t = case t of
     MustMintValueWithRedeemer r v           -> Constraints.checkScriptContext @() @() (Constraints.mustMintValueWithRedeemer r v)
     MustMintValue v                         -> Constraints.checkScriptContext @() @() (Constraints.mustMintValue v)
 
+{-# INLINEABLE mkMustMintPolicyV2 #-}
+mkMustMintPolicyV2 :: ConstraintParams -> PV2.ScriptContext -> Bool
+mkMustMintPolicyV2 t = case t of
+    MustMintCurrencyWithRedeemer mph r tn i -> TCV2.checkScriptContext @() @() (Constraints.mustMintCurrencyWithRedeemer mph r tn i)
+    MustMintCurrency mph tn i               -> TCV2.checkScriptContext @() @() (Constraints.mustMintCurrency mph tn i)
+    MustMintValueWithRedeemer r v           -> TCV2.checkScriptContext @() @() (Constraints.mustMintValueWithRedeemer r v)
+    MustMintValue v                         -> TCV2.checkScriptContext @() @() (Constraints.mustMintValue v)
```
james-iohk, line 330: Will want to add the new `ConstraintParams` to use in the policy:
- mustMintCurrencyWithRedeemerAndReference
- mustMintValueWithRedeemerAndReference
- mustMintValueWithReference
- mustMintCurrencyWithReference

Then you want to use one for your positive case test `mustReferenceOutputV2Validator` and for one test demonstrating phase-2 failure due to unexepcted reference txo.

plutus-contract/test/Spec/TxConstraints/MustMint.hs
```diff
@@ -94,6 +106,47 @@ mustMintCurrencyContract = do
...
+mustReferenceOutputV2ValidatorAddress =
+    PV2.mkValidatorAddress mustReferenceOutputV2Validator
+
+mustMintValueWithReferenceContract :: Contract () Empty ContractError ()
+mustMintValueWithReferenceContract = do
+    utxos <- ownUtxos
+    myAddr <- Con.ownAddress
+    let (utxoRef, utxo) = head $ drop 5 $ Map.toList utxos
+        MintingPolicyHash mph = coinMintingPolicyHashV2
+        lookups0 = Constraints.plutusV2MintingPolicy coinMintingPolicyV2
```
james-iohk, line 131: Should also use a minting policy with one of the new constraints to prove the phase-2 check passes.

plutus-contract/test/Spec/TxConstraints/MustMint.hs
```diff
@@ -94,6 +106,47 @@ mustMintCurrencyContract = do
...
+                (ScriptHash mph)
+                Nothing
+                (Ada.adaValueOf 30)
+    ledgerTx0 <- submitTxConstraintsWith @UnitTest lookups0 tx0
+    awaitTxConfirmed $ Tx.getCardanoTxId ledgerTx0
+
+    utxos' <- ownUtxos
+    let
+        refScriptUtxo = fst . head . filter (isJust . Tx._ciTxOutReferenceScript . snd) . Map.toList $ utxos'
+        lookups1 = Constraints.unspentOutputs (Map.singleton utxoRef utxo <> utxos')
```
james-iohk, line 143: Should also use a minting policy with one of the new constraints to prove the phase-2 check passes.

sjoerdvisscher (approved):

james-iohk (approved):

koslambrou (approved):

berewt (approved):

PR #762, opened 2022-10-18
Fix Ledger.Tx.getCardanoTxData

koslambrou:
Fix Ledger.Tx.getCardanoTxData so that the cardano-api transactions returns all the datums of the TxBody, including the datums of the Plutus scripts.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-ledger/src/Ledger/Tx.hs
+++ b/plutus-ledger/src/Ledger/Tx.hs
@@ -76,7 +76,7 @@ import Data.Data (Proxy (Proxy))
 import Data.Default (def)
 import Data.Map (Map)
 import Data.Map qualified as Map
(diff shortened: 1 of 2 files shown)


sjoerdvisscher (approved):
Nice! This should have been part of the PR that replaced `TxIn` with `TxInput`. We're probably really close to remove `TxIn`.

koslambrou:
@sjoerdvisscher Curious, how does this bring us closer to remove `TxIn`?

sjoerdvisscher:
@koslambrou We were storing some data in TxInType and some in txData, which was confusing everyone. This is one step of moving everything to txData, and then we won't need TxIn anymore (I hope).

PR #765, opened 2022-10-19
PLT-658: inline datum witnesses

sjoerdvisscher:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


plutus-ledger/src/Ledger/Typed/Scripts.hs
```diff
@@ -1,45 +1,45 @@
...
+mkForwardingMintingPolicy vl@(Versioned _ PlutusV1) = Versioned (PV1.mkForwardingMintingPolicy (Untyped.validatorHash vl)) PlutusV1
+mkForwardingMintingPolicy vl@(Versioned _ PlutusV2) = Versioned (PV2.mkForwardingMintingPolicy (Untyped.validatorHash vl)) PlutusV2
 
-instance Eq (DatumType a) => Eq (TypedScriptTxIn a) where
-  l == r =
-    tyTxInTxIn l == tyTxInTxIn r
-      && tyTxInOutRef l == tyTxInOutRef r
+-- | Make a 'TypedValidator' (with no type constraints) from an untyped 'Validator' script.
+unsafeMkTypedValidator :: Versioned Validator -> TypedValidator Any
+unsafeMkTypedValidator vl =
```
berewt, line 35: Not really related to the PR but still: did we officially decide of a strategy regarding the content of `Ledger.Typed.Scripts` vs `PSU.Vx.Typed.Scripts` modules? The deployment of V2 scripts may require clarification. We have an epic about cleaning the code after the Babbage era release, maybe we need to add a review/clean up/harmonisation of these modules to it? (cc @koslambrou)

berewt (approved):
No blocker. The major concern I have is about the organisation of PSU and `Ledger.Typed.Script` but it could be discussed outside of thisPR.

plutus-ledger/src/Ledger/Typed/Scripts.hs
sjoerdvisscher: I'd say at least keep the version specific things as much as possible in `PSU.Vx.Typed.Scripts`, and let the rest work with `Versioned` scripts. It's tempting to export "the latest version" code from `Ledger.Typed.Scripts` too, but I'm not sure we should do that, and at least it would silently break a lot of our tests at the moment. (But that's a bug really, tests that depend on something being version 1 should explicitly import from `V1` modules.)

plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
```diff
@@ -781,24 +779,23 @@ processConstraintFun = \case
...
+                unbalancedTx . tx %= Tx.addScriptTxInput ref validator red (if isInline then Nothing else Just datum)
                 valueSpentInputs <>= provided value
             _ -> throwError $ MultipleMatchingOutputsFound vh
 
 resolveScriptTxOut
     :: ( MonadReader (ScriptLookups a) m
        , MonadError MkTxError m
        )
-    => ChainIndexTxOut -> m (Maybe (Versioned Validator, Datum, Value))
+    => ChainIndexTxOut -> m (Maybe (Versioned Validator, (Datum, Bool), Value))
```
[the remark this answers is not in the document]
sjoerdvisscher: I'm not sure, the `Bool` really belongs to the `Datum`. I think I should just make yet another specific datatype. (It's really annoying that als these datum-related types have to be slightly different!)

plutus-ledger/src/Ledger/Typed/Scripts.hs
berewt: > I'd say at least keep the version specific things as much as possible in `PSU.Vx.Typed.Scripts`, and let the rest work with `Versioned` scripts. It's tempting to export "the latest version" code from `Ledger.Typed.Scripts` too, but I'm not sure we should do that, and at least it would silently break a lot of our tests at the moment. (But that's a bug really, tests that depend on something being version 1 should explicitly import from `V1` modules.)

Yes I was thinking of something alike, it would be quite aligned with Konstantinos' ADR about cardano API as well. We should probably stick to this behaviour in `Ledger.Test` as well (and maybe in other places, but I don't see where). At least it would give us a set of clearer rules to deal with plutus versions. This would probably imply moving Versioned to plutus-ledger as well?

plutus-ledger/src/Ledger/Tx/Internal.hs
```diff
@@ -61,9 +61,10 @@ import Prettyprinter (Pretty (..), hang, viaShow, vsep, (<+>))
 
 -- | The type of a transaction input.
 data TxInType =
-      ScriptAddress !(Either (Versioned Validator) (Versioned TxOutRef)) !Redeemer !Datum
+      ScriptAddress !(Either (Versioned Validator) (Versioned TxOutRef)) !Redeemer !(Maybe Datum)
       -- ^ A transaction input that consumes (with a validator) or references (with a txOutRef)
       -- a script address with the given the redeemer and datum.
+      -- Datum is optional in case the input references an inline datum.
```
koslambrou, line 67: Mmm... What do you mean here _exactly_?

Do you actually mean "Datum is optional if the input refers to a script output which contains an inline datum"?

plutus-ledger/src/Ledger/Tx/Internal.hs
```diff
@@ -532,22 +534,24 @@ addMintingPolicy vvl rd tx@Tx{txMintingScripts, txScripts} = tx
         mph@(MintingPolicyHash b) = mintingPolicyHash vvl
 
 -- | Add validator together with the redeemer and datum into txInputs, txData and txScripts accordingly.
-addScriptTxInput :: TxOutRef -> Versioned Validator -> Redeemer -> Datum -> Tx -> Tx
-addScriptTxInput outRef vl rd dt tx@Tx{txInputs, txScripts, txData} = tx
-    {txInputs = TxInput outRef (TxScriptAddress rd (Left vlHash) dtHash) : txInputs,
+-- Datum is optional when the reference points to an inline datum.
```
koslambrou, line 537: Same comment as above

plutus-ledger/src/Ledger/Tx/Internal.hs
```diff
@@ -532,22 +534,24 @@ addMintingPolicy vvl rd tx@Tx{txMintingScripts, txScripts} = tx
...
-        dtHash = datumHash dt
+        mdtHash = fmap datumHash mdt
         vlHash@(ValidatorHash b) = validatorHash vl
 
 -- | Add script reference together with the redeemer and datum into txInputs and txData accordingly.
-addReferenceTxInput :: TxOutRef -> Versioned TxOutRef -> Redeemer -> Datum -> Tx -> Tx
-addReferenceTxInput outRef vref rd dt tx@Tx{txInputs, txData} = tx
-    {txInputs = TxInput outRef (TxScriptAddress rd (Right vref) dtHash) : txInputs,
-     txData = Map.insert dtHash dt txData}
+-- Datum is optional when the reference points to an inline datum.
```
koslambrou, line 548: Same comment as above.

koslambrou (approved):
LGTM. Just documentation comments.

Issue #696, opened 2022-09-05
Emulator's ChainState produces an EmulatorState with empty WalletState Map


james-iohk:
### Summary

Using the Emulator's ChainState to produce an EmulatorState results in an empty WalletState Map. This is probably minor because there is `agentState` for this purpose, but still good to investigate.

### Steps to reproduce the behavior

There is a commented out test to recreate this bug, see here: **include once merged**.

### Actual Result

```haskell
let trace :: Trace.EmulatorTrace () = do
        thisChainState <- Trace.chainState
        let traceBlockchain = thisChainState ^. chainNewestFirst
            traceEmulatorState = emulatorState traceBlockchain
            walletStateMap = traceEmulatorState ^. walletStates
            w1State = fromJust $ M.lookup w1 walletStateMap -- Fails here: Maybe.fromJust: Nothing
```

### Expected Result

It is possible to get all `TxOutRef` for any wallet from Trace's ChainState.

### System info

**Include git commit once merged**

berewt:
Indeed `emulatorState` only recreates the `chainState` of an `EmulatorState`. Its main purpose is to perform a partial initialisation of the `EmulatorState` when we start the emulator, not to provide access to the whole `emulatorState`.
Its behaviour is aligned with the behaviour of the other initializer: create the chain index and update the wallets through the `EmulatorEvent`s. In this particular case, as we don't replay the events on the newly (re)created state, the wallet list is empty.

We have already discussed the need to clean up and clarify the API, especially to clarify which parts of it are public. I think it's a pretty good example of it. I suggest that we remove the test, and that we initiate an action item (epic or story (?)) about cleaning and clarifying the API.

PR #766, opened 2022-10-19
PLT-797 Fix the test(s) for bug #695

### LINKED ISSUE #695: Wallet's utxos in Emulator's diskstate have incorrect indexes
### Summary

Accessing the wallet's utxos from the WalletState of the EmulatorTrace results in TxOutRefs with indexes in the range 0-9 for all wallets, this cannot be correct because each wallet must have unique utxos. The Contract sees unique TxOutRefs for each wallet, e.g. w1 uses index range 50-59 and w2 uses 20-29.

There are currently workarounds for this bug in existing tests `mustSpendPubKeyOutput` plutus-context tests, such as [this](https://github.com/input-output-hk/plutus-apps/blob/a8540dbc92ecf5e6883eb30e3c47efab3f35b446/plutus-contract/test/Spec/TxConstraints/MustSpendPubKeyOutput.hs#L151). Please remove workarounds once fixed.

There is a commented out test to recreate this bug, [see here](https://github.com/input-output-hk/plutus-apps/blob/a8540dbc92ecf5e6883eb30e3c47efab3f35b446/plutus-contract/test/Spec/TxConstraints/MustSpendPubKeyOutput.hs#L108-L120). _Edit: this also now affects MustReferenceOutput tests_

### Actual Result

Getting `TxOutRef` from `WalletState`,  `w1TxoRefs` will have indexes in range 0-9. 
```haskell
let trace :: Trace.EmulatorTrace () = do
        w1State <- Trace.agentState w1
        let w1TxoRefs = S.elems $ head $ M.elems $ w1State ^. chainIndexEmulatorState . diskState . addressMap . unCredentialMap
```
Logging in Contract will print indexes in range 50-59
```haskell
logInfo @String $ "pubKeyUtxos:: " ++ show pubKeyUtxos
```

### Expected Result

Wallets' utxos should be consistent.

_No response_

### System info

https://github.com/input-output-hk/plutus-apps/commit/a8540dbc92ecf5e6883eb30e3c47efab3f35b446

berewt:
Fixes #695 

The bug is not really a bug, but rather a lack of API documentation. The right call to the API provides the correct outputs. 
I've modified the tests accordingly and documented `WalletState` to mitigate the risk of misinterpreting its content.

2 questions:

1. do we keep the bug test or do we remove it?
2. do we want to review how information are stored in `WalletState`?

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


koslambrou (approved):

koslambrou:
1. I think we can remove bug695 test case
2. What exactly do you want to review?

berewt:
>     2. What exactly do you want to review?

There seems to be some information duplication in the emulator state between `ChainState` and `WalletStates` and it can be confusing (like, in this case, when a wallet state contains information about the whole blockchain). We also have some difficulty providing a coherent state across these duplications (see #696).
So, the main argument in favour of revisiting the `EmulatorState` is: we can probably carry less information, which could speed up the emulator and reduce discrepancies inside a state. The cons is the effort of revisiting the free monad stack and the chance to break things that are currently working.

james-iohk (changes requested):
Thanks for removing the index override @berewt
Please may you also apply this to the MustReferenceOutput tests too?

james-iohk (approved):

PR #772, opened 2022-10-21
PLT-768 Renamed currentTime to currentNodeClientTimeRange and renamed currentPABSlot to currentNodeClientSlot

koslambrou:
Renamed currentTime to currentNodeClientTimeRange and renamed currentPABSlot to currentNodeClientSlot

* Deprecated 'Plutus.Contract.Request.currentTime'

* Created 'Plutus.Contract.Request.currentNodeClientTimeRange' which replaces 'currentTime'

* Deprecated 'Plutus.Contract.Request.currentPABSlot'

* Created 'Plutus.Contract.Request.currentNodeClientSlot' which replaces 'currentPABSlot'

* Updated the plutus use case examples

* Updated the PABReq and PABResp request handlers for the emulator and the PAB

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-contract/src/Plutus/Contract/Request.hs
```diff
@@ -214,21 +216,31 @@ isSlot ::
...
 
 -- | Get the current slot number
-{-# DEPRECATED currentSlot "It was renamed to 'currentPABSlot', this function will be removed" #-}
+{-# DEPRECATED currentSlot "Use currentNodeClientSlot instead" #-}
 currentSlot ::
     forall w s e.
     ( AsContractError e
     )
     => Contract w s e Slot
 currentSlot = currentPABSlot
```
berewt, line 225: Do we have to remove this one? It's a nice and short name to get the current slot, and we kinda know it's the client node's one (we can still precise it in the doc if needed)

plutus-contract/src/Plutus/Contract/Request.hs
```diff
@@ -283,6 +296,17 @@ currentTime ::
...
 
+-- | Get the 'POSIXTime' range of the current slot.
+--
+-- Example: if slot length is 3s and current slot is 2, then `currentTimeRange`
+-- returns the time interval @[3, 5[@.
+currentNodeClientTimeRange ::
+    forall w s e.
+    ( AsContractError e
+    )
+    => Contract w s e (POSIXTime, POSIXTime)
```
berewt, line 307: As the doc says interval, and as being open or closed may be a relevant info, shouldn't we return an `Interval` instead? (and probably re-export the Interval API to ease their manipulation)

koslambrou: Happy to change it, but after looking at `Interval`, it seemed to me harder to manipulate an `Interval` than (POSIXTime, POSIXTme), especially when you want to get the actual lower bound of the interval. The reason is that the lowerbound can be `-Inf`, `Finite` or `Inf`. Same for the upper bound.

Thoughts?

berewt: Oh yes you're right. It's asd that we cant have the boundaries closure to be really explicit about it, but I don't think it's sad enough to create a new type for this, so I guess we can go with a pair.

plutus-contract/src/Plutus/Contract/Request.hs
koslambrou: Honestly, I prefer explicit naming.

The name `currentPABSlot` is ambiguous as it can refer to the slot that is currently indexed by PAB, or the slot of the local running node. In the current implementation, it returns the slot of the local running, hence why I changed the name to make it more clear. If we really need the slot that PAB has currently indexed, then yes we can create `currentPABSlot`, but there is not actual use case for that.

Also, PAB is too specific to the `plutus-pab` package. And since it's future is unclear, might as well change the name now.

berewt: `currentPABSlot` is bad, I was advocating for `currentSlot` here. But if you think it can be misleading, we can deprecate it, no problem. :)

plutus-playground-server/usecases/Vesting.hs
```diff
@@ -207,7 +207,6 @@ retrieveFundsC vesting payment = do
                 -- we don't need to add a pubkey output for 'vestingOwner' here
                 -- because this will be done by the wallet when it balances the
                 -- transaction.
-    void $ waitNSlots 1 -- wait until we reach a slot in the validity range
```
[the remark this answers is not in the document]
koslambrou: Yep :)

With he old implementation, doing `now <- currentTime` and `mustValidateIn (from now)` will call `posixTimeRangeToContainedSlotRange (from now)` which will return the next slot for the lower bound.

So if
* `currentSlot` is 1
* slot config has start time of 0 and slot length of 1000

`currentTime` will return `1999` (end time of the current slot)

Then doing `posixTimeRangeToContainedSlotRange (from 1999)` will yield slot range [2, inf]. Hence why we have to wait for one slot before submitting the transaction.

plutus-contract/src/Plutus/Contract/Request.hs
koslambrou: Oh! We actually used to have `currentSlot` and end-users were confused as to what that refers to (PAB? chain-index? node client?).

berewt: Ok, so it's preferable to deprecate it in this case.

PR #779, opened 2022-10-24
PLT-989 Improve behaviour of duplicated `MustSpend`

berewt:
Throw errors if we try to use different redeemer or different reference script in different MustSpend, ignore all but one constraint otherwise.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-ledger-constraints/src/Ledger/Constraints/TxConstraints.hs
+++ b/plutus-ledger-constraints/src/Ledger/Constraints/TxConstraints.hs
@@ -697,6 +697,9 @@ mustProduceAtLeast = singleton . MustProduceAtLeast
 -- the 'Ledger.Constraints.OffChain.ScriptLookups' with
 -- 'Ledger.Constraints.OffChain.unspentOutputs'.
 --
+-- If several calls to 'mustSpendPubKeyOutput' are performed for the same 'TxOutRef',
+-- only one instance of the constraint is kept when the transaction is created.
+--
 -- If used in 'Ledger.Constraints.OnChain', this constraint verifies that the
 -- transaction spends this @utxo@.
 mustSpendPubKeyOutput :: forall i o. TxOutRef -> TxConstraints i o
@@ -714,6 +717,11 @@ mustSpendPubKeyOutput = singleton . MustSpendPubKeyOutput
 -- 'Ledger.Constraints.OffChain.unspentOutputs' or through
 -- 'Ledger.Constraints.OffChain.otherData'.
 --
+-- If several calls to 'mustSpendScriptOutput' are performed for the same 'TxOutRef',
+-- if the two constraints have different redeemers, an error will be thrown when the transaction will be submitted.
+-- Otherwise, only one instance of the constraint is kept.
+-- If combine with 'mustSpendScriptOutputWithReference' for the same 'TxOutRef', see 'mustSpendScriptOutputWithReference'.
+--
 -- If used in 'Ledger.Constraints.OnChain', this constraint verifies that the
 -- transaction spends this @utxo@.
 mustSpendScriptOutput :: forall i o. TxOutRef -> Redeemer -> TxConstraints i o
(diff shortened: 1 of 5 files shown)


koslambrou:
Did you think of the alternative of "simply" ignoring all  but one `mustSpendScriptOutput` call instead of throwing errors?

berewt:
> Did you think of the alternative of "simply" ignoring all but one `mustSpendScriptOutput` call instead of throwing errors?

Yes. As discussed in JIRA, I'll move back to this behaviour instead, it sounds more consistent with the lookup behaviour.

berewt:
Or, I keep it as-is (@james-iohk sounds ok with it) but in any case, I need to document the `mustSpend` constraints, which is no done yet.

sjoerdvisscher:
I think that's a bad idea. You might have used the wrong variable somewhere. A missing lookup is quickly found out, but a duplicate `mustSpendScriptOutput` could go unnoticed.

berewt:
Ok, let's keep it as is then. I'll add documentation to the smart builders and ship it.

plutus-ledger-constraints/src/Ledger/Constraints/TxConstraints.hs
```diff
@@ -714,6 +717,11 @@ mustSpendPubKeyOutput = singleton . MustSpendPubKeyOutput
 -- 'Ledger.Constraints.OffChain.unspentOutputs' or through
 -- 'Ledger.Constraints.OffChain.otherData'.
 --
+-- If several calls to 'mustSpendScriptOutput' are performed for the same 'TxOutRef',
+-- if the two constraints have different redeemers, an error will be thrown when the transaction will be submitted.
+-- Otherwise, only one instance of the constraint is kept.
+-- If combine with 'mustSpendScriptOutputWithReference' for the same 'TxOutRef', see 'mustSpendScriptOutputWithReference'.
```
sjoerdvisscher, line 723:
```suggestion
-- If combined with 'mustSpendScriptOutputWithReference' for the same 'TxOutRef', see 'mustSpendScriptOutputWithReference'.
```

sjoerdvisscher (approved):

plutus-ledger-constraints/src/Ledger/Constraints/TxConstraints.hs
```diff
@@ -714,6 +717,11 @@ mustSpendPubKeyOutput = singleton . MustSpendPubKeyOutput
 -- 'Ledger.Constraints.OffChain.unspentOutputs' or through
 -- 'Ledger.Constraints.OffChain.otherData'.
 --
+-- If several calls to 'mustSpendScriptOutput' are performed for the same 'TxOutRef',
+-- if the two constraints have different redeemers, an error will be thrown when the transaction will be submitted.
```
koslambrou, line 721: I think the error is not thrown when the transaction is submitted, but when the transaction is created with `mkTx`.

plutus-ledger-constraints/src/Ledger/Constraints/TxConstraints.hs
```diff
@@ -731,6 +739,17 @@ mustSpendScriptOutput txOutref red = singleton $ MustSpendScriptOutput txOutref
 -- 'Ledger.Constraints.OffChain.unspentOutputs' or through
 -- 'Ledger.Constraints.OffChain.otherData'.
 --
+-- If several calls to 'mustSpendScriptOutputWithReference' are performed for the same 'TxOutRef',
+-- if the two constraints have different redeemers,
+-- or if the two constraints use a different 'TxOutRef' as a TxOutRef, an error will be thrown when the transaction will
+-- be submitted.
```
koslambrou, line 745: Same comment as above

koslambrou (approved):
LGTM :) Minor comment

PR #781, opened 2022-10-25
PLT-1034 Proposing an ADR to modify the `MustValidateIn` constraint

koslambrou:
PLT-1034 Proposing an ADR to modify the `MustValidateIn` constraint in `plutus-ledger-constraints` so that it can work with slots instead of POSIXTime.

The link to the ADR is here: https://plutus-apps--781.org.readthedocs.build/en/781/adr/0013-tx-validity-time-range-upper-bound.html

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/doc/adr/0013-tx-validity-time-range-fix.rst
+++ b/doc/adr/0013-tx-validity-time-range-fix.rst
@@ -1,0 +1,199 @@
+.. _tx_validity_time_range_fix:
+
+ADR 13: Transaction validity time range fix
+===========================================
+
+Date: 2022-10-19
+
+Author(s)
+---------
+
+koslambrou <konstantinos.lambrou@iohk.io>
+
+Status
+------
+
+Draft
+
+Context
+-------
+
+The following code samples were executed with ``cabal repl plutus-ledger`` on the plutus-apps commit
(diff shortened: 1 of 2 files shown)


doc/adr/0013-tx-validity-time-range-upper-bound.rst
```diff
@@ -0,0 +1,200 @@
...
+
+  .. code-block:: haskell
+
+    data TxConstraint =
+      ...
+      MustValidateInTimeRange !(ValidityInterval POSIXTime)
+
+    mustValidateInTimeRange :: !(ValidityInterval POSIXTime) -> TxConstraints
+
+* We will deprecate the ``MustValidateIn`` constraint and the ``mustValidateIn`` smart constructor
```
sjoerdvisscher, line 129: Perhaps we should remove `MustValidateIn` immediately, since it would be dead weight in the script size.

sjoerdvisscher (approved):

berewt (approved):

doc/adr/0013-tx-validity-time-range-upper-bound.rst
koslambrou: :+1:

PR #792, opened 2022-10-27
PLT-777 Remove plutus ledger from Marconi' dependencies

berewt:
Here are the major type changes:

* `TxOutRef` (from `plutus-ledger-api`) has been replaced by an alias with the same name to `TxIn` of `cardano-api` that conveys the same information.
* `Address` (from `plutus-ledger`) has been replaced by `AddressAny` from `cardano-api`, that is basically an address in era, without the `era` witness type.
* `TxOut` (from `plutus-ledger`) has been replaced by `TxOut` from `cardana-api`.
* `Datum` and `DatumHash` (from `plutus-ledger-api`) has been replaced by `ScriptData` and `Hash ScriptData` from `cardano-api`.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


berewt:
I'm gonna make a PR to cardano-nod to expose the function I needed to duplicate and fix the corresponding comment in my code once before merging it.

koslambrou:
You're waiting for the node PR to merge before merging this PR?

berewt:
> You're waiting for the node PR to merge before merging this PR?

I think we'd better merge it before the node PR, just to avoid adding new features that rely on `plutus-ledger` in the meantime.

kayvank (approved):
Thank you for doing the work. I specially like :heart: the Marconi.Types and plan on moving some of data structure to there from the Indexers.hs

berewt:
> I'll merge mine after this PR is merged to main

If yours is ready to merge, don't worry and merge it, I'll handle the conflict if needed.

raduom (approved):
LGTM

marconi/src/Marconi/Index/Utxo.hs
```diff
@@ -163,9 +170,10 @@ onInsert _ix _update = pure []
...
-                                 , _reference = ref
-                                 })
+  & map (\(C.TxOut addr _ _ _, ref) ->
+        UtxoRow { _address   = addr
+                , _reference = ref
+                })
 
 onlyAt :: Address -> UtxoRow -> Bool
-onlyAt address' row = address' == (row ^. address)
+onlyAt address' = has $ address . only address'
```
eyeinsky, line 188: I think here the original version was clearer. I don't know what `has` and `only` do so to know it I'd have to go and look at lens documentation and even then it will take a bit. The original version on the other hand pretty much says what it does.

marconi/src/Marconi/CardanoAPI.hs
```diff
@@ -0,0 +1,47 @@
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE PatternSynonyms    #-}
+{-# LANGUAGE TypeFamilies       #-}
+-- | This module provides several type aliases and utility functions to deal with the current Cardano era.
+module Marconi.CardanoAPI
+       (
+       CurrentEra,
+       pattern AsCurrentEra,
+       C.TxIn(C.TxIn),
```
eyeinsky: I think this isn't worth it: exporting TxIn from here and using it imported Marconi.CardanoAPI in other modules will give rise to the question of how is it different from the one in Cardano.Api, and the answer is that there are no differences. Maybe when we get the package `cardano-api-extras` that will serve as a front for cardano-api that we directly control and which' contents we want to upstream, it would make sense to re-export Cardano.Api types from there, but otherwise I think it's better to use things from their original location as it builds upon knowledge on having used those modules earlier.

berewt: I hesitated to do so for that exact reason. The pros were that we centralised our dependencies on cardano-api, whether or not these types are era-dependent. But you're right, even if reexporting is a common practice, it probably doesn't worth the effort here.

marconi/src/Marconi/Index/Utxo.hs
berewt: No problem, I have this bad habit of putting everything in one lens when I see one usage of them in an expression, I'm reverting it.

marconi/src/Marconi/CardanoAPI.hs
eyeinsky: Reexporting is fine for exporting definitions from modules that are otherwise internal, or to provide a nice API from a single location. But Cardano.Api is not internal, and it already is _the_ entrypoint to what cardano-api has to offer; and also as Marconi.CardanoAPI doesn't export all of Cardano.Api then most of the time Cardano.Api will already be imported, so having one type go through another module raises the question of why - does it do anything different? - but it doesn't, and it isn't era-dependent either.

eyeinsky (approved):
LGTM!

PR #793, opened 2022-10-27
PLT-1011: Add mkTxWithParams

sjoerdvisscher:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-contract/src/Plutus/Contract.hs
+++ b/plutus-contract/src/Plutus/Contract.hs
@@ -97,6 +97,8 @@ module Plutus.Contract(
     , Request.isTxConfirmed
     -- ** Tx output confirmation
     , Request.awaitTxOutStatusChange
+    -- * Parameters
+    , Request.getParams
     -- * Checkpoints
     , Plutus.Contract.Types.checkpoint
     , Plutus.Contract.Types.checkpointLoop
--- a/plutus-contract/src/Plutus/Contract/Request.hs
+++ b/plutus-contract/src/Plutus/Contract/Request.hs
@@ -99,10 +99,11 @@ module Plutus.Contract.Request(
     , submitTxConfirmed
     , mkTxConstraints
     , yieldUnbalancedTx
+    -- ** Parameters
+    , getParams
     -- * Etc.
     , ContractRow
     , pabReq
-    , mkTxContract
     , MkTxLog(..)
     ) where
 
@@ -126,8 +127,8 @@ import Data.Void (Void)
 import GHC.Generics (Generic)
 import GHC.Natural (Natural)
 import GHC.TypeLits (Symbol, symbolVal)
-import Ledger (AssetClass, DiffMilliSeconds, POSIXTime, PaymentPubKeyHash (PaymentPubKeyHash), Slot, TxId, TxOutRef,
-               Value, addressCredential, fromMilliSeconds, txOutRefId)
+import Ledger (AssetClass, DiffMilliSeconds, POSIXTime, Params, PaymentPubKeyHash (PaymentPubKeyHash), Slot, TxId,
+               TxOutRef, Value, addressCredential, fromMilliSeconds, txOutRefId)
 import Ledger.Constraints (TxConstraints)
 import Ledger.Constraints.OffChain (ScriptLookups, UnbalancedTx)
 import Ledger.Constraints.OffChain qualified as Constraints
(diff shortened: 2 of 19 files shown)


plutus-contract/src/Plutus/Contract/Request.hs
```diff
@@ -965,19 +973,6 @@ data MkTxLog =
         deriving stock (Show, Generic)
         deriving anyclass (ToJSON, FromJSON)
 
--- | Build a transaction that satisfies the constraints
-mkTxConstraints :: forall a w s e.
```
koslambrou, line 969: The `Plutus.Contract.Request` is a public facing API module. We need to deprecate it.

plutus-contract/test/Spec/Balancing.hs
```diff
@@ -58,7 +58,7 @@ balanceTxnMinAda =
                         unitDatum
                         (Value.scale 100 ff <> Ada.toValue Ledger.minAdaTxOut)
                  <> L.Constraints.mustIncludeDatumInTx unitDatum
-                utx1 = either (error . show) id $ L.Constraints.mkTx @Void mempty constraints1
+                utx1 = either (error . show) id $ L.Constraints.mkTxWithParams @Void def mempty constraints1
```
koslambrou, line 61: Instead of `def`, how about using the `getParams` request?

plutus-contract/src/Plutus/Contract.hs
```diff
@@ -97,6 +97,8 @@ module Plutus.Contract(
     , Request.isTxConfirmed
     -- ** Tx output confirmation
     , Request.awaitTxOutStatusChange
+    -- * Parameters
+    , Request.getParams
```
koslambrou, line 101: Oh, and I would suggest the name `getNodeParams` just for clearness

plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
```diff
@@ -424,15 +425,16 @@ data SomeLookupsAndConstraints where
 -- | Given a list of 'SomeLookupsAndConstraints' describing the constraints
 --   for several scripts, build a single transaction that runs all the scripts.
 mkSomeTx
-    :: [SomeLookupsAndConstraints]
+    :: Params
+    -> [SomeLookupsAndConstraints]
     -> Either MkTxError UnbalancedTx
-mkSomeTx xs =
+mkSomeTx params xs =
```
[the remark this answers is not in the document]
berewt: I don't see any reason to export it aside testability. But as it is exposed, depracation and a new `mkSomeTxWithParams` function?

berewt (commented):
LGTM

plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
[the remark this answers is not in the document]
sjoerdvisscher: IMHO we're still too much a beta level library to be doing this deprecation dance. Even for `mkTx` I find it dubious since its type changes, so nothing would silently break and the fix is pretty clear, but it is probably so widely used that it's worth doing the dance.

plutus-contract/src/Plutus/Contract.hs
sjoerdvisscher: Hmm shouldn't `Params` be renamed `NodeParams` then too? (I actually never thought of `Params` as just node parameters, but that's indeed what they are.)

plutus-ledger-constraints/src/Ledger/Constraints.hs
```diff
@@ -59,6 +59,7 @@ module Ledger.Constraints(
     , OC.UnbalancedTx(..)
     , OC.MkTxError(..)
     , OC.mkTx
```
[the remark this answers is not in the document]
sjoerdvisscher: I did that right? Or am I missing something?

plutus-contract/src/Plutus/Contract/Request.hs
sjoerdvisscher: `mkTxConstraints` is not deleted, just moved. `mkTxContract` is deleted, but that had an awkward type and only exported under the "Etc." header, and is probably not used outside of our codebase.

plutus-contract/src/Plutus/Contract.hs
sjoerdvisscher: I can do that in the PR that fixes the PAB to actually query the node, until then `getNodeParams` could falsely imply you'd get the actual values from the node.

koslambrou: Oh, I though we had `NodeParams`, that's why I proposed it :D Then good point about chaning `Params` to `NodeParams`

plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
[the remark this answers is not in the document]
koslambrou: Good point. Since a number of people have complained about us doing too much breaking changes in the API, I though deprecating would be the way to go, even though we're still a beta level library.

I understand that we need a deprecation policy and a well defined public API.

plutus-contract/src/Plutus/Contract/Request.hs
koslambrou: :+1:

plutus-ledger-constraints/src/Ledger/Constraints.hs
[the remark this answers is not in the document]
koslambrou: Oops my bad.

PR #795, opened 2022-10-28
Add evalEmulatorTrace

sjoerdvisscher:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


koslambrou:
What's missing is a test case in `Spec.Emulator` which showcases how it works?

plutus-contract/test/Spec/Emulator.hs
```diff
@@ -314,6 +315,13 @@ pubKeyTransactions2 = do
     Trace.liftWallet wallet1 $ payToPaymentPublicKeyHash_ def W.always (Ada.adaValueOf 20) pubKey2
     void Trace.nextSlot
 
+evalEmulatorTraceTest :: Property
+evalEmulatorTraceTest = property $ do
+    let trace = Trace.payToWallet wallet1 wallet2 (Ada.adaValueOf 10)
+        res = Trace.evalEmulatorTrace def trace
+    Hedgehog.annotateShow res
+    Hedgehog.assert (either (const False) (const True) res)
```
koslambrou, line 323: I think we also need to test that the result returned by the trace is the correct one.

For example, something like

```
let trace = do
     Trace.payToWallet wallet1 wallet2 (Ada.adaValueOf 10)
     pure "DONE" -- Or something else
...
Hedgehog.assert (either (const False) (\x -> x == "DONE") res)
```

plutus-contract/src/Plutus/Trace/Emulator.hs
```diff
@@ -249,6 +269,17 @@ runEmulatorTrace cfg trace =
     $ foldEmulatorStreamM (generalize list)
     $ runEmulatorStream cfg trace
 
+runEmulatorTrace'
```
koslambrou, line 272: I think you want to refactor `runEmulatorTrace` following `runEmulatorTrace'`. Or just replace one with the other.

plutus-contract/src/Plutus/Trace/Scheduler.hs
```diff
@@ -291,35 +290,35 @@ runThreads e = do
...
 -- | Run the threads that are scheduled in a 'SchedulerState' to completion.
-loop :: forall effs systemEvent.
+loop :: forall a effs systemEvent.
     ( Eq systemEvent
     , Member (LogMsg SchedulerLog) effs
     )
-    => SchedulerState effs systemEvent
-    -> Eff effs ()
+    => SchedulerState effs systemEvent a
+    -> Eff effs (Maybe a)
```
koslambrou, line 298: How do you know that result here is the one from the initialThread and not from the wallet threads?

plutus-contract/test/Spec/Emulator.hs
sjoerdvisscher: The type ensures that, it can't pluck a `TxId` from thin air. And the actual value of the TxId can change so I don't wan't to test for that.

plutus-contract/src/Plutus/Trace/Scheduler.hs
sjoerdvisscher: You only get a result from an `exit` call. There's only one place where `exit` is called and that's in `handleEmulatorTrace` which calls it with the result of the incoming `action`.

plutus-contract/src/Plutus/Trace/Emulator.hs
sjoerdvisscher: Thanks, which the old version I couldn't do that, because of the uncertainty of the initial thread returning or not. With the `exit` mechanism the uncertainty is not there so everything can use the same function.

plutus-contract/test/Spec/Emulator.hs
koslambrou: I'm not sure I understand. I wasn't talking about `TxId`. I'm just thinking of returning a value from the trace and then verify that `evalEmulatorTrace` correctly returns that value.

sjoerdvisscher: `Trace.payToWallet` returns an `EmulatorTrace TxId`. Which means `res` is of type `Either EmulatorErr TxId`. So if `res` is a `Right something` that something has to be the `TxId` returned by `payToWallet`.

plutus-contract/src/Plutus/Trace/Scheduler.hs
koslambrou: Then should we mention this somewhere? Comment on `runThreads`?

plutus-contract/test/Spec/Emulator.hs
koslambrou: But doesn't that specific test ensure that you implemented the correct behavior? For example, if you returned the result of the wrong thread in `runThreads` (because `exit` was called by another thread), then you would get the wrong `Right something`, no?

sjoerdvisscher: That could only happen is that other thread would call `exit` with a `TxId`, and that could only be the `TxId` created by `payToWallet`. Replacing the `TxId` with a `String` doesn't change anything.

sjoerdvisscher: That's the power of type variables, theorems for free! https://www.cs.utah.edu/~mflatt/past-courses/cs7520/public_html/s06/wadler89.pdf

PR #626, opened 2022-07-25
Replace Ledger.Tx.ChainIndexTxOut

andreabedini:
tl;dr. 

```
data DecoratedTxOut = DecoratedTxOut
  { decoratedTxOutAddress         :: Address
  , decoratedTxOutValue           :: Value
  , decoratedTxOutValidator       :: Maybe (Either ValidatorHash Validator)
  , decoratedTxOutDatum           :: Maybe (Either DatumHash Datum)
  , decoratedTxOutReferenceScript :: Maybe Script
  }
```

The idea is that at the contract level, we need a `TxOut` with additional information like the resolved validator & datum. I stole the word "decorated" from @michaelpj but I am happy to consider alternatives. In principle one could stick with `TxOut` and rely on qualified names.

Taking a  KISS approach I have defined a type that has all the information. The type is not "tight" sinceit has invariants:

If the address corresponds to a script then
1. the validator field must match
2. the datum must be present

(Note that `decoratedTxOutDatum` still returns a `Maybe`, perhaps we can improve on this).

Therefore the `DecoratedTxOut` constructor is private. I considered other internal representations but I am tempted to stick with this

There are few smart constructors exported:

```
mkDecoratedTxOut :: Address -> Value -> Maybe (Either DatumHash Datum) -> Maybe Script -> DecoratedTxOut
mkDecoratedTxOutFromValidator :: Validator -> Value -> Either DatumHash Datum -> Maybe Script -> DecoratedTxOut
mkDecoratedTxOutFromValidatorHash :: ValidatorHash -> Value -> Either DatumHash Datum -> Maybe Script -> DecoratedTxOut
```

Also there's a function to turn a `DecoratedTxOut` into a `TxOut`. I don't like this function because the type offers no guarantee to be equivalent to a `TxOut` but for the moment we need it because `plutus-script-utils` is lower level than `plutus-ledger` and takes `TxOut`s.

```
decoratedTxOutToV1 :: DecoratedTxOut -> TxOut
```

`Plutus.ChainIndex.Types.ChainIndexTxOut` is left alone as a internal type for plutus-chain-index. The chain-index API talks to the exernal parties thorugh `DecoratedTxOut` as defined in `plutus-ledger`.

Pre-submit checklist:
- Branch
    - [ ] Tests are provided (if possible)
    - [ ] Commit sequence broadly makes sense
    - [ ] Key commits have useful messages
    - [ ] Formatting, PNG optimization, etc. are updated
- PR
    - [ ] Self-reviewed the diff
[...]

PR #799, opened 2022-11-02
PLT-619 Refactor and rename ChainIndexTxOut

berewt:
It mostly relies on the proposal of PR #626.
It provides type safety for script / pubkey txout (storing either the script hash or the pubkey), an extra getter to get the address from a txout, and smart constructors to build txout from an address.
In the refactoring of other modules, I relied as much as possible on lenses to avoid direct access to the partial field accessors (I'd like to avoid exposing them, but they are used in record syntax constructors in a few places, and it makes sense to use them in this context).

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


plutus-ledger/src/Ledger/Tx.hs
```diff
@@ -120,89 +134,116 @@ datumInDatumFromQuery _ DatumUnknown    = pure DatumUnknown
...
--- This datatype was created in order to be used in
--- 'Ledger.Constraints.processConstraint', specifically with the constraints
--- 'MustSpendPubKeyOutput' and 'MustSpendScriptOutput'.
-data ChainIndexTxOut =
-    PublicKeyChainIndexTxOut {
-      -- | Address of the transaction output. The address is protected by a
-      -- public key hash.
-      _ciTxOutAddress         :: Address,
+-- | Offchain view of a transaction output.
+data OffchainTxOut =
```
koslambrou, line 138: Sorry for not saying it earlier...

May I suggest the initial name of the previous PR? `DecoratedTxOut`? My reasoning:
There is already `Cardano.Api.TxOut` which is an offchain `TxOut`. But the current `OffchainTxOut` is not exactly that. It is an offchain `TxOut` but with extra stuff inside. Hence why a name like `DecoractedTxOut`, `TxOutWithExtras`, `EnrichedTxOut`, etc. might be better.

plutus-contract/test/Spec/Contract/TxConstraints.hs
```diff
@@ -0,0 +1,299 @@
+{-# LANGUAGE BlockArguments        #-}
```
koslambrou, line 1: Why was this file added? Seems like an old file that was added back.. Am I wrong?

berewt: I guess I didn't bay enough attention during the rebase (I probably changed something to it and git thought I wanted to keep it while James removed it).

plutus-ledger/src/Ledger/Tx.hs
berewt: Let's go with `Decorated` then. If I'm fortunate enough my `sed` skills should be sufficient to change it quickly.

plutus-ledger/src/Ledger/Tx.hs
```diff
@@ -120,89 +134,116 @@ datumInDatumFromQuery _ DatumUnknown    = pure DatumUnknown
...
+
+_offchainTxOutAddress :: OffchainTxOut -> Address
+_offchainTxOutAddress PublicKeyOffchainTxOut{_offchainTxOutPubKeyHash, _offchainTxOutStakingCredential} =
+    V1.Address (V1.PubKeyCredential _offchainTxOutPubKeyHash) _offchainTxOutStakingCredential
+_offchainTxOutAddress ScriptOffchainTxOut{_offchainTxOutValidatorHash, _offchainTxOutStakingCredential} =
+    V1.Address (V1.ScriptCredential _offchainTxOutValidatorHash) _offchainTxOutStakingCredential
+
+offchainTxOutAddress :: Getter OffchainTxOut Address
+offchainTxOutAddress = let
+    in to _offchainTxOutAddress
```
koslambrou, line 207: Remove "let in"?

PR #800, opened 2022-11-03
PLT-885 fix a bug in onchain MustPayToOtherScript

berewt:
We were making an exact comparison for non-ada values while we must have checked that the provided values of the tx cover the expectations of the constraint.

I added a test for partial spending of other tokens. 

While at it, I refactored the `MustPayToOtherScript` test suite to harmonise it with the other constraints suites.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-contract/test/Spec/Contract/Tx/Constraints/MustPayToOtherScript.hs
+++ b/plutus-contract/test/Spec/Contract/Tx/Constraints/MustPayToOtherScript.hs
@@ -14,6 +14,7 @@ import Control.Monad (void)
 import Test.Tasty (TestTree, testGroup)
 
 import Ledger qualified
+import Ledger qualified as PSU
 import Ledger.Ada qualified as Ada
 import Ledger.Constraints qualified as Constraints
 import Ledger.Constraints.OnChain.V1 qualified as Constraints
@@ -49,15 +50,15 @@ v1Tests :: SubmitTx -> TestTree
 v1Tests sub = testGroup "Plutus V1" $
    [ v1FeaturesTests
    , v2FeaturesNotAvailableTests
-   ] ?? sub ?? languageContextV1
+   ] ?? sub ?? PSU.PlutusV1
 
 v2Tests :: SubmitTx -> TestTree
 v2Tests sub = testGroup "Plutus V2" $
   [ v1FeaturesTests
   , v2FeaturesTests
-  ] ?? sub ?? languageContextV2
+  ] ?? sub ?? PSU.PlutusV2
 
-v1FeaturesTests :: SubmitTx -> LanguageContext -> TestTree
+v1FeaturesTests :: SubmitTx -> PSU.Language -> TestTree
 v1FeaturesTests sub t = testGroup "Plutus V1 features" $
     [ successfulUseOfMustPayToOtherScriptWithDatumInTxWithMintedToken
     , successfulUseOfMustPayToOtherScriptWithDatumInTxWhenOffchainIncludesTokenAndOnchainChecksOnlyToken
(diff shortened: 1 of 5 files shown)


plutus-contract/test/Spec/Contract/Tx/Constraints/MustPayToOtherScript.hs
```diff
@@ -49,32 +50,33 @@ v1Tests :: SubmitTx -> TestTree
...
-  ] ?? sub ?? languageContextV2
+  ] ?? sub ?? PSU.PlutusV2
 
-v1FeaturesTests :: SubmitTx -> LanguageContext -> TestTree
+v1FeaturesTests :: SubmitTx -> PSU.Language -> TestTree
 v1FeaturesTests sub t = testGroup "Plutus V1 features" $
     [ successfulUseOfMustPayToOtherScriptWithDatumInTxWithMintedToken
     , successfulUseOfMustPayToOtherScriptWithDatumInTxWhenOffchainIncludesTokenAndOnchainChecksOnlyToken
-    --, successfulUseOfMustPayToOtherScriptWithDatumInTxWhenOffchainIncludesTokenAndOnchainChecksOnlyAda -- FAILING when onchain checks for only ada value and token is present -- PLT-885
+    , successfulUseOfMustPayToOtherScriptWithDatumInTxWhenOffchainIncludesTokenAndOnchainChecksOnlyAda -- FAILING when onchain checks for only ada value and token is present -- PLT-885
```
koslambrou, line 65: Remove comment?

berewt: 🤦🏻

plutus-ledger-constraints/src/Ledger/Constraints/OnChain/V2.hs
```diff
@@ -153,12 +153,12 @@ checkTxConstraint ctx@ScriptContext{scriptContextTxInfo} = \case
                 -- provide datum.
                    Ada.fromValue txOutValue >= Ada.fromValue vl
                 && Ada.fromValue txOutValue <= Ada.fromValue vl + Ledger.maxMinAdaTxOut
-                && Value.noAdaValue txOutValue == Value.noAdaValue vl
+                && geq (Value.noAdaValue txOutValue) (Value.noAdaValue vl)
```
koslambrou, line 156: What is the logic for this instead of `==`?

berewt: The same as for ada values, we want to have at least the requested amounts of each requested values available.

koslambrou: Mmm... but for the Ada values the only reason we did that was because of the min Ada per utxo requirement. I thought this wouldn't be needed for non-Ada values. Maybe I need to check the test case to understand.

berewt:
@james-iohk Here's what's going on: `mustPayToOtherScript` was designed initially on-chain to check if the exact given value was paid to the script. **But** there was a workaround to deal with the minimum Ada of a transaction. Most of your tests were right because of this workaround, and fail if we spend a larger amount of ADA.
After a long discussion, we have decided to change the behaviour on-chain to check if the other script receive **at least** the given value and clarify it in the constraint documentation.

plutus-ledger-constraints/src/Ledger/Constraints/OnChain/V2.hs
```diff
@@ -152,13 +152,11 @@ checkTxConstraint ctx@ScriptContext{scriptContextTxInfo} = \case
...
                    Ada.fromValue txOutValue >= Ada.fromValue vl
-                && Ada.fromValue txOutValue <= Ada.fromValue vl + Ledger.maxMinAdaTxOut
-                && Value.noAdaValue txOutValue == Value.noAdaValue vl
+                && geq (Value.noAdaValue txOutValue) (Value.noAdaValue vl)
                 && txOutAddress == addr
             checkOutput (TxOutDatumInTx _) TxOut{txOutAddress, txOutValue, txOutDatum=OutputDatumHash h} =
                    Ada.fromValue txOutValue >= Ada.fromValue vl
-                && Ada.fromValue txOutValue <= Ada.fromValue vl + Ledger.maxMinAdaTxOut
-                && Value.noAdaValue txOutValue == Value.noAdaValue vl
+                && geq (Value.noAdaValue txOutValue) (Value.noAdaValue vl)
```
james-iohk, line 159: Why is this not also applied below for when tx uses inline datum?

plutus-ledger-constraints/src/Ledger/Constraints/OnChain/V2.hs
james-iohk: I had a similar thought. The most important thing is that MustMintValue is strict with the amount, which it is.

james-iohk (changes requested):

plutus-ledger-constraints/src/Ledger/Constraints/OnChain/V2.hs
berewt: As noticed during the review with James, the good thing is that the new semantic for `MustPayToOtherScript` onchain is now similar to the one of the `MusPayToPubKey` constraint.

james-iohk (approved):

koslambrou (approved):

PR #817, opened 2022-11-10
PLT-1219 Remove offchain logic for MustProduceAtLeast and MustSpendAtLeast

berewt:
Offchain use of `MustProduceAtLeast` and `MustSpendAtLeast` now checks if we have produced/spent at least the given amount when we create the Tx, and throw an error if it isn't the case. (see PLT-665 for the rationale behind this change)

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-contract/test/Spec/Contract/Tx/Constraints/MustProduceAtLeast.hs
+++ b/plutus-contract/test/Spec/Contract/Tx/Constraints/MustProduceAtLeast.hs
@@ -8,22 +8,24 @@
 {-# LANGUAGE TypeFamilies        #-}
 module Spec.Contract.Tx.Constraints.MustProduceAtLeast(tests) where
 
+import Control.Lens (_1, _head, has, makeClassyPrisms, only, (&))
 import Control.Monad (void)
 import Test.Tasty (TestTree, testGroup)
 
-import Data.Function ((&))
+import Data.Text qualified as Text
 import Ledger qualified
 import Ledger.Ada qualified as Ada
 import Ledger.CardanoWallet (paymentPrivateKey)
-import Ledger.Constraints.OffChain qualified as Constraints (MkTxError (OwnPubKeyMissing), ownPaymentPubKeyHash,
-                                                             typedValidatorLookups, unspentOutputs)
+import Ledger.Constraints.OffChain qualified as Constraints (MkTxError, ownPaymentPubKeyHash, typedValidatorLookups,
+                                                             unspentOutputs)
 import Ledger.Constraints.OnChain.V1 qualified as Constraints (checkScriptContext)
 import Ledger.Constraints.TxConstraints qualified as Constraints (collectFromTheScript, mustIncludeDatumInTx,
-                                                                  mustPayToTheScriptWithDatumInTx, mustProduceAtLeast)
+                                                                  mustPayToTheScriptWithDatumInTx,
+                                                                  mustPayWithDatumInTxToPubKey, mustProduceAtLeast)
 import Ledger.Generators (someTokenValue)
 import Ledger.Tx qualified as Tx
 import Ledger.Typed.Scripts qualified as Scripts
-import Plutus.Contract as Con (Contract, ContractError (ConstraintResolutionContractError, WalletContractError), Empty,
+import Plutus.Contract as Con (Contract, ContractError (WalletContractError), Empty, _ConstraintResolutionContractError,
                                awaitTxConfirmed, submitTxConstraintsWith, utxosAt)
 import Plutus.Contract.Test (assertContractError, assertFailedTransaction, assertValidatedTransactionCount,
                              changeInitialWalletValue, checkPredicateOptions, defaultCheckOptions,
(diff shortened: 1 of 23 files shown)


sjoerdvisscher (approved):
Given my `Address` PR, I'm all for removing `slOwnPaymentPubKeyHash` and `slOwnStakingCredential`.

berewt:
> Given my `Address` PR, I'm all for removing `slOwnPaymentPubKeyHash` and `slOwnStakingCredential`.

It's great to see synergies on apparently disconnected PR :)

koslambrou (approved):

plutus-contract/test/Spec/Contract/Tx/Constraints/MustProduceAtLeast.hs
```diff
@@ -79,10 +90,9 @@ mustProduceAtLeastContract offAmt onAmt baseScriptValue pkh = do
     let lookups2 = Constraints.typedValidatorLookups typedValidator
             <> Constraints.unspentOutputs pubKeyUtxos
             <> Constraints.unspentOutputs scriptUtxos
-            <> Constraints.ownPaymentPubKeyHash pkh
         tx2 =
             Constraints.collectFromTheScript scriptUtxos ()
-            <> Constraints.mustIncludeDatumInTx (Datum $ PlutusTx.toBuiltinData onAmt)
+            <> Constraints.mustPayWithDatumInTxToPubKey w1PaymentPubKeyHash (Datum $ PlutusTx.toBuiltinData onAmt) offAmt
```
sjoerdvisscher, line 95: Let's do everything with `Address` from now on (unless you're testing one of the pubkey functions specifically).

plutus-use-cases/src/Plutus/Contracts/GameStateMachine.hs
```diff
@@ -108,6 +108,8 @@ data GuessArgs =
     GuessArgs
         { guessArgsGameParam     :: GameParam
         -- ^ The parameters for parameterizing the validator.
+        , guessTokenTarget       :: PaymentPubKeyHash
```
sjoerdvisscher, line 111: This should be `Address`

plutus-use-cases/src/Plutus/Contracts/Governance.hs
```diff
@@ -92,8 +92,8 @@ data GovState = GovState
 
 data GovInput
     = MintTokens [TokenName]
-    | ProposeChange Proposal
-    | AddVote TokenName Bool
+    | ProposeChange PaymentPubKeyHash Proposal
```
sjoerdvisscher, line 95: These should be `Address` too.

plutus-use-cases/src/Plutus/Contracts/Governance.hs
```diff
@@ -175,19 +175,19 @@ transition Params{..} State{ stateData = s, stateValue} i = case (s, i) of
                 (zip initialHolders tokenNames)
         in Just (constraints <> Constraints.mustMintValue total, State s stateValue)
 
-    (GovState law mph Nothing, ProposeChange proposal@Proposal{tokenName}) ->
-        let constraints = ownsVotingToken mph tokenName
+    (GovState law mph Nothing, ProposeChange owner proposal@Proposal{tokenName}) ->
+        let constraints = ownsVotingToken owner mph tokenName
```
sjoerdvisscher, line 179: Instead of adding an `owner` parameter you should use the `ownAddress` function. I'm not even sure it is secure this way since you can pass a different value than your own address.

sjoerdvisscher (changes requested):

plutus-ledger-constraints/src/Ledger/Constraints/OffChain.hs
```diff
@@ -573,33 +581,6 @@ adjustTxOut params txOut = do
...
-addMissingValueSpent = do
-    missing <- gets totalMissingValue
-    if Value.isZero missing
-        then pure ()
-        else do
-            -- add 'missing' to the transaction's outputs. This ensures that the
-            -- wallet will add a corresponding input when balancing the
-            -- transaction.
-            -- Step 4 of the process described in [Balance of value spent]
-            pkh <- asks slOwnPaymentPubKeyHash >>= maybe (throwError OwnPubKeyMissing) pure
```
sjoerdvisscher, line 593: I think this means we can remove `OwnPubKeyMissing`?

berewt: I think so, I'm doing it.

plutus-use-cases/src/Plutus/Contracts/Governance.hs
berewt: We're out of the contract monad at this stage though, so `ownAddress` won't work. From a security perspective, the `owner` address is the one where we'll send the vote token back, we still use the current wallet to initiate the transaction, so it should be ok.

sjoerdvisscher: Oh, I get it now. It's it only for specifying where the token should go to, which `mustSpendAtLeast` didn't check either.

sjoerdvisscher (approved):

PR #825, opened 2022-11-18
Use the new indexer api in the existing indexers.

raduom:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi/src/Marconi/Index/ScriptTx.hs
+++ b/marconi/src/Marconi/Index/ScriptTx.hs
@@ -17,7 +17,7 @@ import Database.SQLite.Simple.FromField qualified as SQL
 import Database.SQLite.Simple.ToField qualified as SQL
 import GHC.Generics (Generic)
 
-import Cardano.Api (SlotNo)
+import Cardano.Api (SlotNo (SlotNo))
 import Cardano.Api qualified as C
 import Cardano.Api.Shelley qualified as Shelley
 -- TODO Remove the following dependencies (and also cardano-ledger-*
@@ -49,6 +49,7 @@ newtype TxCbor = TxCbor BS.ByteString
 data ScriptTxRow = ScriptTxRow
   { scriptAddress :: !ScriptAddress
   , txCbor        :: !TxCbor
+  , txSlot        :: !SlotNo
   } deriving (Generic)
 
 instance SQL.ToField ScriptAddress where
@@ -60,9 +61,15 @@ instance SQL.FromField ScriptAddress where
       (\b -> maybe cantDeserialise (return . ScriptAddress) $ Shelley.deserialiseFromRawBytes Shelley.AsScriptHash b) . deserialiseOrFail
     where
       cantDeserialise = SQL.returnError SQL.ConversionFailed f "Cannot deserialise address."
+instance SQL.ToField SlotNo where
+  toField (SlotNo n) =  SQL.toField $ (fromIntegral n :: Int)
+instance SQL.FromField SlotNo where
+  fromField f = SlotNo <$> SQL.fromField f
 
 instance SQL.ToRow ScriptTxRow where
-  toRow o = [SQL.toField $ scriptAddress o, SQL.toField $ txCbor o]
+  toRow o = [ SQL.toField $ scriptAddress o
+            , SQL.toField $ txCbor o
+            , SQL.toField $ txSlot o ]
 
 -- * Indexer
 
@@ -100,7 +107,7 @@ open :: (ScriptTxIndex -> ScriptTxUpdate -> IO [()]) -> FilePath -> Depth -> IO 
 open onInsert dbPath (Depth k) = do
   ix <- fromJust <$> Ix.newBoxed query store onInsert k ((k + 1) * 2) dbPath
   let c = ix ^. Ix.handle
-  SQL.execute_ c "CREATE TABLE IF NOT EXISTS script_transactions (scriptAddress TEXT NOT NULL, txCbor BLOB NOT NULL)"
+  SQL.execute_ c "CREATE TABLE IF NOT EXISTS script_transactions (scriptAddress TEXT NOT NULL, txCbor BLOB NOT NULL, slotNo INT NOT NULL)"
   SQL.execute_ c "CREATE INDEX IF NOT EXISTS script_address ON script_transactions (scriptAddress)"
   pure ix
 
@@ -109,13 +116,13 @@ open onInsert dbPath (Depth k) = do
     store ix = do
       buffered <- Ix.getBuffer $ ix ^. Ix.storage
       let rows = do
-            ScriptTxUpdate txScriptAddrs _slotNo <- buffered
+            ScriptTxUpdate txScriptAddrs slotNo <- buffered
             (txCbor', scriptAddrs) <- txScriptAddrs
             scriptAddr <- scriptAddrs
-            pure $ ScriptTxRow scriptAddr txCbor'
+            pure $ ScriptTxRow scriptAddr txCbor' slotNo
       SQL.execute_ (ix ^. Ix.handle) "BEGIN"
       forM_ rows $
-        SQL.execute (ix ^. Ix.handle) "INSERT INTO script_transactions (scriptAddress, txCbor) VALUES (?, ?)"
+        SQL.execute (ix ^. Ix.handle) "INSERT INTO script_transactions (scriptAddress, txCbor, slotNo) VALUES (?, ?, ?)"
       SQL.execute_ (ix ^. Ix.handle) "COMMIT"
 
 query :: ScriptTxIndex -> Query -> [ScriptTxUpdate] -> IO Result
--- a/marconi/src/Marconi/Index/ScriptTx.hs
+++ b/marconi/src/Marconi/Index/ScriptTx.hs
@@ -4,14 +4,13 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
 
 module Marconi.Index.ScriptTx where
 
-import Codec.Serialise (deserialiseOrFail)
-import Control.Lens.Operators ((^.))
 import Data.ByteString qualified as BS
-import Data.Foldable (forM_, toList)
-import Data.Maybe (catMaybes, fromJust)
+import Data.Foldable (foldl', forM_, toList)
+import Data.Maybe (catMaybes)
 import Database.SQLite.Simple qualified as SQL
 import Database.SQLite.Simple.FromField qualified as SQL
 import Database.SQLite.Simple.ToField qualified as SQL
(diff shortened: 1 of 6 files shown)


marconi/src/Marconi/Index/New/ScriptTx.hs
```diff
@@ -0,0 +1,418 @@
...
+   The first type we introduce is the monad in which the database (and by extension,
+   the indexer) runs. -}
+
+type instance StorableMonad ScriptTxHandle = IO
+
+{- The next type we introduce is the type of events. Events are the data atoms that
+   the indexer consumes. They depend on the `handle` because they need to eventually
+   be persisted in the database, so the database has to be able to accomodate them.
+
+   THe original implementation used two spearate data structures for storing data
```
koslambrou, line 78: THe -> The

marconi/src/Marconi/Index/New/ScriptTx.hs
```diff
@@ -0,0 +1,418 @@
...
+data instance StorableEvent ScriptTxHandle = ScriptTxEvent
+  { txScripts :: [(TxCbor, [StorableQuery ScriptTxHandle])]
+  , slotNo    :: !SlotNo
+  } deriving (Show)
+
+{- The resume and query functionality requires a way to specify points on the chain
+   from which we want to resume, or points up to which we want to query. Next we
+   define the types of these points. -}
+
+type instance StorablePoint ScriptTxHandle = SlotNo
```
koslambrou, line 93: Curious: isn't it boilerplate because the resume/query point is **always** of type `SlotNo`? Is there reason to have another datatype? (Other than maybe the fact someone might use different types than the ones from `cardano-api`)

marconi/src/Marconi/Index/New/ScriptTx.hs
```diff
@@ -0,0 +1,418 @@
...
+            Shelley.ScriptInEra _ script' -> Just $ C.hashScript script'
+      _ -> [] -- Byron transactions have no scripts
+    hashes = catMaybes hashesMaybe :: [Shelley.ScriptHash]
+  in map ScriptTxAddress hashes
+
+getTxScripts :: forall era . C.Tx era -> [StorableQuery ScriptTxHandle]
+getTxScripts (C.Tx txBody _ws) = getTxBodyScripts txBody
+
+{- Now that all connected data types have been defined, we go on to implement some
+   of the type classes required for information storage and retreival. -}
```
koslambrou, line 181: retreival -> retrieval

marconi/src/Marconi/Index/New/ScriptTx.hs
```diff
@@ -0,0 +1,418 @@
...
+
+  {- We want to potentially store data in two formats. The first one is similar (if
+     not identical) to the format of data stored in memory; it should contain information
+     that allows knowing at which point the data was generated.
+
+     We use this first format to support rollbacks for disk data. The second format,
+     which is not always necessary and does not have any predetermined structure,
+     should be thought of as an aggregate of the previously produced events.
+
+     For this indexer we don't really need an aggregate, so our "aggregate" has almost the same
```
koslambrou, line 218: If we did need an aggregate, where would we define it?

marconi/src/Marconi/Index/New/ScriptTx.hs
```diff
@@ -0,0 +1,418 @@
...
+     structure as the in-memory data. We pretend that there is an aggregate by
+     segregating the data into two sections, by using the `hdlDiskStore` parameter. We
```
koslambrou, line 220: The `hdlDiskStore` doesn't seem to be used anywhere.

marconi/src/Marconi/Index/New/ScriptTx.hs
raduom: A separate database table with whatever structure is efficient for your queries.

raduom: I can talk about an example when implementing a UTXOs aggregate. Do you think that would be a good idea?

marconi/src/Marconi/Index/New/ScriptTx.hs
raduom: It is used here: https://github.com/input-output-hk/plutus-apps/pull/825/files#diff-b167f09d5ef11e569397b644012c7dda8a65fd0e7b74dd34bb54b41456eb8eb2R229

marconi/src/Marconi/Index/New/ScriptTx.hs
koslambrou: Thanks :)

marconi/src/Marconi/Index/New/ScriptTx.hs
koslambrou: I think the best would be to actually write an indexer that implements a simple aggregate. For example, an indexer which aggregates the `Value` for each `Address` (like in an accounting model). I could create a story for this?

marconi/src/Marconi/Index/New/ScriptTx.hs
raduom: This is an example of why I don't think that `newtype`-ing over a SQLConnection to make a handle is such a bad thing. I would like to believe that we will need specific configurations for most indexers.

marconi/src/Marconi/Index/New/ScriptTx.hs
raduom: Also, think about side-chains (midnight). Or proprietary data streams coming from centralised exchanges (there was a project for building a dex aggregator that I am thinking about). Maybe one would want to correlate them somehow in a hybrid measure of time. In that case we may see very different types of points.

marconi/test/Integration.hs
```diff
@@ -316,7 +316,11 @@ testIndex = H.integration . HE.runFinallies . workspace "chairman" $ \tempAbsBas
     let
       queryLoop n = do
         H.threadDelay 250_000 -- wait 250ms before querying
-        txCbors <- liftIO $ ScriptTx.query indexer (ScriptTx.ScriptAddress plutusScriptHash) []
+        -- With the new indexer, this should pass right away (by finding the in-memory
```
eyeinsky, line 319: If the loop isn't required anymore, then just remove it!

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -62,12 +62,14 @@ import GHC.Generics (Generic)
...
 data family StorableEvent h
 
-data family StorablePoint h
+type family StorablePoint h
 
 data family StorableQuery h
 
 data family StorableResult h
 
+type family StorableMonad h :: * -> *
```
eyeinsky, line 71: Why `StorableMonad`, can the monad be stored?

If this is a type family tied to the type classes below, why not make it an associated type family? Or, will it ever be used separately?

koslambrou (approved):
There's some missing tests that would make sure the SQL queries work correctly (for stuff stored on disk). However, the main goal of this PR was to actually show how to use this new indexer (which I believe it succeeds). Awesome documentation @raduom, really helped me.

So I'm happy merging this. Then we can start focusing on updating the other more important indexers with the new API in the near future.

raduom:
> There's some missing tests that would make sure the SQL queries work correctly (for stuff stored on disk). However, the main goal of this PR was to actually show how to use this new indexer (which I believe it succeeds). Awesome documentation @raduom, really helped me.
> 
> So I'm happy merging this. Then we can start focusing on updating the other more important indexers with the new API in the near future.

Actually, it does test the SQL queries (see: https://github.com/input-output-hk/plutus-apps/pull/825/files#diff-bf8a4e4a83265fb689408cb94064d62a73daa8bf72e63c3652689c078d327688R319-R323), so those are also probably working well.

Could we do a better work testing this? Yes, I think we could. I think Kayvan has a good idea on how we could do that (which is the way he tested the indexer he wrote).

rewindable-index/src/RewindableIndex/Storable.hs
raduom: It is the monad in which the storable operations run (in this case IO). It is also one monad that is used across all 4 type classes so I would rather define it once.

marconi/test/Integration.hs
raduom: It's not that it is not required. It is required if we want to check the data that was stored in the database. I don't think it's a bad idea to do it like this. @koslambrou is also interested in testing that the SQL queries work properly, so I think this is a good thing to have.

eyeinsky:
Yes, it does test the SQL query, it's just that there is nothing to wait upon to know when something is persisted to disk.

@kayvank's good idea should be moved into the indexer framework and used such that poll loops were not required, that's what I'm trying to say. :)

raduom:
> Yes, it does test the SQL query, it's just that there is nothing to wait upon to know when something is persisted to disk.
> 
> @kayvank's good idea should be moved into the indexer framework and used such that poll loops were not required, that's what I'm trying to say. :)

What @kayvank did was randomised testing of the indexer, without connecting it to the test net. Because it's not connected to the test net you can run a lot of tests.

If we take K to be the number of buffered events and K*2 the number of on-disk events, followed by an aggregate, then by generating between (K*3 + 1) - K*N (where N>3) events you will cover all possible data locations (in-memory, on disk as events and on disk as aggregate).

If you generate this data as instructed, then a good property to test is that whatever goes in, comes out. This will test your indexer's storage and query code. If we also sparkle some rollbacks, we also test ... rollbacks. And we can go from there.

However, since each indexer has its own storage schema and queries, what we can do is at most provide testing guidelines. I am not sure it would be worth the effort to automate this.

raduom:
> I think we are talking about two different things (I don't know what the property test is what you are talking about). What Kayvan also did was make the query work such that both the in-memory and on-disk data can be queried. The reason for the poll-loop is that the indexer is built in such a way that the in-memory part is inaccessible for any of the actual indexers we have. If the query had both available, then polling wouldn't be required.

Yes. What I wrote before covers that problem without making available the observation of what is on disk and what is in memory while testing both.

eyeinsky:
Right, so for an integration test, the in-memory part isn't accessible for the query?

marconi/src/Marconi/Index/ScriptTx.hs
```diff
@@ -32,110 +31,265 @@ import Cardano.Ledger.Keys qualified as LedgerShelley
...
+   type, meant as a wrapper for the connection type (in this case the SQLite
+   connection).
 
-newtype Depth = Depth Int
+   However this is a very good place to add some more configurations
+   that the indexer may require (for example performance tuning settings). In our
+   case we add the number of events that we want to return from the on-disk buffer -}
+
+data ScriptTxHandle = ScriptTxHandle
+  { hdlConnection :: SQL.Connection
```
kayvank, line 49: Totally like that we create the connections here rather than just passing in the filepath. thnx

marconi/src/Marconi/Index/ScriptTx.hs
```diff
@@ -32,110 +31,265 @@ import Cardano.Ledger.Keys qualified as LedgerShelley
...
+  persistToStorage
+    :: Foldable f
+    => f (StorableEvent ScriptTxHandle)
+    -> ScriptTxHandle
+    -> IO ScriptTxHandle
+  persistToStorage es h = do
+    let rows = foldl' (\ea e -> ea ++ flatten e) [] es
+        c    = hdlConnection h
+    SQL.execute_ c "BEGIN"
+    forM_ rows $
```
kayvank, line 186: nit-pic
use [executeMany](https://hackage.haskell.org/package/sqlite-simple-0.4.18.2/docs/Database-SQLite-Simple.html#v:executeMany) to take advantage of SQL batching?

kayvank (approved):

PR #826, opened 2022-11-18
PLT-1040: fix currentChainIndexSlot

sjoerdvisscher:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-chain-index-core/src/Plutus/ChainIndex/Emulator/Handlers.hs
+++ b/plutus-chain-index-core/src/Plutus/ChainIndex/Emulator/Handlers.hs
@@ -278,16 +278,14 @@ appendBlocks [] = pure ()
 appendBlocks blocks = do
     let
         processBlock (utxoIndexState, txs) (Block tip_ transactions) = do
-            if null transactions then return (utxoIndexState, txs)
-            else do
-                case UtxoState.insert (TxUtxoBalance.fromBlock tip_ (map fst transactions)) utxoIndexState of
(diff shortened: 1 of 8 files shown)


plutus-chain-index-core/src/Plutus/ChainIndex/Handlers.hs
```diff
@@ -398,9 +399,16 @@ appendBlocks ::
     , Member (BeamEffect Sqlite) effs
     , Member (LogMsg ChainIndexLog) effs
     )
-    => [ChainSyncBlock] -> Eff effs ()
-appendBlocks [] = pure ()
-appendBlocks blocks = do
+    => L.Slot -> [ChainSyncBlock] -> Eff effs ()
+appendBlocks newSlot [] = do
```
berewt, line 403: Is `appendBlocks :: Either L.Slot (List1 ChainSyncBlock) -> Eff effs ()` an overkill here? It would avoid expecting that the slot is taken into account when some blocks are passed.

berewt (approved):

plutus-chain-index-core/src/Plutus/ChainIndex/Handlers.hs
sjoerdvisscher: Thanks for prodding. After looking at it again the real fix was somewhere else.

kayvank (approved):
LGTM

PR #838, opened 2022-11-24
Add script to log marconi cpu, mem, sync%

james-iohk:
You may find  it useful, don't need to merge if we don't want to.


koslambrou:
I'm fine with merging it :) Seems pretty useful and we can definitely upgrade it later if we need to.
Thanks James!

koslambrou (approved):

raduom (approved):
LGTM

koslambrou:
@james-iohk Do you want to fix the warnings so that we can merge this?

PR #842, opened 2022-11-29
mkUntyped* functions work with PlutusV1 and PlutusV2

berewt:
I chose to implement it as a typeclass implemented by `ScriptContext`s as it doesn't require manipulating an extra type to set the Plutus version and it doesn't require complex type machinery to work (we need to assert `ScriptContext PlutusV1Script ~ PV1.ScriptContext` in some places to use `mkUntyped` functions along with functions that manipulate `ScriptContext`).

The drawback though, is that we often need to specify explicitly the type of context we use to determine the typeclass and the order is different than the one of the parameters, as the type determining the instance needs to be provided first.
A possible mitigation is to provide explicitly the signature of the script function.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-example/src/PlutusExample/PlutusVersion1/MintingScript.hs
+++ b/plutus-example/src/PlutusExample/PlutusVersion1/MintingScript.hs
@@ -19,9 +19,8 @@ import Data.ByteString.Lazy qualified as LB
 import Data.ByteString.Short qualified as SBS
 
 import Plutus.Script.Utils.Typed qualified as Scripts
-import Plutus.Script.Utils.V1.Typed.Scripts qualified as Scripts
-import Plutus.V1.Ledger.Contexts (ScriptContext)
-import Plutus.V1.Ledger.Scripts (Script, Validator (Validator), mkMintingPolicyScript, unMintingPolicyScript)
+import Plutus.V1.Ledger.Api (MintingPolicy, Script, ScriptContext, Validator (Validator), mkMintingPolicyScript,
+                             unMintingPolicyScript)
 import PlutusTx qualified
 import PlutusTx.Prelude hiding (Semigroup (..), unless, (.))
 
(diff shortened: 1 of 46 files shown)


plutus-example/src/PlutusExample/PlutusVersion1/MintingScript.hs
```diff
@@ -18,6 +18,7 @@ import Codec.Serialise
 import Data.ByteString.Lazy qualified as LB
 import Data.ByteString.Short qualified as SBS
 
+import Plutus.Script.Utils.Typed qualified as Scripts
 import Plutus.Script.Utils.V1.Typed.Scripts qualified as Scripts
```
sjoerdvisscher, line 22: In some places you have removed the V1 import and in other places not. Is there a reason for that?

plutus-script-utils/src/Plutus/Script/Utils/Typed.hs
```diff
@@ -110,3 +119,129 @@ vForwardingMintingPolicy = tvForwardingMPS
 --   validator
 forwardingMintingPolicyHash :: TypedValidator a -> PV1.MintingPolicyHash
 forwardingMintingPolicyHash = tvForwardingMPSHash
+
+class PV1.UnsafeFromData a => ScriptContext a where
```
sjoerdvisscher, line 123: `IsScriptContext` perhaps? And I wouldn't use `a` here, that's more for type variables that can be (almost) anything.

sjoerdvisscher (approved):
The only thing I worry about is that a future plutus version might use the same `ScriptContext` type, but does need different `mkUntyped*` functions. But at first glance that seems unlikely and we can deal with that when it happens.

plutus-example/src/PlutusExample/PlutusVersion1/MintingScript.hs
berewt: I removed it when I encounter the `Unused` warning, I'm gonna check consistency across the examples.

berewt:
> The only thing I worry about is that a future plutus version might use the same `ScriptContext` type, but does need different `mkUntyped*` functions. But at first glance that seems unlikely and we can deal with that when it happens.

Yes, I wondered about it and came to the same conclusion, it would be hard to reuse existing functions if the signature changed anyhow.

plutus-example/src/PlutusExample/PlutusVersion1/MintingScript.hs
berewt: Ok, it was because some import Plutus.Vx.Ledger.Api and others don't, relying on PSU.

koslambrou (approved):

PR #847, opened 2022-12-03
Doctests configuration for plutus-ledger

koslambrou:
Was able to get doctests working following the example here: https://github.com/input-output-hk/haskell.nix/tree/master/test/cabal-doctests

This PR only activates it for `plutus-ledger`.

The only thing I need to verify is that we don't need to add dependencies in plutus-ledger lib for modules only used in doctests.

EDIT:

Cool! Using dependencies like `generic-arbitrary`  don't need to be added to the `cardano-ledger` lib, only in the `doctests` test-suite.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-ledger/Setup.hs
+++ b/plutus-ledger/Setup.hs
@@ -1,0 +1,6 @@
+module Main where
+
+import Distribution.Extra.Doctest (defaultMainWithDoctests)
+
+main :: IO ()
+main = defaultMainWithDoctests "doctests"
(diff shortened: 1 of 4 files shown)


koslambrou:
Oh, just noticed that `cabal-doctest` is deprecated :(. However, it still works. Using `doctest` directly instead doesn't work out of the box. Maybe I should drop this PR and create a ticket for `doctest`?

michaelpj:
Personally, I just don't think doctests are worth it any more. Yes, they're somewhat nice, but you can't demonstrate most complex things in them, and they're such a pain to get running :(

But if this works then fine!

PR #851, opened 2022-12-06
Standardisation - Part 1 (docs + shell)

zeme-iohk:
Nix files that begin with `TODO(std) DUP` are files that have been copy-pasted from plutus. 
They most likely don't need to be reviewed. 
Once plutus-apps has been standardised, I'll look at a way to reference those files using the plutus flake instead.
Hopefully this will eventually remove all duplication between the two repos


__std__/cells/plutus-apps/devshells/plutus-apps-shell.nix
```diff
@@ -0,0 +1,230 @@
...
+    pkgs.awscli2
+    pkgs.bzip2
+
+    pkgs.zlib
+    pkgs.cacert
+  ];
+
+  devshell.startup."pre-commit-check".text = cell.packages.pre-commit-check.shellHook;
+
+  # TODO(std) maybe this isn't needed in the devshell, check later
```
michaelpj, line 132: It's definitely not needed, we're using a newer nix that doesn't need this.

michaelpj (approved):
looks plausible

Issue #854, opened 2022-12-08
Coverage Report Static Analysis Issue


Ali-Hill:
### Summary

When producing the coverage report for the Governance contract it is reported that the nothing case in the state machine is uncovered. This is intentionally uncovered so I believe it should be considered as ignored by static analysis rather than uncovered. 

### Steps to reproduce the behavior

Download repository: https://github.com/Ali-Hill/plutus-apps/tree/new-main

From the root of the repository run the following:

nix-shell
cabal repl plutus-use-cases-test
import Spec.Governance
check_propGovernanceWithCoverage

This produces a coverage report it and stores it as Governance.html in the plutus-use-cases directory. 

### Actual Result

The nothing case in line 207 is reported as uncovered. 
![Screenshot 2022-12-08 at 15 14 39](https://user-images.githubusercontent.com/11505148/206483365-e1e38a57-6030-4fb1-b5e0-e1bf05a06fb0.png)

### Expected Result

The nothing case in line 207 is ignored. 

### System info

OS: Mac
https://github.com/Ali-Hill/plutus-apps/commit/f3571c16a10dd574f0314d17f614171ebd1a59f5

MaximilianAlgehed:
The issue is that you're using `getCovIdx` when you should be using `computeRefinedCoverageIndex`. I don't know that the design right now is entirely optimal, but I'm also not sure that you want the analysis to be the default... Anyway, that should solve your problem :)

PR #860, opened 2022-12-12
Fix partiality bug when dealing with empty datatypes in static analysis

MaximilianAlgehed:
This fixes a bug identified by Alasdair Hill.


berewt:
Is it #854?

PR #867, opened 2022-12-15
Set darwin temp dir via TMPDIR env and use the original `workspace` from hedgehog-extras

eyeinsky:
- On macos there was an issue with paths being too long, so the `workspace` function sets temp dir to `/tmp` manually.
- Our CI macos'es run tests as several users, so the `workspace` function [needed to be modified](https://github.com/input-output-hk/hedgehog-extras/pull/18)
- To re-use the modification, use the `TMPDIR` env variable to set temp dir to `/tmp` instead. (This fix should have been used from the beginning)

Pre-submit checklist:
- Branch
- PR
    - [X] Self-reviewed the diff
    - [ ] Reviewer requested


koslambrou (approved):
Don't have a Mac so I can't test this, but looks good I think.

PR #885, opened 2023-01-01
Address-Utxo indexer Use new marconi API

kayvank:
PLT-1314, Use the new rewindable indexes for Marconi Utxos.

**DRAFT, DO NOT MERGE pending mamba related changes**

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


marconi/test/Spec/Utxo.hs
```diff
@@ -54,31 +170,40 @@ txToUtxoTest = property $ do
...
+    (void . liftIO . Storable.insertMany events)  indexer
+    (fromStorage :: [StorableEvent Utxo.UtxoHandle]) <- (liftIO . Storable.getEvents) indexer
+      -- liftIO . traverse (Storable.query Storable.QEverything indexer) $ qs
+    (fromMemory :: [StorableEvent Utxo.UtxoHandle]) <- liftIO $ (Storable.getMemoryEvents (indexer ^. Storable.storage )) & V.freeze <&> V.toList
+    -- let (rows :: [Utxo.UtxoRow]) =
+    --       concatMap (\Utxo.UtxoResult  rs -> rs) results
+    --     ces = Utxo.rowsToEvents rows
+    --     computedEvents =
+
+  -- TODO bad tests, remove once we resolve the bug
```
kayvank, line 193: @raduom this tests fails. The failure may be to my misunderstanding of the Rewindable API.
 It seems as if the getMemoryEvents may have a bug, or I am not using it correctly.  Here are my observation:
- for small number of events, the Storable.getEvents  returns no events.
- When I increase the number of events per test, the Storable.getEvents returns some events, less than expected
- The bug seem to be with Storable.getMemoryEvents 
- for sufficiently large number of events, events are flushed to DB
Strangely, I can see events being written to DB when I increase the number of events per test, however, the getEvents doesnot report
- The print statements are there to drive the point in the discrepancies.  I'll remove them once the issue is resolve.

marconi/src/Marconi/Index/Utxo.hs
```diff
@@ -239,227 +268,373 @@ instance FromField C.BlockNo where
...
+             FROM unspent_transactions u
+             LEFT JOIN spent s ON
+                    u.txId = s.txInTxId
+                AND u.txId IS NOT NULL
+                AND u.txIx = s.txInTxIx
+             WHERE u.address = ?
+             ORDER BY u.slotNo ASC|] (Only q)
+        QInterval _ (C.ChainPoint sn _) -> SQL.query c
+          [r|SELECT u.address, u.txId, u.txIx , u.datum , u.datumHash, u.value
+                  , u.inLineScript, u.inLineScriptHash, u.blockNo, u.slotNo
```
raduom, line 494: Isn't this missing `blockHash`? Do you think we should use a generic query template and add the `slotNo` condition if necessary?

marconi/src/Marconi/Index/Utxo.hs
```diff
@@ -239,227 +268,373 @@ instance FromField C.BlockNo where
...
+             WHERE
+                    u.slotNo <= ?
+                AND u.txId IS NOT NULL
+                AND u.address = ?
+             ORDER BY u.slotNo ASC|] (sn, q)
+        QInterval _ (C.ChainPointAtGenesis) -> pure []
+      asEs <- rowsToEvents (getTxIns c) persisted
+
+      let updates = filterWithQueryInterval qi (asEs <> toList es)
+      pure . UtxoResult $ filterByAddress q updates where
```
raduom, line 508: I don't think we need to `filterByAddress` since we already did that int he `SQL`.

marconi/src/Marconi/Index/Utxo.hs
```diff
@@ -239,227 +268,373 @@ instance FromField C.BlockNo where
...
+              C.PlutusScriptLanguage (C.PlutusScriptV1)) script)  ->
+            (Just . C.serialiseToCBOR $ script, Just . C.hashScript $ script)
+          Shelley.ReferenceScript _  (
+            Shelley.ScriptInAnyLang  (
+              C.PlutusScriptLanguage (C.PlutusScriptV2)) script)  ->
+            (Just . C.serialiseToCBOR $ script, Just . C.hashScript $ script)
+       in
+        Utxo {..}
+{-
+ TODO we need to remove spent from in-memory utxoEvents.
```
raduom, line 596: We should also look into this scenario, but for now I don't see where this function is used. I don't think we should ever remove spent transactions from memory.

marconi/src/Marconi/Index/Utxo.hs
```diff
@@ -239,227 +268,373 @@ instance FromField C.BlockNo where
...
-        C.ScriptInvalid -> case txInsCollateral of
-                                C.TxInsCollateralNone     -> []
-                                C.TxInsCollateral _ txins -> txins
-  in Set.fromList inputs
+  -> C.BlockNo
+  -> C.ChainPoint
+  -> Maybe (StorableEvent UtxoHandle) -- ^ UtxoEvents are stored in storage after conversion to UtxoRow
+getUtxoEvents maybeTargetAddresses txs blkNo cp =
+  let
+    utxos = Set.fromList(concat . fmap (getUtxos maybeTargetAddresses) $ txs)
```
raduom, line 618: Is the last `$` required?

marconi/src/Marconi/Indexers.hs
```diff
@@ -117,29 +117,31 @@ isInTargetTxOut targetAddresses (C.TxOut address _ _ _) = case address of
     _                                                -> False
 
 utxoWorker
-    :: (Utxo.UtxoIndex -> IO Utxo.UtxoIndex)    -- ^ CPS function used in the queryApi thread, needs to be non-blocking
-    -> Maybe TargetAddresses                    -- ^ Target addresses to filter for
-    -> Worker
-utxoWorker indexerCallback maybeTargetAddresses Coordinator{_barrier} ch path =
-    Utxo.open path (Utxo.Depth 2160) >>= indexerCallback >>= innerLoop
+  :: (Storable.StorableEvent Utxo.UtxoHandle -> IO [()])
+  -> (Utxo.UtxoIndexer -> IO Utxo.UtxoIndexer)  -- ^ CPS function used in the queryApi thread, needs to be non-blocking
```
raduom, line 121: I think we can do better here. My dream would be to have `start....` return a `TVar UtxoIndex` so we can generically call whatever we want whenever we want. I am not sure my idea is better than a callback though. What do you think?

marconi/src/Marconi/Indexers.hs
```diff
@@ -117,29 +117,31 @@ isInTargetTxOut targetAddresses (C.TxOut address _ _ _) = case address of
...
-      void $ indexerCallback index -- refresh the query STM/CPS with new storage pointers/counters state
       event <- atomically $ readTChan ch
       case event of
-        RollForward (BlockInMode (Block (BlockHeader slotNo _ blkNo) txs) _) _ct ->
-            case (Utxo.getUtxoEvents maybeTargetAddresses slotNo blkNo txs) of
-                  Just us ->  Ix.insert (us) index  >>= innerLoop
-                  _       -> innerLoop index
+        RollForward (BlockInMode (Block (BlockHeader slotNo hsh blkNo) txs) _) _ct ->
+            case (Utxo.getUtxoEvents maybeTargetAddresses txs blkNo (C.ChainPoint slotNo hsh)) of
+              Just u -> void $ onInsert u -- TODO can we use this insead of the callback?
```
raduom, line 134: I have another (better?) idea, written before.

marconi/test/Spec/Utxo.hs
```diff
@@ -22,27 +30,135 @@ import Cardano.Api qualified as C
 import Gen.Cardano.Api.Typed qualified as CGen
 import Marconi.Index.Utxo qualified as Utxo
 import Marconi.Types (CurrentEra, TargetAddresses)
-import RewindableIndex.Index.VSplit qualified as Ix
+-- import RewindableIndex.Storable (Buffered (getStoredEvents, persistToStorage), HasPoint (getPoint),
```
raduom, line 33: Erase EVERY trace :)

marconi/test/Spec/Utxo.hs
```diff
@@ -22,27 +30,135 @@ import Cardano.Api qualified as C
...
-addressFilteredEventsTest :: Property
-addressFilteredEventsTest = property $ do
+deriving instance Eq (StorableEvent Utxo.UtxoHandle)
+deriving instance Ord (StorableEvent Utxo.UtxoHandle)
+
+
+equivalentSets :: Eq a => Set a -> Set a -> Bool
+equivalentSets us us' = equivalentLists (Set.toList us) (Set.toList us')
+
+-- | Proves two sets are equivalant, but not identical
```
raduom, line 107: Why not convert them to `Set` and use `Set` equality?

marconi/test/Spec/Utxo.hs
```diff
@@ -22,27 +30,135 @@ import Cardano.Api qualified as C
...
+  length us == length us'
+  &&
+  all (const True) [u `elem` us'| u <- us]
+  &&
+  all (const True) [u `elem` us| u <- us']
+
+-- | UtxoEvents equivalent relationship
+-- Not all utxoEvents have defined `Eq` and `Ord` relationship.
+-- As events are disassembled and reassembled, the Ordering of these sub-parts may change
+-- Therefore we used the Equivalence relationship to show two event are morally the same
```
raduom, line 120: Doesn't this mean we should use `Set` rather than `List` to represent these parts?

raduom (commented):
We should have a discussion about the testing, before merging, but otherwise it seems to be ok.

marconi/test/Spec/Utxo.hs
kayvank: I don't think we can use List for UTXOs, per  Pluts docs on [EUTXO-model figure 3 and Note 3](chrome-extension://efaidnbmnnnibpcajpcglclefindmkaj/https://cicero.ci.iog.io/api/fact/ffdcd1ee-97e2-4661-936d-4b41086d646a/binary) UTXOs should be List and the TxInput should be `Set`
@raduom, would please take a look and let me know if I understood the Specs correctly?

marconi/test/Spec/Utxo.hs
kayvank: see my note below re EUTXO spec and having to use List for UTXOs
 [EUTXO-model figure 3,Page 6, and Note 3, page 15,](https://cicero.ci.iog.io/api/fact/ffdcd1ee-97e2-4661-936d-4b41086d646a/binary)

marconi/src/Marconi/Index/Utxo.hs
```diff
@@ -84,102 +63,153 @@ import Cardano.Api qualified as C
...
-$(makeLenses ''UtxoRow)
+data instance  StorableEvent UtxoHandle = UtxoEvent
+  { ueUtxos       :: !(Set Utxo)
+  , ueInputs      :: (Set C.TxIn)
+  , ueBlockNo     :: C.BlockNo
+  , ueChainPoint  :: C.ChainPoint
+  } deriving (Show, Generic)
 
-type Result = Maybe [UtxoRow]
+class SameChainPoint e where
```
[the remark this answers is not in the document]
kayvank: on closer look, I agree that we dont need the class. Adding blockHash to the Spent table will provide the same functionality, grouping Spent for a given event

marconi/src/Marconi/Indexers.hs
kayvank: the main reason for the `callback` is to provide access to the indexer's latest value,  to functions that run on a different thread, as is the case with mamba. I like your suggestion as it reduces the noise and simplifies the worker function type. However, I think it adds a slight complexity to the `RewindableIndex.Storable` cursor computation. That computation sill have to update the TVar in the `appendEvent` and `flushBufferEvent`. Small price to pay.  I like the idea, I can implement it in my Innovation - Marconi related work

marconi/test/Spec/Utxo.hs
kayvank: ok, my orig analysis was incorrect. we can use Set. lthought there might be more than one UTXO with the same txId, per specs above, the txIx will be different. So we can continue using Set for the UTXOs

marconi/src/Marconi/Index/Utxo.hs
```diff
@@ -239,227 +268,373 @@ instance FromField C.BlockNo where
 instance ToField C.BlockNo where
   toField (C.BlockNo s) = SQLInteger $ fromIntegral s
 
+instance Ord C.ChainPoint where
```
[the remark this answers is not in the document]
kayvank: I can add that to my innovation work.

marconi-mamba/src/Marconi/Api/HttpServer.hs
```diff
@@ -32,11 +33,10 @@ bootstrap env =  runSettings
         (env ^. httpSettings)
         (serve (Proxy @API) (server (env ^. queryEnv ) ) )
 
-server :: DBQueryEnv -> Server API
+server :: UtxoIndexerEnv -> Server API
```
koslambrou, line 36: Given that signature, if we want to add other indexers, we need to add additional parameters or should we just modify `UtxoIndexerEnv` (and rename to `DBQueryEnv` or something)?

marconi/src/Marconi/Index/Utxo.hs
```diff
@@ -1,185 +1,215 @@
...
+  } deriving (Show, Eq, Ord, Generic)
+
+$(makeLenses ''UtxoRow)
+
+newtype instance  StorableResult UtxoHandle = UtxoResult [UtxoRow] deriving Show
+
+data instance  StorableEvent UtxoHandle = UtxoEvent
+  { ueUtxos       :: !(Set Utxo)
+  , ueInputs      :: Set C.TxIn
+  , ueBlockNo     :: C.BlockNo
```
koslambrou, line 128: I think you don't need this one right?

marconi/src/Marconi/Index/Utxo.hs
```diff
@@ -1,185 +1,215 @@
...
+  } deriving (Show, Generic)
+
+eventIsBefore :: StorableEvent UtxoHandle  -> C.ChainPoint -> Bool
+eventIsBefore (UtxoEvent _ _ _ (C.ChainPoint _ slot)) (C.ChainPoint _ slot') =  slot < slot'
+eventIsBefore _ _                                                            = False
+
+instance Semigroup (StorableEvent UtxoHandle) where
+  (UtxoEvent u i b cp) <> (UtxoEvent u' i' _ cp') =
+    if cp /= cp' then
+      error "can not combine UtxoEvents from different ChainPoints"
```
koslambrou, line 139: An `error` in the instance? Mmm... I don't like that :D 
Is that common practice?

marconi/test/Spec/Utxo.hs
```diff
@@ -22,73 +26,180 @@ import Cardano.Api qualified as C
...
-    let (rows :: [Utxo.UtxoRow]) = nub . concatMap Utxo.toRows $ events
-        (addresses :: [C.AddressAny]) = nub . fmap (\r -> r ^. Utxo.utxoRowUtxo . Utxo.utxoAddress ) $ rows
-    ix <- liftIO $ Ix.insertL (events) ndx
-    rowsFromStore <- liftIO $ (concat . catMaybes) <$> forM addresses (\a -> Utxo.queryPlusVolatile ix a  )
-    (null rowsFromStore) === False
-    let (tid :: Utxo.UtxoRow -> C.AddressAny) = (\r -> r ^. Utxo.utxoRowUtxo . Utxo.utxoAddress)
-    (all (`elem` (tid <$> rows)) (tid <$> rowsFromStore)) === True
+  events <- forAll $ Gen.list (Range.linear 2 7) genEvents
+  (storedEvents :: [StorableEvent Utxo.UtxoHandle]) <-
+    (liftIO . Utxo.open ":memory:") (Utxo.Depth  10) -- no DB writes
```
koslambrou, line 186: `Utxo.Depth  10` -> `Utxo.Depth 10`

marconi/test/Spec/Utxo.hs
```diff
@@ -22,73 +26,180 @@ import Cardano.Api qualified as C
...
-    [testPropertyNamed
-     "marconi-index utxos event builder" "Spec. Utxo from Cardano.Api.Tx with targetAddresses"
-     txToUtxoTest
-    , testPropertyNamed
-      "marconi-index utxos event at address" "Spec. Filter UtxoEvents for an address"
-      addressFilteredEventsTest
-    , testPropertyNamed "marconi-index utxos in-memory store" "Spec. Save and retreive UtxoEvents"
+tests = testGroup "Marconi.Utxo.Indexer.Specs are:"
+    [
+     testPropertyNamed "marconi-utxo split-by-address property"
```
koslambrou, line 78: Would be very helpful to add a comment on each property test function describing what you are testing.

marconi/test/Spec/Utxo.hs
```diff
@@ -22,73 +26,180 @@ import Cardano.Api qualified as C
...
+utxoInsertAndQueryTest :: Property
+utxoInsertAndQueryTest = property $ do
+  events <- forAll $ Gen.list (Range.linear 2 9) genEvents
+  depth <- forAll $ Gen.int (Range.linear 2 7) -- force DB writes
+  indexer <- liftIO $ Utxo.open ":memory:" (Utxo.Depth depth)
+             >>= liftIO . Storable.insertMany events
+  let
+    qs :: [StorableQuery Utxo.UtxoHandle]
+    qs = fmap Utxo._address . concatMap (Set.toList . Utxo.ueUtxos) $ events
+  results <- liftIO . traverse (Storable.query Storable.QEverything indexer) $ qs
```
koslambrou, line 137: You also need some property tests with QInterval

marconi/test/Spec/Utxo.hs
```diff
@@ -22,73 +26,180 @@ import Cardano.Api qualified as C
...
 
+storableQueryToShelleyAddr :: StorableQuery Utxo.UtxoHandle -> Maybe (C.Address C.ShelleyAddr)
+storableQueryToShelleyAddr (Utxo.UtxoAddress (C.AddressShelley a ) ) = Just a
+storableQueryToShelleyAddr _                                         = Nothing
+
 addressAnyToShelley :: C.AddressAny -> Maybe (C.Address C.ShelleyAddr)
-addressAnyToShelley  (C.AddressShelley a )=Just a
-addressAnyToShelley  _                     = Nothing
+addressAnyToShelley  (C.AddressShelley a) = Just a
+addressAnyToShelley  _                    = Nothing
```
koslambrou, line 205: Some property tests for `resumeFromStorage` as well.

Some examples would be:
* returns at least the ChainPointAtGenesis point
* it should return the last point of the data indexed on disk

marconi/test/Spec/Utxo.hs
```diff
@@ -22,73 +26,180 @@ import Cardano.Api qualified as C
...
+genChainPoint genB genS = do
+  (C.BlockHeader sn hsh _) <- genBlockHeader genB genS
+  pure $ C.ChainPoint sn hsh
+
+genChainPoint' :: Hedgehog.MonadGen m => m C.ChainPoint
+genChainPoint' = genChainPoint genBlockNo genSlotNo
+
+genEvents :: Gen (Utxo.StorableEvent Utxo.UtxoHandle)
+genEvents = do
+    txs <- Gen.list (Range.linear 2 5)(CGen.genTx C.ShelleyEra)
```
koslambrou, line 65: `(Range.linear 2 5)(CGen.genTx C.ShelleyEra)` -> `(Range.linear 2 5) (CGen.genTx C.ShelleyEra)`

koslambrou, line 65: Why not `1 5` instead of `2 5`?

marconi/src/Marconi/Index/Utxo.hs
```diff
@@ -1,185 +1,213 @@
...
-newtype Depth = Depth Int
+$(makeLenses ''Utxo)
+
+instance Eq Utxo where
+  left == right
+    =  _txId left == _txId right
+    && _txIx left == _txIx right
+
+instance Ord Utxo where
+  left <= right = _txId left <= _txId right
```
eyeinsky, line 111: Should this include comparing the `_txIx` field too?

marconi/src/Marconi/Index/Utxo.hs
```diff
@@ -1,185 +1,213 @@
...
+newtype Depth = Depth Int
 
-type UtxoIndex
-    = SqliteIndex
-      UtxoEvent
-      Notification
-      C.AddressAny
-      Result
+data Utxo = Utxo
+  { _address          :: StorableQuery UtxoHandle
```
eyeinsky, line 93: Why not just have the field `C.AddressAny`? Seems circular to include indexer framework's "main types" within an event.

marconi/src/Marconi/Index/Utxo.hs
```diff
@@ -1,185 +1,213 @@
...
+  , _datumHash        :: Maybe (C.Hash C.ScriptData)
+  , _value            :: C.Value
+  , _inlineScript     :: Maybe Shelley.ScriptInAnyLang -- ByteString -- ^ ReferenceScript
+  , _inlineScriptHash :: Maybe C.ScriptHash
+  } deriving (Show, Generic)
 
-newtype Depth = Depth Int
+$(makeLenses ''Utxo)
+
+instance Eq Utxo where
```
eyeinsky, line 105: Can't the `Eq` instance be simply derived? as this would return true if any of the other fields are different.

marconi/test/Spec/Utxo.hs
```diff
@@ -22,73 +26,180 @@ import Cardano.Api qualified as C
...
+  Hedgehog.assert (equivalentLists storedEvents events)
 
 addressesFromTxOuts :: [C.TxOut C.CtxTx CurrentEra] -> Maybe TargetAddresses
-addressesFromTxOuts [(C.TxOut addressInEra _ _ _)]
+addressesFromTxOuts [C.TxOut addressInEra _ _ _]
     = nonEmpty
     . nub
-    . catMaybes
-    . (fmap (addressAnyToShelley . Utxo.toAddr) )
+    . mapMaybe (addressAnyToShelley . Utxo.toAddr)
```
eyeinsky, line 195: Given that `addressesFromTxOuts` returns a `Maybe (NonEmpty a)`, then why not just `[C.Address C.ShelleyAddr]`?

marconi/src/Marconi/Indexers.hs
```diff
@@ -111,42 +112,43 @@ datumWorker Coordinator{_barrier, _channel} path = do
...
-    :: (Utxo.UtxoIndex -> IO Utxo.UtxoIndex)    -- ^ CPS function used in the queryApi thread, needs to be non-blocking
-    -> Maybe TargetAddresses                    -- ^ Target addresses to filter for
-    -> Worker
-utxoWorker indexerCallback maybeTargetAddresses Coordinator{_barrier, _channel} path = do
-    ix <- Utxo.open path (Utxo.Depth 2160)
-    workerChannel <- atomically . dupTChan $ _channel
-    void . forkIO $ innerLoop workerChannel ix
-    pure [ChainPointAtGenesis]
+utxoWorker_
+  :: (Utxo.UtxoIndexer -> IO ())  -- ^ CPS function used in the queryApi thread, needs to be non-blocking
```
eyeinsky, line 116: I think a callback is not CPS: in continuation passing style the callback is called as the very last statement in the function, but here the callback is called in the middle? (Unless I'm confusing something?)

eyeinsky (commented):
The tests pass, but take a long time, at least on my machine:
```
  Marconi.Utxo.Indexer.Specs are:
    marconi-utxo split-by-address property:       OK (16.26s)
        ✓ Split UtxoEvents by address passed 100 tests.
    marconi-utxo event-to-sqlRows property:       OK (65.56s)
        ✓ Roundtrip UtxoEvents to UtxoRows converion passed 100 tests.
    marconi-utxo storable-query address property: OK (11.26s)
        ✓ Compute StorableQuery addresses from computed Utxo and generated Cardano.Api.Tx passed 100 tests.
    marconi-utxo storage-roundtrip property:      OK (64.15s)
        ✓ Roundtrip storage test passed 100 tests.
    marconi-utxo insert-query property:           OK (76.17s)
        ✓ Insert Events, and then query for events by address test passed 100 tests.
```
Any idea why? There is no test with cardano-testnet and no threadDelay's either (other than in the json-rpc example, but that is not part of these tests).

marconi/src/Marconi/Index/Utxo.hs
kayvank: I did that to save a bit of processing as txid and txix combination is unique. However, I think your point is valid. Will fix

marconi/src/Marconi/Index/Utxo.hs
kayvank: no, I changed it.

marconi-mamba/src/Marconi/Api/HttpServer.hs
kayvank: if we go with this design, we should add the new idexer to  [Marconi.Api.Types.UtxoIndexerWrapper ](https://github.com/kayvank/plutus-apps/blob/1314/address-utxo-indexer-new-marconi-api/marconi-mamba/src/Marconi/Api/Types.hs#L46) and change name(s) accordingly

marconi/test/Spec/Utxo.hs
kayvank: The signature for theTargetAddress is NonEmpty

kayvank:
> The tests pass, but take a long time, at least on my machine:
> 
> ```
>   Marconi.Utxo.Indexer.Specs are:
>     marconi-utxo split-by-address property:       OK (16.26s)
>         ✓ Split UtxoEvents by address passed 100 tests.
>     marconi-utxo event-to-sqlRows property:       OK (65.56s)
>         ✓ Roundtrip UtxoEvents to UtxoRows converion passed 100 tests.
>     marconi-utxo storable-query address property: OK (11.26s)
>         ✓ Compute StorableQuery addresses from computed Utxo and generated Cardano.Api.Tx passed 100 tests.
>     marconi-utxo storage-roundtrip property:      OK (64.15s)
>         ✓ Roundtrip storage test passed 100 tests.
>     marconi-utxo insert-query property:           OK (76.17s)
>         ✓ Insert Events, and then query for events by address test passed 100 tests.
> ```
> 
> Any idea why? There is no test with cardano-testnet and no threadDelay's either (other than in the json-rpc example, but that is not part of these tests).

we discussed this in our weekly meeting and the extra time seem to be related to the Cardano API generators. I made a few changes and use different UtxoEvent generator. The results seem to be better on my dev machine:
```
Running 1 test suites...
Test suite marconi-test: RUNNING...
Marconi
  Marconi.Utxo.Indexer.Specs are:
    marconi-utxo split-by-address property:       OK (0.14s)
        ✓ filter UtxoEvent for Utxos with address in the TargetAddress passed 100 tests.
    marconi-utxo event-to-sqlRows property:       OK (0.39s)
        ✓ Roundtrip UtxoEvents to UtxoRows converion passed 100 tests.
    marconi-utxo storable-query address property: OK (11.27s)
        ✓ Compute StorableQuery addresses from computed Utxo and generated Cardano.Api.Tx passed 100 tests.
    marconi-utxo storage-roundtrip property:      OK (0.47s)
        ✓ Roundtrip storage test passed 100 tests.
    marconi-utxo insert-query property:           OK (0.66s)
        ✓ Insert Events, and then query for events by address test passed 100 tests.

All 5 tests passed (12.92s)
Test suite marconi-test: PASS
Test suite logged to:
/home/kayvan/dev/workspaces/workspace-iohk/plutus-apps/dist-newstyle/build/x86_64-linux/ghc-8.10.7/marconi-1.1.0.0/t/marconi-test/noopt/test/marconi-1.1.0.0-marconi-test.log
1 of 1 test suites (1 of 1 test cases) passed.
```

marconi/test/Spec/Utxo.hs
eyeinsky: Right, so why not a `pure <$> addressAnyToShelley (Utxo.toAddr addressInEra)` then? (I.e why convert it to a single element list first at all and then operate on that when you don't need to. The `nub` is entirely unneeded there too, as the list will ever contain 0 or 1 value)

But a broader question about this function:  it takes a list of `[TxOut]` but only returns a `Just TargetAddresses` when the argument list has exactly one element and otherwise returns `Nothing`, why is that?

marconi/test/Spec/Utxo.hs
```diff
@@ -22,73 +26,180 @@ import Cardano.Api qualified as C
...
+
+-- NOTE --
+-- | UtxoEvents equivalent relationship
+-- Not all utxoEvent attributes have defined `Eq` and/or `Ord` relationship defined.
+-- As events are disassembled and reassembled, the Ordering of these sub-parts may change in the coresponding collections.
+-- Therefore we used the Equivalence relationship to show two event are morally equal.
+
+eventsToRowsRoundTrip :: Property
+eventsToRowsRoundTrip  = property $ do
+  events <- forAll $ Gen.list (Range.linear 2 7 )genEvents
```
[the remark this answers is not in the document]
koslambrou: @kayvank

marconi-mamba/test/Spec/UtxoIndexersQuery.hs
```diff
@@ -1,102 +1,152 @@
...
+     "Spec. Insert events and query for utxo's with address in the generated ShelleyEra targetAddresses"
+     queryTargetAddressTest
+    ]
+
+-- | generate some Utxo events, store them and fetch them.
+--
+queryTargetAddressTest :: Property
+queryTargetAddressTest = property $ do
+  events <- forAll $ Gen.list (Range.linear 1 3) genShelleyEvents
+  depth <- forAll $ Gen.int (Range.linear 6 9 ) -- force DB writes
```
koslambrou, line 125: Remove space after `9`

marconi/test/Spec/Utxo.hs
```diff
@@ -22,73 +28,284 @@ import Cardano.Api qualified as C
...
+utxoStorageTest :: Property
+utxoStorageTest = property $ do
+  events <- forAll $ Gen.list (Range.linear 1 5) genEvents
+  (storedEvents :: [StorableEvent Utxo.UtxoHandle]) <-
+    (liftIO . Utxo.open ":memory:") (Utxo.Depth  10)
+     >>= liftIO . Storable.insertMany events
+     >>= liftIO . Storable.getEvents
+  Hedgehog.assert (equivalentLists storedEvents events)
+
+-- Insert Utxo events in storage, and retreive the events by address
```
koslambrou, line 188: retreive -> retrieve

marconi/test/Spec/Utxo.hs
```diff
@@ -22,73 +28,284 @@ import Cardano.Api qualified as C
...
+    qs = fmap (Utxo.UtxoAddress . Utxo._address) . concatMap (Set.toList . Utxo.ueUtxos) $ events
+  results <- liftIO . traverse (Storable.query Storable.QEverything indexer) $ qs
+  let rows = concatMap (\(Utxo.UtxoResult rs) -> rs ) results
+  computedEvent <-
+    liftIO . Utxo.rowsToEvents (Utxo.getTxIns (getConn indexer) ) $ rows
+  Hedgehog.assert (equivalentLists
+                   computedEvent
+                   (filter (\e -> Utxo.ueChainPoint e /= C.ChainPointAtGenesis) events) )
+
+-- Insert Utxo events in storage, and retreive the events by address and query interval
```
koslambrou, line 207: retreive -> retrieve

marconi/test/Spec/Utxo.hs
```diff
@@ -22,73 +28,284 @@ import Cardano.Api qualified as C
...
+  let
+    qs :: [StorableQuery Utxo.UtxoHandle]
+    qs = fmap (Utxo.UtxoAddress . Utxo._address) . concatMap (Set.toList . Utxo.ueUtxos) $ events
+  results <- liftIO . traverse (Storable.query (Storable.QInterval (chainpoints !! 0)(chainpoints !! 1)) indexer) $ qs
+  let rows = concatMap (\(Utxo.UtxoResult rs) -> rs ) results
+  computedEvent <-
+    liftIO . Utxo.rowsToEvents (Utxo.getTxIns (getConn indexer) ) $ rows
+  Hedgehog.assert (equivalentLists computedEvent [event0,event1])
+
+-- This test is in supporto TargetAddresses user's may have entered through CLI
```
koslambrou, line 227: > This test is in supporto TargetAddresses user's may

What?

marconi/test/Spec/Utxo.hs
```diff
@@ -22,73 +28,284 @@ import Cardano.Api qualified as C
...
+          . concatMap (Set.toList . Utxo.ueUtxos)
+          $ sameAddressEvents
+        (actualAddresses :: [C.AddressAny])
+          = toListOf (folded
+                      . Utxo.address
+                      . filtered (== targetAddress) )
+            $ Utxo.ueUtxos event
+    computedAddresses === actualAddresses
+
+-- Test to make sure we only make Utxo's from chain events for the TargetAddresses user has provided thrugh CLI
```
koslambrou, line 254: thrugh -> through

marconi/test/Spec/Utxo.hs
```diff
@@ -22,73 +28,284 @@ import Cardano.Api qualified as C
...
+  results <- liftIO . traverse (Storable.query (Storable.QInterval (chainpoints !! 0)(chainpoints !! 1)) indexer) $ qs
+  let rows = concatMap (\(Utxo.UtxoResult rs) -> rs ) results
+  computedEvent <-
+    liftIO . Utxo.rowsToEvents (Utxo.getTxIns (getConn indexer) ) $ rows
+  Hedgehog.assert (equivalentLists computedEvent [event0,event1])
+
+-- This test is in supporto TargetAddresses user's may have entered through CLI
+-- Remove from UtxoEvent Utxos with address not in TargetAddress list
+-- TODO cleaner approach
+-- Generate Addresse -> generate Utxos -> generateEvents and use a subset of those addresses to filter in/out
```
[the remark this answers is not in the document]
kayvank: I was considering this approach:
- generate addresses
- generate events for those addresses
- run the test while randomly picking targetAddresses from the set of addresses.
However, I think I would like to implement that approach in the morconi-test related work, [plt-1360](https://input-output.atlassian.net/browse/PLT-1360)

marconi/test/Spec/Utxo.hs
kayvank: I re-worded the comment

marconi/test/Spec/Utxo.hs
kayvank: the function is poorly named, but the comment still applies.
You're correct as the nub is nor required.
The main idea here is the create/make a TargetAddress that is relevant. Another words, a targetAddress that is part of the transaction. So the TargetAddress is made in reverse, we look into the TxOut and extract an address, and convert it to TargetAddress.

A better approach, is to 
- generate addresses
- generate events based on those addresses
- randomly select a an address as TargetAddress and run the test.
-
This required a bit more work, so I opted out to do the work in the [PLT-1360](https://input-output.atlassian.net/browse/PLT-1360)

koslambrou (approved):

PR #903, opened 2023-01-16
PLT-1350 Add submitTx* functions for cardano Tx

berewt:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


james-iohk:
@berewt `mustReferenceOutput` has the following comment:
```
-- plutus-tx-constraints tests
-- all below to be covered by the above tests when MustMint is implemented (PLT-672)
```
Can we now do away with all below?

PR #904, opened 2023-01-18
Threat modelling framework

UlfNorell:
Updates the quickcheck-contractmodel dependency to the latest version with the new threat modelling framework.

- Reenable double satisfaction tests (this now comes with the threat modelling)
- Some simple threat model examples (GameStateMachine and Escrow)
- Use min Ada balance change logic from quickcheck-contractmodel

### DIFF
--- a/plutus-use-cases/plutus-use-cases.cabal
+++ b/plutus-use-cases/plutus-use-cases.cabal
@@ -224,6 +224,38 @@ test-suite plutus-use-cases-test
     , tasty-quickcheck
     , text
 
+  -- General dependencies
+  build-depends:
+      base           >=4.7 && <5
+    , containers
+    , contractmodel
+    , lens
+    , mmorph
+    , mtl
+    , pretty
+
+  -- QuickCheck dependencies
+  build-depends:
+      QuickCheck
+    , quickcheck-dynamic  >=2.0.0
+
+  -- Plutus and Cardano dependencies
+  build-depends:
+      cardano-api                  >=1.35
+    , cardano-binary
+    , cardano-ledger-alonzo
+    , cardano-ledger-babbage
+    , cardano-ledger-core
+    , cardano-ledger-shelley
+    , cardano-ledger-shelley-ma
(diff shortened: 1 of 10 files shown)


plutus-use-cases/plutus-use-cases.cabal
```diff
@@ -224,6 +224,38 @@ test-suite plutus-use-cases-test
...
+    , cardano-ledger-shelley
+    , cardano-ledger-shelley-ma
+    , cardano-slotting
+    , ouroboros-consensus
+    , ouroboros-consensus-cardano
+    , plutus-ledger-api
+    , plutus-tx
+    , strict-containers
+    , time
+
```
koslambrou, line 258: These should go inside one of the categories above.

plutus-use-cases/test/Spec/Auction.hs
```diff
@@ -358,9 +367,9 @@ check_propAuctionWithCoverage = do
         (set minLogLevel Critical options) covopts (const (pure True))
   writeCoverageReport "Auction" cr
 
--- TODO: re-activate when thread modelling framework is merged
--- prop_doubleSatisfaction :: Actions AuctionModel -> Property
--- prop_doubleSatisfaction = checkDoubleSatisfactionWithOptions options defaultCoverageOptions
+-- | Note: this property fails, because the contract is vulnerable to double satisfaction.
```
koslambrou, line 370: If it fails, it shouldn't pass CI right?

koslambrou (approved):
LGTM. Conflicts need to be fixed, I'll merge once that's done :) Sorry for the delay.

plutus-use-cases/test/Spec/Auction.hs
UlfNorell: It included in the test suite under an `expectFailure` (see below)`, so CI passes only when it *does* fail.

koslambrou: Oh missed it. Thanks

PR #908, opened 2023-01-24
Upgrade plutus-use-cases to V2

james-iohk:
First commit is just for Crowdfunding, then decided to just do all the rest.

This is mostly changing imports and qualifiers from V1 to V2  but there are significant changes such as handling OutputDatum in Uniswap, e.g. see [here](https://github.com/input-output-hk/plutus-apps/pull/908/files#diff-1e97de19ee4582f83ed6914fdffc7755b86522bb0ef654532f83c31814a4e8a1R164-R169).


sjoerdvisscher (approved):
How hard would it be to get rid of V1 imports altogether? And get those (even though they are the exact same type) from a V2 module or from an version-less module?

james-iohk:
> How hard would it be to get rid of V1 imports altogether? And get those (even though they are the exact same type) from a V2 module or from an version-less module?

@sjoerdvisscher Have removed remaining V1 where possible. Two that are not exposed through Plutus V2 Api are `Plutus.V1.Ledger.Address (toPubKeyHash)` and `Plutus.V1.Ledger.Scripts (scriptSize)`, so had to leave them.  It would be nice if Plutus API included more in its common interface (e.g. POSIXTime).

koslambrou (approved):

berewt (approved):
It's great to see that we have everything in place to make the transition that smooth!

PR #909, opened 2023-01-27
Update hydra refs in docs

zeme-iohk:


kwxm:
This looks right, and I can confirm that it works.

kwxm (approved):
This looks right, and I can confirm that it works.

koslambrou (approved):

PR #910, opened 2023-01-27
Mention GC_DONT_GC=1 in CONTRIBUTING.adoc

zeme-iohk:
Several people have complained that running `nix-shell` segfaults.
We can't expect them to know the `GC_DONT_GC=1` trick.
This PR updates the CONTRIBUTING

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -8,7 +8,8 @@ Make sure you have set up the link:./README{outfilesuffix}#prerequisites[prerequ
 
 === How to get a shell environment with tools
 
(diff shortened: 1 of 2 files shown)


michaelpj:
We can write it in CONTRIBUTING instead? In some ways it's just as surprising to have to run a different script to do it.

michaelpj:
I mean, why not just say in CONTRIBUTING "if you get segfaults, do this"?

michaelpj:
Where "this" = "use this environment variable"

zeme-iohk:
ok sounds good

PR #913, opened 2023-01-27
Add MintBurn indexer

eyeinsky:
Ready for review, changes since the mob review:
- handling of intervals and a new interval test
- end-to-end test now also uses generator to create TxMintValue

I think I've also found and fixed two bugs in rewindable index:
- [4af8f94](https://github.com/input-output-hk/plutus-apps/pull/913/commits/4af8f946905c733abec642a65707a644b5900b76) `filterWithQueryInterval` wasn't using the start of interval to filter out elements
- [594f20a](https://github.com/input-output-hk/plutus-apps/pull/913/commits/594f20a57c765d4944023ea61d05215d40ae46a3) `rewind` didn't account for gaps in events

Not sure about the `filterWithQueryInterval` fix as it makes a test fail in rewindable-index: New index properties -> Insert is folding the structure.

@raduom should inspect these and perhaps add tests to for regressions.

(CI fails currently due to a bug in cabal-fmt, being worked on here: https://github.com/input-output-hk/plutus-apps/pull/921)

---

Pre-submit checklist:
- Branch
    - [X] Tests are provided (if possible)
    - [X] Important changes are reflected in changelog.d of the affected packages
- PR
    - [X] Self-reviewed the diff
    - [X] Reviewer requested

### DIFF
--- a/marconi/test/MintBurn.hs
+++ b/marconi/test/MintBurn.hs
@@ -1,0 +1,413 @@
+{-# LANGUAGE DataKinds          #-}
+{-# LANGUAGE FlexibleContexts   #-}
+{-# LANGUAGE LambdaCase         #-}
+{-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE OverloadedStrings  #-}
+{-# LANGUAGE TemplateHaskell    #-}
+module MintBurn where
+
+import Codec.Serialise (serialise)
+import Control.Concurrent qualified as IO
+import Control.Concurrent.Async qualified as IO
+import Control.Concurrent.STM qualified as IO
+import Control.Exception (catch)
+import Control.Lens qualified as Lens
+import Control.Monad (foldM, forM, replicateM, void, when)
+import Control.Monad.IO.Class (liftIO)
+import Data.ByteString.Lazy qualified as LBS
+import Data.ByteString.Short qualified as SBS
+import Data.Coerce (coerce)
+import Data.List.NonEmpty (NonEmpty ((:|)))
+import Data.List.NonEmpty qualified as NonEmpty
+import Data.Map qualified as Map
+import Data.Maybe (mapMaybe)
+import Data.Set qualified as Set
+import Data.String (fromString)
+import Data.Word (Word64)
+import Prettyprinter (defaultLayoutOptions, layoutPretty, pretty, (<+>))
+import Prettyprinter.Render.Text (renderStrict)
+import Streaming.Prelude qualified as S
+import System.Directory qualified as IO
+import System.FilePath ((</>))
+
+import Hedgehog (Gen, Property, forAll, (===))
+import Hedgehog qualified as H
+import Hedgehog.Extras.Test qualified as HE
+import Hedgehog.Extras.Test.Base qualified as H
+import Hedgehog.Gen qualified as Gen
+import Hedgehog.Range qualified as Range
+import Test.Tasty (TestTree, testGroup)
+import Test.Tasty.Hedgehog (testPropertyNamed)
+
+import Cardano.Api qualified as C
+import Cardano.Api.Shelley qualified as C
+import Cardano.BM.Setup (withTrace)
+import Cardano.BM.Trace (logError)
+import Cardano.BM.Tracing (defaultConfigStdout)
+import Cardano.Streaming (ChainSyncEventException (NoIntersectionFound), withChainSyncEventStream)
+import Gen.Cardano.Api.Typed qualified as CGen
+import Plutus.V1.Ledger.Api qualified as PlutusV1
+import PlutusTx qualified
+import Test.Base qualified as H
+import Testnet.Cardano qualified as TN
+
+import Helpers qualified as TN
+import Marconi.Index.MintBurn qualified as MintBurn
+import Marconi.Indexers qualified as M
+import Marconi.Logging ()
+import RewindableIndex.Storable qualified as RI
+
+-- | Each test case is described beside every top level property
+-- declaration.
+tests :: TestTree
+tests = testGroup "MintBurn"
+  [ testPropertyNamed
+      "Mints in `TxBodyContent` survive `makeTransactionBody` and end up in expected place in `TxBody`"
+      "mintsPreserved" mintsPreserved
+  , testPropertyNamed
+      "Mint events created and indexed are returned when querying the indexer"
+      "queryMintedValues" queryMintedValues
+  , testPropertyNamed
+      "Querying a resumed indexer returns only the persisted events"
+      "resume" resume
+  , testPropertyNamed
+      "Rewinding to any slot forgets any newer events than that slot"
+      "rewind" rewind
+  , testPropertyNamed
+      "Intervals work as expected"
+      "intervals" intervals
+  , testPropertyNamed
+      "Indexing a testnet and then submitting a transaction with a mint event to it has the indexer receive that mint event"
+      "endToEnd" endToEnd
+  ]
+
+-- | This is a sanity-check test that turns a TxBodyContent with mint
+-- events into a TxBody through `makeTransactionBody` and checks if
+-- the mint events are found in the result. It doesn't test an
+-- indexer.
+mintsPreserved :: Property
+mintsPreserved = H.property $ do
+  mintValue <- forAll genTxMintValue
+  C.Tx txb _ :: C.Tx C.AlonzoEra <- forAll (genTxWithMint mintValue) >>= \case
+    Left err  -> fail $ "TxBodyError: " <> show err
+    Right tx' -> return tx'
+  -- Index the transaction:
+  let mints = MintBurn.txbMints txb
+      gottenPolicyAssets = map (\mint -> (MintBurn.mintAssetPolicyId mint, MintBurn.mintAssetAssetName mint, MintBurn.mintAssetQuantity mint)) mints
+  -- Print footnote should the test fail:
+  let generatedPolicyAssets = getPolicyAssets mintValue
+  H.footnote $ "Assets to be created: " <> show generatedPolicyAssets <> "\n"
+            <> "Assets gotten: " <> show gottenPolicyAssets
+  -- The assets that were used to construct the transaction were found
+  -- in the generate transaction:
+  equalSet generatedPolicyAssets gottenPolicyAssets
+
+-- | Create transactions, index them, query indexer and find mint events.
+queryMintedValues :: Property
+queryMintedValues = H.property $ do
+  (indexer, insertedEvents, _) <- generateAndIndexEvents ":memory:"
+  -- Query results:
+  MintBurn.MintBurnResult queryResult <- liftIO $ RI.query RI.QEverything indexer MintBurn.Everything
+  -- Compare the sets of events inserted to the indexer and the set
+  -- gotten out of the indexer:
+  equalSet (MintBurn.groupBySlotAndHash insertedEvents) (MintBurn.fromRows queryResult)
+
+-- | Insert some events to an indexer, then recreate it from what is
+-- on disk (the in-memory part is lost), then query it and find all
+-- persisted events and none of the in-memory events.
+resume :: Property
+resume = H.property $ do
+  -- Index events that overflow:
+  (indexer, events, (bufferSize, _nTx)) <- generateAndIndexEvents ":memory:"
+  -- Open a new indexer based off of the old indexers sql connection:
+  indexer' <- liftIO $ mkNewIndexerBasedOnOldDb indexer
+  MintBurn.MintBurnResult queryResult <- liftIO $ RI.query RI.QEverything indexer' MintBurn.Everything
+  let expected = MintBurn.groupBySlotAndHash $ take (eventsPersisted bufferSize (length events)) events
+  -- The test: events that were persisted are exactly those we get from the query.
+  equalSet expected (MintBurn.fromRows queryResult)
+
+-- | Test that rewind (rollback for on-disk events) behaves as
+-- expected: insert events such that buffer overflows, rollback so far
+-- back that some events were already persisted, find no newer events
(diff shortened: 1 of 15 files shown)


marconi/src/Marconi/Index/Common.hs
```diff
@@ -0,0 +1,19 @@
...
+
+
+instance SQL.ToField (C.Hash C.BlockHeader) where
+  toField f = SQL.toField $ C.serialiseToRawBytes f
+
+instance SQL.FromField (C.Hash C.BlockHeader) where
+   fromField f =
+      SQL.fromField f <&>
+        fromMaybe (error "Cannot deserialise block hash") .
+          C.deserialiseFromRawBytes (C.proxyToAsType Proxy)
```
koslambrou, line 19: Shouldn't this go in the Orphans module?

marconi/test/MintBurn.hs
```diff
@@ -0,0 +1,400 @@
...
+  , testPropertyNamed "endToEnd" "endToEnd" endToEnd
+  ]
+
+-- | This is a sanity-check test that turns a TxBodyContent with mint
+-- events into a TxBody and checks if the mint events are found in the
+-- result. It doesn't test an indexer.
+--
+-- Create a transaction that mints a value, convertit to a @Tx@
+-- through `makeTransactionBody` and `signShelleyTransaction` and find
+-- that the minted assets are still there.
```
koslambrou, line 78: Maybe reword this? The two paragraphs seem to repeat the same thing but in different terms.

koslambrou (approved):
LGTM, but I'm not sure about the changes in `rewindable-index`. I'll let @raduom provide input there.

marconi/src/Marconi/CLI.hs
```diff
@@ -156,6 +158,10 @@ optionsParser =
                       <> Opt.help "disable epoch stakepool size indexers."
                       <> Opt.showDefault
                      )
+    <*> Opt.switch (Opt.long "disable-mintburn"
+                      <> Opt.help "disable mint/burn indexers."
+                      <> Opt.showDefault
```
kayvank, line 163: showDefault doesn't apply to switch and not required here. It has no effect!

BTW, I am at fault as I had erroneously, used them on a previous PR :(

marconi/src/Marconi/Index/MintBurn.hs
```diff
@@ -0,0 +1,310 @@
...
+    rs' <- NE.groupBy1 ((==) `on` txId) rs -- group by TxId
+    pure (txId r, rowToMintAsset <$> rs')
+  where
+    slotNo = view L._1 :: Row -> C.SlotNo
+    hash = view L._2 :: Row -> C.Hash C.BlockHeader
+    txId = view L._3 :: Row -> C.TxId
+    rowToMintAsset :: Row -> MintAsset
+    rowToMintAsset row = MintAsset (row^.L._4) (row^.L._5) (row^.L._6) (row^.L._7) (row^.L._8)
+
+sqliteSelectByTxIdPolicyId :: SQL.Connection -> (SQL.Query, [NamedParam]) -> C.TxId -> C.PolicyId -> IO [TxMintEvent]
```
kayvank, line 195: :heart:

kayvank (approved):

marconi/src/Marconi/Index/MintBurn.hs
```diff
@@ -0,0 +1,310 @@
...
+      \        , assetName, quantity, redeemerIx, redeemerData \
+      \     FROM minting_policy_events                         \
+      \     " <> whereClause <> "                              \
+      \ ORDER BY slotNo, txId                                  "
+
+intervalToWhereClause :: RI.QueryInterval C.ChainPoint -> (SQL.Query, [NamedParam])
+intervalToWhereClause qi = case qi of
+  RI.QEverything -> ("", [])
+  RI.QInterval from to
+    | Just from' <- slotMaybe from, Just to' <- slotMaybe to -> ("slotNo BETWEEN :fromSlot AND :toSlot" , [":fromSlot" := from', ":toSlot" := to'])
```
raduom, line 220: This is not how intervals actually work. The condition for the persistent layer should be that the slot < end (where end is the top of the interval).

Reference: https://github.com/input-output-hk/plutus-apps/blob/d9eebba0e5b7f3f0d955e9755c7a031eb4de1303/rewindable-index/src/RewindableIndex/Storable.hs#L318-L329

marconi/src/Marconi/Index/MintBurn.hs
```diff
@@ -0,0 +1,310 @@
...
+    | Just to' <- slotMaybe to -> ("slotNo <= :toSlot" , [":toSlot" := to'])
+    | otherwise -> ("FALSE", [])
+    where
+      slotMaybe :: C.ChainPoint -> Maybe C.SlotNo
+      slotMaybe = \case
+        C.ChainPoint slotNo _ -> Just slotNo
+        C.ChainPointAtGenesis -> Nothing
+
+groupBySlotNo :: [TxMintEvent] -> [TxMintEvent]
```
raduom, line 229:
```suggestion
groupBySlotAndHash :: [TxMintEvent] -> [TxMintEvent]
```
You are grouping over two fields, not just one.

marconi/src/Marconi/Index/MintBurn.hs
```diff
@@ -0,0 +1,310 @@
...
+
+data instance RI.StorableQuery MintBurnHandle
+  = ByTxIdAndPolicyId C.TxId C.PolicyId
+  | Everything
+
+data instance RI.StorableResult MintBurnHandle
+  = MintBurnResult [Row]
+
+instance RI.Queryable MintBurnHandle where
+  queryStorage queryInterval memoryEvents (MintBurnHandle sqlCon _k) query = case query of
```
raduom, line 260: You are not getting the proper data here, see the previous comment.

You need to include all the data persisted from the point you found to the bottom of the event stack. The way that the state is computed is by running a fold over the in-memory + persisted data sets, so you will need all the data from the persistence store.

Then you stitch the persisted events and the in-memory events together and you run them through the `filterWithQueryInterval` function that should return the final result.

For reference, here how it is done in the test indexer implementation: https://github.com/input-output-hk/plutus-apps/blob/d9eebba0e5b7f3f0d955e9755c7a031eb4de1303/rewindable-index/test/RewindableIndex/Spec/Sqlite.hs#L192-L206

marconi/src/Marconi/Index/MintBurn.hs
```diff
@@ -0,0 +1,310 @@
...
+    pure h
+
+  getStoredEvents (MintBurnHandle sqlCon k) =
+    map MintBurnEvent . fromRows <$> SQL.query sqlCon query (SQL.Only k)
+    where
+      query =
+        " SELECT slotNo, blockHeaderHash, txId, policyId, assetName, quantity  \
+        \        redeemerIx, redeemerData                                      \
+        \   FROM minting_policy_events                                         \
+        \  WHERE slotNo >= (SELECT MAX(slotNo) - ? FROM minting_policy_events) \
```
raduom, line 287: I would replace this one with a `LIMIT k`. It will give more events which for resume (where this is used) is a good thing.

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -335,14 +335,7 @@ filterWithQueryInterval
...
 filterWithQueryInterval QEverything es = es
-filterWithQueryInterval (QInterval start end) es =
-  let es' = takeWhile (withPoint (\p -> p <= end)) es
-   in if not (null es') && withPoint (\p -> p >= start) (last es')
-      then es'
-      else []
-  where
-    withPoint :: (StorablePoint h -> Bool) -> StorableEvent h -> Bool
-    withPoint f e = let p = getPoint e in f p
+filterWithQueryInterval (QInterval start end) es = takeWhile (\e -> getPoint e <= end) $ dropWhile (\e -> getPoint e < start) es
```
raduom, line 338: This would be a nice implementation, but it does not work properly. I think the comment may not be doing a very good job at explaining what this function does. Maybe I need to reword it.

marconi/src/Marconi/Index/MintBurn.hs
eyeinsky: Yup, the comment I read several times without added clarity. So I want all events from the very beginning up to the latest point between start and end?

marconi/src/Marconi/Index/MintBurn.hs
eyeinsky: Right, but given that my isn't a fold, but a map, then why do I need to load all the events from the database?

marconi/src/Marconi/Index/MintBurn.hs
eyeinsky: Hmm, right. I wonder if `getStoredEvents` is a good function to use for resume points anyway? This idea I took from the ScriptTx indexer as you used it there. But it might be better to write the sql directly into `resumeFromStorage` instead?

Trying to understand the meaning of `getStoredEvents`: it's used [here in RewindableIndex.Storable.getEvents](https://github.com/input-output-hk/plutus-apps/blob/main/rewindable-index/src/RewindableIndex/Storable.hs#L246) -- does it need to get *all* events, or some of the events? In ScriptTx indexer it seems to get only some of the events, while for `RewindableIndex.Storable.getEvents` it appears to get all of them?

marconi/src/Marconi/CLI.hs
eyeinsky: a4deee5dab1c664e71443373c7ceceb89e94228b

marconi/src/Marconi/Index/MintBurn.hs
eyeinsky: I agree -- it saved me copy-pasting of the same sql query multiple times. I think it was also you who started using the named parameters, I seem to remember them from a pull request but only now got around using them.

marconi/src/Marconi/Index/MintBurn.hs
eyeinsky: 8f213ed7a, 9005d2981976a5c5fa3bc60eb87dba848f1c21d7

marconi/src/Marconi/Index/Common.hs
koslambrou: Yep looks good

rewindable-index/src/RewindableIndex/Storable.hs
koslambrou: @eyeinsky That part needs to be reverted by I think

eyeinsky: Right, but the tests will fail then. The wider issue here is that it is unclear what an interval means and how could I test it.

eyeinsky: a928d3434, 4d2e1e617

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -288,28 +288,28 @@ rewind
...
-      v <- V.freeze $ VM.slice 0 (s ^. storage . cursor) (s ^. storage . events)
-      pure $ do
-        ix   <- VG.findIndex (\e -> getPoint e == p) v
-        pure $ s & storage . cursor .~ (ix + 1)
-    resetMemory :: State h -> h -> State h
-    resetMemory s' h =
-      s' & handle  .~ h
-         & storage . cursor .~ 0
+rewind p s = if s ^. storage . cursor == 0
+  -- Buffer is empty, rewind storage just in case:
```
raduom, line 296: What is the difference between this code and the one before it (other than the change from `=` to `<` and `ix + 1` to `ix`)? Wouldn't the previous code work with just those changes?

raduom (approved):

rewindable-index/src/RewindableIndex/Storable.hs
eyeinsky: The difference is that we don't rewind storage when it is not required.

I.e, with just the changes you mention, `findIndex` would return `Nothing` in both when buffer is empty or when all events are older than the rewind point. In the first case storage needs to be rewound, in the second case since we know that all events in buffer are older than rewind point then everything in storage must be as well.

Also, with just the changes you mention, `findIndex` would return `Just ix` in both when an event with point later than rewind point is the *first element* in the buffer or when it's somewhere in the middle. If the found index is the first point in buffer then storage needs to be rewound, in the second case only the buffer needs to be rewound as there are events in the buffer that are older than the rewind point -- which means events in storage are even older still.

marconi/src/Marconi/Index/MintBurn.hs
```diff
@@ -0,0 +1,309 @@
...
+    rowToMintAsset row = MintAsset (row^.L._4) (row^.L._5) (row^.L._6) (row^.L._7) (row^.L._8)
+
+sqliteSelectByTxIdPolicyId :: SQL.Connection -> (SQL.Query, [NamedParam]) -> C.TxId -> C.PolicyId -> IO [TxMintEvent]
+sqliteSelectByTxIdPolicyId sqlCon (conditions, params) txId policyId =
+  fromRows <$> SQL.queryNamed sqlCon query ([":txId" := txId, ":policyId" := policyId] <> params)
+  where query =
+          " SELECT slotNo, blockHeaderHash, txId, policyId                    \
+          \      , assetName, quantity, redeemerIx, redeemerData              \
+          \   FROM minting_policy_events                                      \
+          \  WHERE txId = :txId AND policyId = :policyId " <> conditions <> " \
```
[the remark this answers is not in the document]
eyeinsky: The `conditions` only has the template (for example `slotNo <= :toSlot`), so sqlite-simple still does variable interpolation. But still good to keep this in mind. :)

eyeinsky:
Some of the Plutus E2E tests fail in CI but (1) those are unrelated to this pr and (2) they are being worked on elsewhere.

PR #915, opened 2023-01-30
PLT-662 Added Address->Datum indexer in Marconi

koslambrou:
* Indexes all datums for a given address 
  * takes about 3h starting at Alonzo era - which is the era that introduces datums
  * total disk db size of ~2GB

* Supports resuming

* Does *not* include the JSON-RPC interface on top of this indexer

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi/src/Marconi/Index/AddressDatum.hs
+++ b/marconi/src/Marconi/Index/AddressDatum.hs
@@ -1,0 +1,510 @@
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleInstances  #-}
+{-# LANGUAGE GADTs              #-}
+{-# LANGUAGE NamedFieldPuns     #-}
+{-# LANGUAGE OverloadedStrings  #-}
+{-# LANGUAGE QuasiQuotes        #-}
+{-# LANGUAGE TemplateHaskell    #-}
+{-# LANGUAGE TupleSections      #-}
+
+-- | Module for indexing the datums for all addresses in the Cardano blockchain.
+
+-- This module will create the SQL tables:
+
+-- + table: address_datums
+--
+-- @
+--    |---------+------------+---------+------------|
+--    | address | datum_hash | slot_no | block_hash |
+--    |---------+------------+---------+------------|
+-- @
+--
+-- + table: datumhash_datum
+--
+-- @
+--    |------------+-------|
+--    | datum_hash | datum |
+--    |------------+-------|
+-- @
+
+-- To create these tables, we extract all transactions outputs from each transactions fetched with
+-- the chain-sync protocol of the local node.
+
+-- Here is a synopsis of the indexing algorithm.
+
+-- Each transaction output contains an address along with an optional datum hash or an optional inline
+-- datum (actual datum).
+-- In the inline datum scenario, we simply create an entry in the `address_datums` table with the hash
+-- of the datum, and add an entry in the `datumhash_datum` table.
+-- In the datum hash scenario, we create an entry in the `address_datums` table, but not in the
+-- `datumhash_datum` table, as we don't know the actual value of the datum and we can't infer it from
+-- the hash.
+-- In that last scenario, we can resolve in the datum hash in one of three ways (which we then simply
+-- add an entry in the `datumhash_datum` table):
+--
+--     * a different transaction output has an inline datum with that same hash
+--
+--     * a datum with that same hash has been found in the in transaction body
+--
+--     * a datum with that same hash was included in the witnesses for a Plutus spending script
+--     which was included in the transaction body
+--
+module Marconi.Index.AddressDatum
+  ( -- * AddressDatumIndex
+    AddressDatumIndex
+  , AddressDatumHandle
+  , StorableEvent(..)
+  , StorableQuery(..)
+  , StorableResult(..)
+  , toAddressDatumIndexEvent
+  , AddressDatumQuery
+  , AddressDatumResult
+  , AddressDatumDepth (..)
+  , open
+  ) where
+
+import Cardano.Api qualified as C
+import Cardano.Api.Shelley qualified as C
+import Cardano.Ledger.Alonzo.TxWitness qualified as Ledger
+import Control.Applicative ((<|>))
+import Control.Monad (forM, forM_)
+import Data.Foldable (Foldable (foldl'), fold, toList)
+import Data.List.NonEmpty (NonEmpty)
+import Data.List.NonEmpty qualified as NonEmpty
+import Data.Map.Strict (Map)
+import Data.Map.Strict qualified as Map
+import Data.Maybe (catMaybes, listToMaybe, mapMaybe)
+import Data.Set (Set)
+import Data.Set qualified as Set
+import Database.SQLite.Simple qualified as SQL
+import Database.SQLite.Simple.ToField qualified as SQL
+import GHC.Generics (Generic)
+import Marconi.Orphans ()
+import RewindableIndex.Storable (Buffered (persistToStorage), HasPoint (getPoint),
+                                 QueryInterval (QEverything, QInterval), Queryable (queryStorage), Resumable,
+                                 Rewindable (rewindStorage), StorableEvent, StorableMonad, StorablePoint, StorableQuery,
+                                 StorableResult, emptyState, filterWithQueryInterval)
+import RewindableIndex.Storable qualified as Storable
+import Text.RawString.QQ (r)
+
+-- | Define the `handler` data type, meant as a wrapper for the connection type (in this case the
+-- SQLite connection). In this indexer, we also add the number of events that we want to return from
+-- the on-disk buffer.
+data AddressDatumHandle = AddressDatumHandle
+    { addressDatumHandleConnection :: SQL.Connection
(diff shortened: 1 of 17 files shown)


marconi/src/Marconi/Index/AddressDatum.hs
```diff
@@ -0,0 +1,517 @@
...
+-- the hash.
+-- In that last scenario, we can resolve in the datum hash in one of three ways (which we then simply
+-- add an entry in the `datumhash_datum` table):
+--
+--     * a different transaction output has an inline datum with that same hash
+--
+--     * a datum with that same hash has been found in the in transaction body
+--
+--     * a datum with that same hash was included in the witnesses for a Plutus spending script
+--     which was included in the transaction body
```
kayvank, line 50: Thank you for the detail, it helped a lot. 
I am a bit confused as to why we need this table as all the fields are already in the unspent_transaction table. I put more detail in the ticket, [PLT-662](https://input-output.atlassian.net/browse/PLT-662)

marconi/src/Marconi/Index/AddressDatum.hs
```diff
@@ -0,0 +1,517 @@
...
+-- database has to be able to accomodate them.
+
+-- We store the datum hashes of each address that we processed.
+-- Then we keep a separate 'Map' which stores the actual datum given a datum hash.
+-- Note that we don't always have the actual datum for a given hash.
+data instance StorableEvent AddressDatumHandle =
+    AddressDatumIndexEvent
+        (Map C.AddressAny (Set (C.Hash C.ScriptData)))
+        (Map (C.Hash C.ScriptData) C.ScriptData)
+        !C.ChainPoint
```
kayvank, line 112: newbie question :) 
I am not clear as to why we need strictness here?

marconi/src/Marconi/Index/AddressDatum.hs
```diff
@@ -0,0 +1,517 @@
...
+    -> C.ChainPoint
+    -> StorableEvent AddressDatumHandle
+toAddressDatumIndexEvent addressFilter txs chainPoint = do
+    let datumsPerAddr = getDatumsPerAddressFromTxs
+        filterFun =
+            case addressFilter of
+              Nothing -> id
+              Just f -> Map.filterWithKey $ \k _ ->
+                  case k of
+                      -- Target addresses filter are only shelley addresses. Therefore, is we
```
kayvank, line 188: Therefore, as we

marconi/src/Marconi/Index/AddressDatum.hs
```diff
@@ -0,0 +1,517 @@
...
+            , datum_hash BLOB NOT NULL
+            , slot_no INT NOT NULL
+            , block_hash BLOB NOT NULL
+            )|]
+    SQL.execute_ c
+        [r|CREATE TABLE IF NOT EXISTS datumhash_datum
+            ( datum_hash BLOB PRIMARY KEY
+            , datum BLOB
+            )|]
+    SQL.execute_ c
```
kayvank, line 513: Is it possible to have duplicate entries? should we add uniqueness constraint?

marconi/src/Marconi/Types.hs
```diff
@@ -48,8 +46,3 @@ type TxOutRef = C.TxIn
 
 txOutRef :: C.TxId -> C.TxIx -> C.TxIn
 txOutRef = C.TxIn
-
```
kayvank, line 51: with this remove, do we still nee the 
```{-# OPTIONS_GHC -Wno-orphans #-}```

File: marconi/src/Marconi/Index/AddressDatum.hs (hunk 2 above)
Comment (koslambrou) [reply marconi/src/Marconi/Index/AddressDatum.hs:112]: I always put strictness by default (that was a practice I read a while ago)

File: marconi/src/Marconi/Index/AddressDatum.hs (hunk 1 above)
Comment (koslambrou) [reply marconi/src/Marconi/Index/AddressDatum.hs:50]: Good point. Here’s what I see:

    Utxo indexer does not resolve the datum hashes as I expect them. The requirements of the Utxo indexer were simply that we store the datum hash and inline datum of the tx output

We could change the Utxo indexer to better resolve the datums. 

This seems definitely something we should try. My intuition with Address-Datum indexer is that queries should be faster because the table is smaller than the Utxo indexer. It would certainly make sense to compare them.

File: marconi/test/Spec/Marconi/Index/AddressDatum/Generators.hs [hunk 6]
```diff
@@ -0,0 +1,61 @@
...
+genAddressInEra :: C.CardanoEra era -> Gen (C.AddressInEra era)
+genAddressInEra era =
+  case C.cardanoEraStyle era of
+    C.LegacyByronEra ->
+      C.byronAddressInEra <$> CGen.genAddressByron
+
+    C.ShelleyBasedEra _ ->
+      Gen.choice
+        [ C.byronAddressInEra   <$> CGen.genAddressByron
+        -- , C.shelleyAddressInEra <$> CGen.genAddressShelley
```
raduom, line 31: Leftover?

marconi/src/Marconi/Index/AddressDatum.hs
```diff
@@ -0,0 +1,517 @@
...
+                ()
+        QInterval (C.ChainPoint b _) (C.ChainPoint e _) ->
+            SQL.query c
+                [r|SELECT address, address_datums.datum_hash, datumhash_datum.datum, slot_no, block_hash
+                   FROM address_datums
+                   LEFT JOIN datumhash_datum
+                   ON datumhash_datum.datum_hash = address_datums.datum_hash
+                   WHERE slot_no >= ? AND slot_no <= ?
+                   ORDER BY slot_no ASC, address, datumhash_datum.datum_hash|]
+                (b, e)
```
raduom, line 385: This is not correct, and is actually way more complicated than one would expect. You should only have `slot_no <= e` here and here is a convoluted explanation as to why :)

The abstract layout of the data (regardless of location) is something like: `e1, e2, e3, e4, acc`. When you give a query interval, what you are saying is try to find a good `e` in that interval and then give me all the `e`s that are lower than that.

This makes sense if you think that what you need to do here is compute `acc` at `e1, e2, e3,...`, but in order to do that you will always need *all* the events lower than the one you select.

Now, the question arises: Why have an interval at all? Why not just have some slot number that everyone synchronises on? The question would be that it *might* be useful for the client to assume a certain risk which sounds like: "if the indexers are within 3 slots of each-other I am fine with absorbing the risk of inconsistent blockchain views".

Context: https://github.com/input-output-hk/plutus-apps/blob/66799bd9c15fe125bb334501f48cf3ee1d740ec3/rewindable-index/src/RewindableIndex/Storable.hs#L318

In order for this function to work properly we need to select all of the DB events which are `<= end`. In retrospect the naming for the intervals is also confusing because slot numbers always grow, and intervals are specified as (low, high) the `end` is actually the latest, highest, top slot number. I find that kind of confusing. Hopefully it's just me.

If you rewrite this function like suggested you get the added bonus of not needing the case that follows it. :)

marconi/src/Marconi/Index/AddressDatum.hs
kayvank: Is it possible to have multiple identical rows, and is it OK to have them?

marconi/src/Marconi/Index/AddressDatum.hs
kayvank: Thnx @koslambrou . I am in favor of smaller tables as the UTXO table already had quite a few fileds.

marconi/src/Marconi/Index/AddressDatum.hs
koslambrou: Are you referring to the index table `address_datums_index`? If so, yes, one address can have multiple datums.

kayvank: right. my question however is with this  table structure, it is possible to have duplicate rows. Is that OK? and consequently, when we query, and get a `list` back, is it OK for the results set to have identical items?

koslambrou: I'm not seeing it. How can a query to `address_datums_index` result in duplicate rows? 
Querying an address will yield a list of datum hashes. Of course, there should no be any duplicate datum hashes for a given address. Should I change anything for that?

marconi/src/Marconi/Index/AddressDatum.hs
koslambrou: Changed to return an empty list instead for the time being. Will implement it correctly in a future PR once we get the ADR for the query interval.

koslambrou:
@raduom Anything more to add? Approve PR?

raduom:
@koslambrou You forgot to ask for review. I will take a look later.

marconi/src/Marconi/Index/AddressDatum.hs
kayvank: thnx, @koslambrou that was mostly my question that is it possible to have duplicate hashes, and the answer is no.

kayvank (approved):

raduom (approved):

PR #916, opened 2023-01-31
Fix datum encoding in marconi indexer

kayvank:
Changed datum encoding in JSON-RPC from printing the ScriptData complete text in JSON to CBOR presentation
[PLT-320]

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

[PLT-320]: https://input-output.atlassian.net/browse/PLT-320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/marconi/src/Marconi/Index/Utxo.hs
+++ b/marconi/src/Marconi/Index/Utxo.hs
@@ -297,22 +297,24 @@ open dbPath (Depth k) = do
                       ( address TEXT NOT NULL
                       , txId TEXT NOT NULL
                       , txIx INT NOT NULL
-                      , datum TEXT
-                      , datumHash TEXT
+                      , datum BLOB
+                      , datumHash BLOB
                       , value BLOB
-                      , inlineScript TEXT
-                      , inlineScriptHash TEXT
+                      , inlineScript BLOB
+                      , inlineScriptHash BLOB
                       , blockNo INT
                       , slotNo INT
                       , blockHash BLOB
                       , UNIQUE (txId, txIx))|]
   SQL.execute_ c [r|CREATE TABLE IF NOT EXISTS spent
-                      ( txInTxId TEXT PRIMARY KEY NOT NULL UNIQUE
+                      ( txInTxId TEXT PRIMARY KEY NOT NULL
                       , txInTxIx INT NOT NULL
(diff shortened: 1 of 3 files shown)


marconi/src/Marconi/Index/Utxo.hs
```diff
@@ -297,22 +297,24 @@ open dbPath (Depth k) = do
...
                       , blockHash BLOB
                       , UNIQUE (txId, txIx))|]
   SQL.execute_ c [r|CREATE TABLE IF NOT EXISTS spent
-                      ( txInTxId TEXT PRIMARY KEY NOT NULL UNIQUE
+                      ( txInTxId TEXT PRIMARY KEY NOT NULL
                       , txInTxIx INT NOT NULL
                       , slotNo INT NOT NULL
-                      , blockHash BLOB NOT NULL)|]
+                      , blockHash BLOB NOT NULL
+                      , UNIQUE (txInTxId, txInTxIx))|]
```
james-iohk, line 314: I remember one argument against this solution was that it would cover up a bug / inefficiency. Having said that, have no idea if there would be any noticeable trade off.

koslambrou (approved):

marconi/src/Marconi/Index/Utxo.hs
eyeinsky: This would prevent duplicates in the table, but given that the rows are added with `INSERT OR REPLACE` then it won't raise an exception. I wonder if using just `INSERT` would be better, as you'll get an exception when resuming isn't implemented correctly?

kayvank: It would not raise an exception, and I agree that on an incorrect resume, we wont get an exception. However, in such case, wouldn't the overriding row, be identical to the one what was already there?

eyeinsky: @kayvank The potential error is that you won't know when you are overwriting old data, and this might hurt when it happens accidentally.

PR #919, opened 2023-02-01
remove invalidScript test

berewt:
invalidScript check for a phase 2 error, something that should be catch by the validation and that does not seem to bring any value to our codebase, as it's out of the scope of our library. Removing it to save some time in the future (transition to Cardano Tx)

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


berewt:
@sjoerdvisscher we discuss this with @koslambrou during the planning of the transition to cardano-tx, it seems that the test is useless now that we moved to cardano-ledger validation, so removing it should be harmless.

sjoerdvisscher (approved):
Agreed!

PR #923, opened 2023-02-04
Removal of assetClassRow from chain-index DB

etiennejf:
The table assetClass consumes a lot of disk space and impacts sync time. All the necessary info to perform query `getUtxoSetWithCurrency` can be retrieved from unspentOutputRows and query getUtxoutFromRef, i.e., similar to what is actually being performed in query `getTxOutSetAtAddress`. Hence this PR performs the following modifications:

- [x] Removal of the assetClassRow from DbSchema
- [x] Rewrite `getUtxoSetWithCurrency`  to filter txouts obtained by coupling a query to fetch all unspent utxo with `getUtxouFromRef`.
- [x] Optimize `getTxOutSetAtAddress` to avoid extra loop on List.zip.


etiennejf:
Wrong PR created

PR #930, opened 2023-02-07
Raduom/datum ticks

raduom:
This function is used to add a checkpoint to the in-memory part of event indexers.
   You can use checkpoints to add synthetic events to the memory buffer. Filling the
   memory buffer will flush the real events to disk faster than not using checkpoints.

   This is useful if the number of events is small and you run the risk of never
   storing them on-disk. When that happens, if the indexer is restarted it will
   resume from the Genesis block, which is something that is not an expected
   behaviour.

   Possible future uses of checkpoints:

   * We currently assume that the indexer has processed blocks up to the latest event
     processed. In case there are very few events this assumption introduces a big
     error. We may eventually use checkpoints to signal to the indexer that the latest
     block is more recent than the latest event processed.

   * Currently synthetic events cannot be stored in the persistent database. We may
     decide to offer this option to implementers if there is interest. This would
     allow for faster resumes if the last stored event is a lot older than the
     currently processed block (similar to the in-memory description of checkpoints).

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/rewindable-index/src/RewindableIndex/Storable.hs
+++ b/rewindable-index/src/RewindableIndex/Storable.hs
@@ -20,11 +20,14 @@ module RewindableIndex.Storable
   , StorableMonad
     -- * API
   , QueryInterval(..)
+  , SyntheticEvent(..)
   , Buffered(..)
   , Queryable(..)
   , Resumable(..)
   , Rewindable(..)
   , HasPoint(..)
+  , syntheticPoint
+  , foldEvents
   , insert
   , insertMany
   , rewind
@@ -36,7 +39,7 @@ import Control.Applicative ((<|>))
 import Control.Lens.Operators ((%~), (.~), (^.))
 import Control.Lens.TH qualified as Lens
 import Control.Monad.Primitive (PrimMonad, PrimState)
-import Data.Foldable (foldlM)
+import Data.Foldable (foldl', foldlM)
 import Data.Function ((&))
 import Data.Functor ((<&>))
 import Data.Vector qualified as V
@@ -189,16 +192,31 @@ newtype Config = Config
   } deriving (Show, Eq)
 $(Lens.makeLenses ''Config)
 
+data SyntheticEvent e p =
+    Event     !e
+  | Synthetic !p
+
+syntheticPoint :: HasPoint e p => SyntheticEvent e p -> p
+syntheticPoint (Event e)     = getPoint e
+syntheticPoint (Synthetic p) = p
+
+foldEvents :: forall f h. Foldable f => f (SyntheticEvent (StorableEvent h) (StorablePoint h)) -> [StorableEvent h]
+foldEvents = reverse . foldl' unwrap []
+  where
+    unwrap :: [StorableEvent h] -> SyntheticEvent (StorableEvent h) (StorablePoint h) -> [StorableEvent h]
+    unwrap acc (Synthetic _) = acc
+    unwrap acc (Event e)     = e : acc
+
 data Storage h = Storage
-  { _events :: VM.MVector (PrimState (StorableMonad h)) (StorableEvent h)
-  , _cursor :: Int
+  { _events :: !(VM.MVector (PrimState (StorableMonad h)) (SyntheticEvent (StorableEvent h) (StorablePoint h)))
+  , _cursor :: !Int
   }
 $(Lens.makeLenses ''Storage)
 
 data State h = State
-  { _config  :: Config
-  , _storage :: Storage h
-  , _handle  :: h
+  { _config  :: !Config
+  , _storage :: !(Storage h)
+  , _handle  :: !h
   }
 $(Lens.makeLenses ''State)
 
@@ -220,7 +238,7 @@ emptyState memBuf hdl = do
 -- Get events from the memory buffer.
 getMemoryEvents
   :: Storage h
-  -> V.MVector (PrimState (StorableMonad h)) (StorableEvent h)
+  -> V.MVector (PrimState (StorableMonad h)) (SyntheticEvent (StorableEvent h) (StorablePoint h))
 getMemoryEvents s = VM.slice 0 (s ^. cursor) (s ^. events)
 
 -- Get events from memory buffer and disk buffer.
@@ -231,7 +249,7 @@ getEvents
   -> StorableMonad h [StorableEvent h]
 getEvents s = do
   memoryEs <- getMemoryEvents (s ^. storage)
-              & V.freeze <&> V.toList
+              & V.freeze <&> foldEvents . V.toList
   diskEs   <- getStoredEvents (s ^. handle)
   pure $ diskEs ++ memoryEs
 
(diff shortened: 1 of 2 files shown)


rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -189,16 +193,31 @@ newtype Config = Config
   } deriving (Show, Eq)
 $(Lens.makeLenses ''Config)
 
+data SyntheticEvent e p =
```
koslambrou, line 196: Haddock comment here. What is that used for and what do the `e` and `p` typically represent. Your PR description is a good starting point as it describes a lot of this :)

koslambrou, line 196: Why not `data SyntheticEvent h = Event !(StorableEvent h) | Synthetic !(StorablePoint h)`?

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -231,10 +250,42 @@ getEvents
...
 getEvents s = do
   memoryEs <- getMemoryEvents (s ^. storage)
-              & V.freeze <&> V.toList
+              & V.freeze <&> foldEvents . V.toList
   diskEs   <- getStoredEvents (s ^. handle)
   pure $ diskEs ++ memoryEs
 
+{- This function is used to add a checkpoint to the in-memory part of event indexers.
+   You can use checkpoints to add synthetic events to the memory buffer. Filling the
+   memory buffer will flush the real events to disk faster than not using checkpoints.
```
koslambrou, line 259: Do you mean instead

```suggestion
   memory buffer will flush the real events to disk faster than using checkpoints.
```

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -231,10 +250,42 @@ getEvents
...
+     processed. In case there are very few events this assumption introduces a big
+     error. We may eventually use checkpoints to signal to the indexer that the latest
+     block is more recent than the latest event processed.
+
+   * Currently synthetic events cannot be stored in the persistent database. We may
+     decide to offer this option to implementers if there is interest. This would
+     allow for faster resumes if the last stored event is a lot older than the
+     currently processed block (similar to the in-memory description of checkpoints).
+-}
+checkpoint
```
koslambrou, line 278: Looking at this function, seems like it will be up to the user to call this at different points in time while running the indexer right?

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -231,10 +250,42 @@ getEvents
...
+{- This function is used to add a checkpoint to the in-memory part of event indexers.
+   You can use checkpoints to add synthetic events to the memory buffer. Filling the
+   memory buffer will flush the real events to disk faster than not using checkpoints.
+
+   This is useful if the number of events is small and you run the risk of never
+   storing them on-disk. When that happens, if the indexer is restarted it will
+   resume from the Genesis block, which is something that is not an expected
+   behaviour.
+
+   Possible future uses of checkpoints:
```
koslambrou, line 266: Would it make sense to add a example using it in the `Spec/Sqlite` test module?

rewindable-index/src/RewindableIndex/Storable.hs
```diff
@@ -267,8 +318,8 @@ flushBuffer s = do
       mx = s ^. config . memoryBufferSize
   if mx == cr
   then do
-    v  <- V.freeze es
-    h' <- persistToStorage v (s ^. handle)
+    es' <- foldEvents <$> V.freeze es
+    h' <- persistToStorage es' (s ^. handle)
```
berewt, line 322: I know it's a breaking change but don't we want to also allow developer to persist synthetic event if they want to, just to track on disk how far the indexer went?

rewindable-index/src/RewindableIndex/Storable.hs
berewt, line 278: Shouldn't we have a function on top of `checkpoint` and `insert` that could process any `SyntheticEvent`

berewt (approved):
IIUC when the developer of an indexer should decide, when a block that doesn't contain an `Event` if it should call `checkpoint` to add a `SyntheticEvent` to the indexer.

Wouldn't it make sense to introduce a typeclass for indexers that would expose a function that process a `BlockInMode` to generate a `Maybe (SyntheticEvent (Event (StorableEvent h) (StorablePoint h)))`?

As I write this comment, I think that maybe this should be a part of another ticket where we could think about the Event generation and processing part.

rewindable-index/src/RewindableIndex/Storable.hs
raduom: Yes. I've had a discussion with @kayvank where I had the idea of putting the event filtering function within the indexer, in which case this function could be called automatically. However, we can make that change later if we choose to do so, so I did not include it in this PR.

rewindable-index/src/RewindableIndex/Storable.hs
raduom: No, but I think I could use better english :)

rewindable-index/src/RewindableIndex/Storable.hs
raduom: It complicates reasoning about the type-system when you need to specify further constraints in the `e` and `p` type variables, leading to the need to enable several ghc extensions. I don't see what exactly we would gain from that.

raduom: I added a comment on the `checkpoint` function. In theory this `SyntheticEvent` should not be shared.

rewindable-index/src/RewindableIndex/Storable.hs
raduom: I think the usage is fairly straight-forward. And I intend to base a lot more of the future improvements on this change. I would wait a bit until we write a new indexer specification so we get some more features in and then think about how we test them, what other properties we should add, etc.

rewindable-index/src/RewindableIndex/Storable.hs
raduom: I think we do. I am not sure about it though so I went for the least disruptive change.

rewindable-index/src/RewindableIndex/Storable.hs
raduom: I am not sure we want to couple indexers to a Cardano type. And if we make this implementation for any kind of type, then what we write is basically a `map` function, which I think is better kept outside of the indexer.

I am debating if we need to have a `filter` function within the indexer for similar reasons. The upside of a `filter` function would be that we can make the use of `checkpoint` automatic (checkpoint whatever the filter throws away).

berewt: Yes, `filter`/`processBlock` (or whatever would be its name) would respond to the same problem. Maybe an `IndexerInput` type family could answer these needs without being Cardano-specific, but it may be hard to type-check if we use it.

rewindable-index/src/RewindableIndex/Storable.hs
koslambrou: Ah, alright

koslambrou: :+1:

koslambrou (approved):

rewindable-index/src/RewindableIndex/Storable.hs
kayvank: I think that will be a nice changes. Doing the filtering as its done with the Utxos currently, is adding a lot of noise to the Indexers.hs module. +1 for me when this change comes in

PR #934, opened 2023-02-08
Add nix-shell to hydraJobs.nix

zeme-iohk:
We bumped HLS to 1.8.0, but new users found themselves building HLS 1.8.0 instead of pulling from cache.
Adding shell.nix itself to hydraJobs.nix should fix this issue

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -23,14 +23,18 @@ let
         narHash = "sha256-3Rnj/g3KLzOW5YSieqsUa9IF1Td22Eskk5KuVsOFgEQ=";
       };
   }).defaultNix;
-  cardano-node = import
-    (pkgs.fetchgit {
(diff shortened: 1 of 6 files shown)


shell.nix
```diff
@@ -23,14 +23,18 @@ let
...
-  cardano-node = import
-    (pkgs.fetchgit {
-      url = "https://github.com/input-output-hk/cardano-node";
-      # A standard release compatible with the cardano-wallet commit above is always preferred.
-      rev = "1.35.4";
-      sha256 = "1j01m2cp2vdcl26zx9xmipr551v3b2rz9kfn9ik8byfwj1z7652r";
-    })
-    { };
+
+  cardano-node = (import sources.flake-compat {
```
zeme-iohk, line 27: See the comment for `nix-pre-commit-hooks`

PR #939, opened 2023-02-09
Bump haskell.nix

andreabedini:
The bump is to support cabal file revisions on CHaP which are necessary to get our packages to build without all those manual constraints we have in cabal.project.

I opened a draft PR to test the waters and warm up the chaces ahead of time since it's relatively big change in haskell.nix. I'll make sure everything works and then mark it as ready.


zeme-iohk:
I think rebasing will fix the build failures.

andreabedini:
Everything seems to work :+1:

zeme-iohk (approved):

PR #943, opened 2023-02-09
Various small refactoring and fixes in `marconi-mamba`

koslambrou:
* Remove the `--utxo-db` CLI parameter in favor of `--utxo-db-fname` which takes the file name instead of the full path

* Added a CLI param (`--db-dir`) for specifiying the directory in which to store the SQLite database files.

* Providing target addresses in the CLI through `--addresses-to-index` is now optional

* Querying the UTXOs for a given address does not limit the address to the
  target addresses provided in the CLI.

* Fixed the resuming capability of the indexers in `marconi-mamba`

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-mamba/src/Marconi/Mamba/Api/Types.hs
+++ b/marconi-mamba/src/Marconi/Mamba/Api/Types.hs
@@ -34,11 +34,12 @@ import Marconi.ChainIndex.Types as Export (TargetAddresses)
 -- | Type represents http port for JSON-RPC
 
 data CliArgs = CliArgs
-  { socket          :: FilePath             -- ^ POSIX socket file to communicate with cardano node
-  , dbPath          :: FilePath             -- ^ filepath to local sqlite for utxo index table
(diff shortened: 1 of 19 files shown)


marconi-mamba/src/Marconi/Mamba/Api/Types.hs
```diff
@@ -60,8 +61,7 @@ data MambaEnv = MambaEnv
 makeClassy ''MambaEnv
 
 data QueryExceptions
-    = AddressNotInListError QueryExceptions
-    | AddressConversionError QueryExceptions
+    = AddressConversionError QueryExceptions
     | TxRefConversionError QueryExceptions
```
kayvank, line 65: this one is also no longer needed, my bad as I should've taken out when we switched to `unspent_transaction` table

kayvank (approved):

PR #952, opened 2023-02-14
PLT-229 Fixed resuming of UTXO indexer

koslambrou:
* Removed 'BlockNo' from the database stored on disk

* Removed the 'DROP TABLES' in the 'open' functions so that we can actually be able to resume at startup

* Fixed the UtxoEvent generator to generate unique utxos and only spend utxos which have been previously created.

* Added property tests for the resuming capabilities

Note: The diff notes some significant changes in newly created UtxoIndex module. However, most of the code inside is from the old test module. The only stuff I have added is the property tests for resuming + some small refactoring.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [x] Reviewer requested


marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -0,0 +1,314 @@
...
+  events <- forAll genUtxoEvents
+  (storedEvents :: [StorableEvent Utxo.UtxoHandle]) <-
+    (liftIO . Utxo.open ":memory:") (Utxo.Depth 10)
+     >>= liftIO . Storable.insertMany events
+     >>= liftIO . Storable.getEvents
+  Hedgehog.assert (equivalentLists storedEvents events)
+
+-- Insert Utxo events in storage, and retrieve the events by address
+--
+utxoInsertAndQueryTest :: Property
```
kayvank, line 132: should we comment out this too for now? I get a warning when building the project
```test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs:133:1: warning: [-Wunused-top-binds]
    Defined but not used: ‘utxoInsertAndQueryTest’
    |
133 | utxoInsertAndQueryTest = property $ do
```

kayvank (approved):

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
koslambrou: Thanks, forgot to comment it for now.

Hydra is also giving a failing test case in `marconi-mamba`. Going to fix it.

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -0,0 +1,314 @@
...
+-- Puporse of this test is to filter out utxos that have a different address than those in the TargetAddress list.
+eventsAtAddressTest :: Property
+eventsAtAddressTest = property $ do
+    event <- head <$> forAll genUtxoEvents
+    let (addresses :: [StorableQuery Utxo.UtxoHandle]) =
+          ( traversed %~ Utxo.UtxoAddress)
+          . nub
+          . toListOf (folded . Utxo.address)
+          . Utxo.ueUtxos
+          $ event
```
eyeinsky, line 179: I know the above is not changed in this PR, but the above could be written as `map (Utxo.UtxoAddress . Utxo._address) $ Set.toList $ Utxo.ueUtxos event` which would be much clearer.

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -0,0 +1,314 @@
...
+rewindTest = property $ do
+  event0 <- forAll $ genEventAtChainPoint C.ChainPointAtGenesis
+  event1 <- forAll $ genEventAtChainPoint (head chainpoints )
+  event2 <- forAll $ genEventAtChainPoint (chainpoints !! 1)
+  event3 <- forAll $ genEventAtChainPoint (chainpoints !! 2)
+  indexer <- liftIO $ Utxo.open ":memory:" (Utxo.Depth 2)
+             >>= liftIO . Storable.insertMany [event0, event1, event2, event3]
+  cps' <- liftIO . resume $ indexer
+  -- we should only see points up to depth: Genesis and the first insert
+  sort cps' ===  [C.ChainPointAtGenesis, head chainpoints]
```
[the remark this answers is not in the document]
koslambrou: Thanks! I'm going to remove this test, because the others I have introduced subsume this one.

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
koslambrou: Thanks :)

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -0,0 +1,314 @@
...
+
+    , testPropertyNamed
+        "marconi-utxo query-interval property"
+        "utxoQueryIntervalTest"
+        utxoQueryIntervalTest
+
+    , testPropertyNamed
+          "The points that indexer can be resumed from should return at least the genesis point"
+          "propResumingShouldReturnAtLeastTheGenesisPoint"
+          propResumingShouldReturnAtLeastTheGenesisPoint
```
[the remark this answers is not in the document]
eyeinsky: (Not sure if this is the case currently but) If one already knows the list of points resume is supposed to return, then doing `expectedPoints === gottenPoints` would compare both the content and the order in a single assertion. Otherwise agree that testing multiple models in a single test is confusing.

A related thought on resume, what if it would just return the optimal point to resume from, something like: from storage find *only* the latest event that is past the security parameter (and otherwise genesis)? Because if you always return both then during the query to node to find the block to continue syncing from, it wouldn't report an error if that latest event is not found and start from genesis instead. But you would want to know when your assumption about the security parameter is broken, right?

PR #958, opened 2023-02-14
Fix script-evaluation-dump.yml

zeme-iohk:


scripts/cardano-node.sh
```diff
@@ -21,21 +20,15 @@ node_config_files=(
...
-  wget -nc https://hydra.iohk.io/build/16607585/download/1/"$x" -P "$NODE_DIR"
-done
+# # Download config files
+# for x in "${node_config_files[@]}"; do
+#   wget -nc https://hydra.iohk.io/build/16607585/download/1/"$x" -P "$NODE_DIR"
+# done
 
 # Launch node
-"$NODE_BIN_DIR"/cardano-node run \
+cardano-node run \
```
zeme-iohk, line 31: I need a little help with this.
The hydra links are now broken, and they need to be replaced with moritz's.

However, I see here that we are downloading the 1.35.2 binary every time.
This is not necessary, as we have 1.35.4 in PATH from the shell.

However we still need to get the node_config_files.

Where can they be found? Must we rely on a hard-coded hydra path?
Aren't they included maybe in the cardano-node derivation? Maybe in share or etc?
Are they part of the cardano-node repo?
Can they instead be copy-pasted in this script folder?
Can someone take on fixing this script and making it work locally, and then I can make sure that it can run as a GHA?

@michaelpj @zliu41

zliu41: @zeme-iohk I'm trying to rebase the code on top of plutus-apps HEAD and make it work. This could take a few days.

I don't know the answer to your questions, in particular how to use a non-hardcoded hydra path.

> downloading the 1.35.2 binary every time.

The job needs to run on the same machine every time (since we don't want to re-sync the node every time), and if it does then it shouldn't need to download it every time.

zeme-iohk: Ok, we'll have to use a hard-coded hydra path for the config then, but again, `cardano-node` is already in PATH so there should be no need to download it (unless you really need the 1.35.2 binary, and not the 1.35.4)

zliu41: @zeme-iohk It's working for me locally now:
```
git checkout zliu41/dump-1.35.4

AWS_ACCESS_KEY_ID=<...> \
  AWS_SECRET_ACCESS_KEY=<...> \
  AWS_DEFAULT_REGION=us-east-1 \
  AWS_ENDPOINT_URL=https://s3.devx.iog.io \
  NODE_DIR=$HOME/mainnet \
  S3_DUMP_DIR=s3://plutus/mainnet-script-dump-1-35-4/ \
  LOCAL_DUMP_DIR=$HOME/mainnet-script-dump \
  ./scripts/create-dump.sh
```

It uses the `cardano-node` on the PATH. For the config files, it downloads them from `book.world.dev.cardano.org`, as instructed [here](https://developers.cardano.org/docs/get-started/running-cardano/).

The job used to run daily on buildkite for 3 hours. It would be great to do the same with GHA.

zeme-iohk: There are no new commits to the PR... am I missing something? 🤔

zeme-iohk: Also, are we using `mainnet-script-dump-1-35-4` or `mainnet-script-dump`?

.github/workflows/script-evaluation-dump.yml
```diff
@@ -33,3 +40,13 @@ jobs:
           S3_DUMP_DIR: s3://plutus/mainnet-script-dump/
           LOCAL_DUMP_DIR: ../mainnet-script-dump 
 
+      - name: Upload Dump
```
zeme-iohk, line 43: Also what about this step? Do we want to upload right after dumping? 
@zliu41 is the `./scripts/upload-event-dump.sh` up-to-date?

scripts/cardano-node.sh
zliu41: There's no PR. It is not mergeable until plutus-app `main` branch starts using a newer version of cardano-node and plutus.

Let's use `mainnet-script-dump-1-35-4` to make it clear the node version used to generate the files.

.github/workflows/script-evaluation-dump.yml
zliu41: The one in `zliu41/dump-1.35.4` should be up to date.

scripts/cardano-node.sh
zliu41: I.e., the implementation of the dump job is in branch `zliu41/dump-1.35.4`, which is the branch the GH action should use.

zeme-iohk: Got it, I'll check it out. thanks!

PR #960, opened 2023-02-15
Plt 1513 marconi fix lint errors warnings

kayvank:
fix: 

- marconi-chain-index lint errors/warnings
- golden tests for marconi-mamba
- golden tests for marconi-chain-index

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/marconi-chain-index/app/Main.hs
+++ b/marconi-chain-index/app/Main.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
--- a/marconi-chain-index/changelog.d/20230217_163000_kayvank_lint.md
+++ b/marconi-chain-index/changelog.d/20230217_163000_kayvank_lint.md
@@ -1,0 +1,7 @@
+### Added
+
+- Added golden tests for the CLI parameter, PLT-1513
+
+### Fixed
+
+- Fixed lint errors and warnings
--- a/marconi-chain-index/marconi-chain-index.cabal
+++ b/marconi-chain-index/marconi-chain-index.cabal
@@ -208,6 +208,7 @@ test-suite marconi-chain-index-test
   hs-source-dirs: test
   other-modules:
     Integration
+    Spec.Marconi.ChainIndex.CLI
     Spec.Marconi.ChainIndex.Indexers.AddressDatum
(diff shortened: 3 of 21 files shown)


koslambrou (approved):

marconi-chain-index/test/Spec/Golden/Cli/marconi-chain-index.help
```diff
@@ -0,0 +1,15 @@
+Missing:
```
eyeinsky, line 1: We discussed this before and seemed to agree on not needing to test cli output. Which kind of bugs do these kind of tests help to prevent?

kayvank: Yes, we discuss this before, however, we had less CLI parameters at that time, and no plan to update marconi README. 
We plan on updating the Marconi README  with additional information, including usage and CLI parameters. That section of docs will be a link to the golden-file(s),  and will get updated as we introduce/update  parameters. 
Additionally,  the golden files, allow us to see what the user's see as they use the software, which I think is valuable.

eyeinsky: I don't know.. Why not paste one of these into the readme then with common flags? Right now there are ~10 new files which look like tests but actually are just documentation. Are we going to link to all of these? Also, we-the-developers can "see what the users see" by running the program instead of looking into any of the file (and first we'd need to find the correct file).

kayvank: agreed that 10 are excessive. Will you be OK with keeping one, which I think is what you're suggesting,  and we could link the out put of that test to the README?  I would much prefer avoid having to remember cut/paste when we update the CLIs

koslambrou: @eyeinsky Golden files for the CLI are very useful for reviewing purposes. Whenever you make changes to the CLI, the reviewer can verify if anything has changed directly in Github, instead of running the software. User facing interface is very important and we need to make sure not to change anything unless absolutely required.

I'm fine with 10 IIF they each provide something valuable for verifying CLI changes.

koslambrou: In this specific case, maybe just 2 are necessary? One with `--help` to show the interface CLI interface, and one which shows an error message whenever we provided an invalid CLI argument?

PR #969, opened 2023-02-18
Add haskell-language-server-wrapper to the devshell

zeme-iohk:
VSCode lookups the `haskell-language-server-wrapper` and falls back to `haskell-language-server`.
So if the user has installed `haskell-language-server-wrapper` on their system, VSCode will find and use that instead of 
the `haskell-language-server` provided by the shell


james-iohk:
I get the same result on `main` and `zeme-iohk/fix-hls` branches.
Screenshot_20230220_114509

james-iohk (approved):
This now works for me, thanks :+1:

james-iohk (changes requested):
Update CONTRIBUTING.adoc important notice

zeme-iohk:
> Update CONTRIBUTING.adoc important notice

Unsure what you mean

james-iohk:
@zeme-iohk In CONTRIBUTING/adoc it reads
> This binary is called haskell-language-server, rather than haskell-language-server-wrapper, which is what some of the editor integrations expect.

This suggests there is still no `haskell-language-server-wrapper`

zeme-iohk:
Thank you, I've updated it

james-iohk (approved):

sjoerdvisscher (approved):

PR #979, opened 2023-02-21
Adapt helper functions in `marconi-chain-index` test-lib to support multiple era.

koslambrou:
Rational is that we currently need to work with an older version of `cardano-testnet` which only supports eras up to Alonzo. However, in those test cases, we want to create transaction with Babbage era. Thus, in order to limit duplication, this commit adapts the existing helper functions to work with any era.

I worked on this issue because it was needed for another coming PR about adding some new test cases which reuse the functionalities in the helper modules (like `emptyTxBodyContent`, but with Babbage era instead of Alonzo)

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


marconi-chain-index/test-lib/Helpers.hs
```diff
@@ -247,3 +291,16 @@ bimSlotNo (C.BlockInMode (C.Block (C.BlockHeader slotNo _ _blockNo) _txs) _era)
 
 bimBlockNo :: C.BlockInMode mode -> C.BlockNo
 bimBlockNo (C.BlockInMode (C.Block (C.BlockHeader _slotNo _ blockNo) _txs) _era) = blockNo
+
+-- | Should probably move in `cardano-api`.
+txFeesExplicitInShelleyBasedEra
+    :: C.ShelleyBasedEra era
+    -> C.TxFeesExplicitInEra era
+txFeesExplicitInShelleyBasedEra shelleyBased =
```
james-iohk, line 299: If you only accept Right from `C.txFeesExplicitInEra` (as done in e2e tests) then you don't need to maintain this case for future eras

marconi-chain-index/test-lib/Helpers.hs
```diff
@@ -84,16 +85,20 @@ readAs as path = do
   H.leftFailM . liftIO $ C.readFileTextEnvelope as path'
 
 -- | An empty transaction
-emptyTxBodyContent :: C.Lovelace -> C.ProtocolParameters -> C.TxBodyContent C.BuildTx C.AlonzoEra
-emptyTxBodyContent fee pparams = C.TxBodyContent
+emptyTxBodyContent
+    :: (C.TxValidityLowerBound era, C.TxValidityUpperBound era)
```
james-iohk, line 89: Reducing the number of args to this function is best to reduce code because not all tests need to set validity interval. Imo only pparams is necessary.

marconi-chain-index/test/Integration.hs
```diff
@@ -145,11 +145,11 @@ testIndex = H.integration $ (liftIO TN.setDarwinTmpdir >>) $ HE.runFinallies $ H
       C.NoStakeAddress :: C.Address C.ShelleyAddr
 
   (tx1in, C.TxOut _ v _ _) <- do
-    utxo <- TN.findUTxOByAddress localNodeConnectInfo address
+    utxo <- TN.findUTxOByAddress @C.AlonzoEra localNodeConnectInfo address
```
james-iohk, line 148: At some point, do we want this test to also run in Babbage era and up? Determining the era once at the start, or before, the test would be a step in the right direction. Same as the e2e tests do.

james-iohk:
>older version of cardano-testnet which only supports eras up to Alonzo.

The 1.35.4 `cardano-testnet` we use does support Babbage. We use it in e2e tests.

marconi-chain-index/test/Integration.hs
eyeinsky: @james-iohk From the e2e tests, Is [something like this](https://github.com/input-output-hk/plutus-apps/blob/main/plutus-e2e-tests/test/Spec/AlonzoFeatures.hs#L46) what you mean? I guess it is already possible to do it without changes to cardano-testnet package?

eyeinsky (approved):
:+1:

If we can, we should somehow join the effort with @james-iohk e2e testing, perhaps make plutus-e2e-tests have a library part and make marconi depend on that?

marconi-chain-index/test/Integration.hs
james-iohk: Yes, can get the `CardanoEra` from `TestnetOptions` in use and pass it to any build/query function that needs it.

james-iohk:
> If we can, we should somehow join the effort with @james-iohk e2e testing, perhaps make plutus-e2e-tests have a library part and make marconi depend on that?

Yeah, not saying the current e2e tests is the best way of doing things but it would make sense to share all we can.

marconi-chain-index/test-lib/Helpers.hs
koslambrou: We can't for a multi era implementation of `emptyTxBodyContent`. The reason is because the upper bound in Shelley era is mandatory, so we can't put an empty upper bound when creating a Shelley tx. 

We can discuss further on how to extend this, but one solution would be to create era specific modules which specialize the functions in this module.

koslambrou:
> The 1.35.4 cardano-testnet we use does support Babbage. We use it in e2e tests

@james-iohk I'll double-check, but I'm pretty sure that it does not. Yes, there is the Testnet.Babbage module, but I needed to use the Testnet.Cardano module in order to be able to run a testnet for multiple eras. The main reason was that the testnet helper in Marconi functions only worked on eras until Alonzo.

What's missing is the `Babbage` dataconstructor in the `Testnet.Cardano.Era` datatype.

marconi-chain-index/test-lib/Helpers.hs
james-iohk: Ok, because in the plutus e2e tests will never use Shelley Era so `C.validityNoUpperBoundSupportedInEra` is used by default.

james-iohk:
> What's missing is the `Babbage` dataconstructor in the `Testnet.Cardano.Era` datatype.

True. This was one reason why e2e tests need the slightly modified `cardano-testnet`. If marconi only needs Alonzo and Babbage then we can use this once #977 is merged
https://github.com/input-output-hk/plutus-apps/blob/56ccc5f8ce902904ce59f3f02995924e33289e9f/plutus-e2e-tests/test/CardanoTestnet.hs#L67

marconi-chain-index/test-lib/Helpers.hs
koslambrou: The reason I created this was exactly for not dealing with an Either in my code. Since I'm working with ShelleyBasedEra, then `txFeesExplicitInShelleyBasedEra` is total compared to `txFeesExplicitInEra`.

james-iohk (approved):

PR #986, opened 2023-02-24
Write the synchronisation ADR.

raduom:
This ADR describes a method of synchronising queries across multiple indexers.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
+++ b/doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
@@ -1,0 +1,72 @@
+
+.. synchronising_queries_across_indexers_running_in_parallel:
+
+ADR 18: Synchronising queries across indexers running in parallel
+=================================================================
+
+Date: 2023-02-24
+
+Authors
+-------
+
+Ometita Radu - Adrian <radu.ometita@iohk.io>
+
+Status
+------
+
+Draft
+
+Context
+-------
+
+Running two indexers in parallel creates the implicit assumption that both indexers are answering queries over the same original (blockchain) data. That is not generally true due to the different speeds at which indexers process blocks.
+
+This has been observed first when the PAB needed to index some blockchain information and a smart contract would query both the PAB and the Chain Index. It would often be the case that their views were not synchronised, requiring the developer to synchronise the two views, which is not a trivial thing to do.
+
+
+Decision
+--------
+
+1. Change the `Query` type to include a preferred slot number and how many results we want to receive. If there is no preferred slot number, the top available slot number is assumed.
+2. Change the `Response` type to include all the requested query results encoded as the full query result, followed by a set of deltas (events).
+3. [Optional] Add a new function to the `Queryable` type class that witnesses the folding of Events into the query result.
+
+Example:
+
+Give a request for slot number 10, with a maximum of 5 results, some of the indexers might output:
+
+Indexer 01: State at slot 6, Delta for slots 7, 8, 9 and 10 (if the indexer is synchronized up to or past slot 10 and has at least 5 more deltas stored as events)
+Indexer 02: State at slot 6, Delta for slots 7 and 8 (if the indexer is only synchronised up to slot 8)
+Indexer 03: State at slot 8, Delta for slots 9 and 10 (if the indexer is synchronised up to or past slot 10, but only has events for slots 9 and 10 stored in the database)
+Indexer 04: No result (the indexer is not synchronised to slot 6)
+Best result (defined as the oldest available slot or all indexers), if calculated for indexers 1, 2 and 3 would be the data at slot 8. If we include the 4th indexer, then there is no result available as the 4th indexer is still synchronising.
+
+Argument
+--------
+
+A fairly simple solution to the synchronisation problem would be to qualify responses by the current slot number. This is a reasonable first thought, but it does have a high probability of returning results that are qualified by different slot numbers (if the indexers are fully synchronised, it would be a small offset, but still dependent on the various indexers processing speeds). The advantage of this solution is that at least, we now know that the indexers were not fully synchronised. If the application does not require them to be, then this is a reasonable solution.
+
+If, however, the application requires all queries to run at the same slot number, this solution will simply not function. What we could do is request multiple query responses from indexers in the hope that we can find a common slot number where all indexers have an answer. Is this hope realistic? I would argue that it is realistic for indexers that have been fully synchronised with the blockchain since the processed slot offset between them should be fairly small.
+
+Now that will solve our functional problem, but it may introduce some non-functional ones. First, the query results may be fairly big, and returning more of them will put pressure on the network and the parsing of data by the client applications. This can be partially mitigated by selecting a smaller interval, but maybe we can help out here by improving the design a bit. One observation is that we don't actually need to send the full query results. We need to send a query result for the oldest slot number available, but for the newer ones, we only need to send the difference recorded by the next slot. This implementation maps very well to the way we store information in the indexer (an accumulator and a set of events).
+
+I think this solution is better than what we currently have because it gives a principled way of composing queries over a set of indexers that run at different speeds. This allows us to run them in parallel on the same processor (as we do now) or distributed. It allows for efficient use of resources since we don't need to use any locking. The implementation is also simple since we don't share any data across threads and we don't need to deal with thead synchronisation issues.
+
+Implications
+------------
+
+However, this solution also complicates the life of our users. Now, they need to also have some function that allows them to fold these differences into the older query responses. I would suggest a couple of strategies that we could employ to simplify their lives, that are not the object of this ADR:
+
+1. Write a set of property-based tests that validate that the folding function is correct and instruct our users to verify their implementation using our acceptance tests.
+2. Develop a new software service that can synchronise queries across several indexers. This service would use our implementation of the folding function (which is verified by our tests) and after receiving all the query responses will select the best ones and provide them to the users already folded.
+
+The only thing remaining is to note that if we want to protect against rollbacks we will also need to store the block hash together with the slot number and at this point our design is complete.
+
+Alternatives
+------------
+
+Currently, the way we almost avoid that situation is to run a set of indexers in parallel by waiting on all of them to finish processing the current block before advancing to the next one. This is a fairly big improvement, but essentially it still has the same problem, except that it will not be triggered as often. One scenario when this could happen is if the query comes while the indexers are about to process a block. Some of these indexers may answer the query before block processing, and some after the block has been processed. This can be mitigated by a query-locking mechanism that spans all the indexers, but this adds complexity while not managing to solve the root problem.
+
+This solution also forgoes the possibility of splitting up the indexers into different processes or machines. Our customers complained about the excessive resource consumption, performance and the inability to easily add their own indexers to the chain index. Giving users the possibility of distributing their indexers will partially address the first concern by splitting resource consumption on several machines (which is also addressed at the indexer design level by an inherently scalable design) and indirectly the second (since indexers are distributable, they have a simple, standard API that allows any customer to roll their own version, implicitly solving the hard problems like rollbacks and synchronisation).
+
+Another alternative, one that assumes that we have indexer notifications implemented would be to have the client listen to block processed notifications from indexers and automatically compute the optimal slot at which a query can be run. In this case we would mitigate the need to have several results returned and no other real downside.
--- a/doc/adr/index.rst
+++ b/doc/adr/index.rst
@@ -47,3 +47,4 @@ The general process for creating an ADR is:
    0015-time-conversion
    0016-cardano-node-emulator
    0017-end-to-end-testing-strategy
+   0018-synchronising-queries-across-indexers-running-in-parallel


doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
```diff
@@ -0,0 +1,51 @@
...
+
+This solution also forgoes the possibility of splitting up the indexers into different processes or machines. Our customers complained about the excessive resource consumption, performance and the inability to easily add their own indexers to the chain index. Giving users the possibility of distributing their indexers will partially address the first concern by splitting resource consumption on several machines (which is also addressed at the indexer design level by an inherently scalable design) and indirectly the second (since indexers are distributable, they have a simple, standard API that allows any customer to roll their own version, implicitly solving the hard problems like rollbacks and synchronisation).
+
+Decision
+--------
+
+1. Change the `Request` type to include a preferred slot number and how many results we want to receive.
```
berewt, line 33: It's unclear how the results will be constructed from this request. If I request `n` results with a preferred slot `s`, would all the result I received be at most at slot `s` or can I receive results that are after slot `s`? Also I think the preferred slot should be optional, as in "give me the `n` most recent results you have".

raduom: Good points. I will update.

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
```diff
@@ -0,0 +1,51 @@
...
+
+However, this solution also complicates the life of our users. Now, they need to also have some function that allows them to fold these differences into the older query responses. I would suggest a couple of strategies that we could employ to simplify their lives, that are not the object of this ADR:
+
+1. Write a set of property-based tests that validate that the folding function is correct and instruct our users to verify their implementation using our acceptance tests.
```
berewt, line 48: It may be challenging as we'll have to adapt the property tests to the user programing language.

berewt (approved):
LGTM (aside from the minor comments I've made)

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
raduom: They both are challenging, but thing REST apis. I am not sure it would be that difficult to implement.

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Now that will solve our functional problem, but it may introduce some non-functional ones. First, the query results may be fairly big, and returning more of them will put pressure on the network and the parsing of data by the client applications. This can be partially mitigated by selecting a smaller interval, but maybe we can help out here by improving the design a bit. One observation is that we don't actually need to send the full query results. We need to send a query result for the oldest slot number available, but for the newer ones, we only need to send the difference recorded by the next slot. This implementation maps very well to the way we store information in the indexer (an accumulator and a set of events).
```
kayvank, line 44: Should this paragraph be under the `Implications (or consequences)` of the[ ADR-template](https://github.com/input-output-hk/plutus-apps/blob/main/doc/adr/0001-record-architecture-decisions.rst#adr-template)

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
raduom: I added an example that shows most (all?) possible indexer responses.

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
```diff
@@ -0,0 +1,51 @@
...
+------
+
+Draft
+
+Context
+-------
+
+Running two indexers in parallel creates the implicit assumption that both indexers are answering queries over the same original (blockchain) data. That is not generally true due to the different speeds at which indexers process blocks.
+
+This has been observed first when the PAB needed to index some blockchain information and a smart contract would query both the PAB and the Chain Index. It would often be the case that their views were not synchronised, requiring the developer to synchronise the two views, which is not a trivial thing to do.
```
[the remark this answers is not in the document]
raduom: They are both correct actually. I think one is used in American english the other in UK.

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
raduom: I think this still describes the solution and the implementation. I am not sure why this would be classified as an implication. Can you, please, elaborate?

raduom: Actually I think I understand what you meant. I added an `Implications` heading after this line.

kayvank (approved):

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
```diff
@@ -0,0 +1,65 @@
...
+
+This solution also forgoes the possibility of splitting up the indexers into different processes or machines. Our customers complained about the excessive resource consumption, performance and the inability to easily add their own indexers to the chain index. Giving users the possibility of distributing their indexers will partially address the first concern by splitting resource consumption on several machines (which is also addressed at the indexer design level by an inherently scalable design) and indirectly the second (since indexers are distributable, they have a simple, standard API that allows any customer to roll their own version, implicitly solving the hard problems like rollbacks and synchronisation).
+
+Decision
+--------
+
+1. Change the `Request` type to include a preferred slot number and how many results we want to receive. If there is no preferred slot number, the top available slot number is assumed.
+2. Change the `Response` type to include all the requested query results encoded as the full query result, followed by a set od deltas (events).
```
koslambrou, line 34: od -> of

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
```diff
@@ -0,0 +1,65 @@
...
+
+This solution also forgoes the possibility of splitting up the indexers into different processes or machines. Our customers complained about the excessive resource consumption, performance and the inability to easily add their own indexers to the chain index. Giving users the possibility of distributing their indexers will partially address the first concern by splitting resource consumption on several machines (which is also addressed at the indexer design level by an inherently scalable design) and indirectly the second (since indexers are distributable, they have a simple, standard API that allows any customer to roll their own version, implicitly solving the hard problems like rollbacks and synchronisation).
+
+Decision
+--------
+
+1. Change the `Request` type to include a preferred slot number and how many results we want to receive. If there is no preferred slot number, the top available slot number is assumed.
```
koslambrou, line 33: Why include in the request the number of results we want to receive?

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
```diff
@@ -0,0 +1,65 @@
...
+
+Implications
+------------
+
+However, this solution also complicates the life of our users. Now, they need to also have some function that allows them to fold these differences into the older query responses. I would suggest a couple of strategies that we could employ to simplify their lives, that are not the object of this ADR:
+
+1. Write a set of property-based tests that validate that the folding function is correct and instruct our users to verify their implementation using our acceptance tests.
+2. Develop a new software service that can synchronise queries across several indexers. This service would use our implementation of the folding function (which is verified by our tests) and after receiving all the query responses will select the best ones and provide them to the users already folded.
+
+The only thing remaining is to note that if we want to protect against rollbacks we will also need to store the block hash together with the slot number and at this point our design is complete.
```
koslambrou, line 65: I would add an _Alternative_ section explaining why we have just make sure that all indexers index at the same speed (like we currently have in `marconi-chain-index`) and explain downsides of this approach vs what you're proposing.

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
```diff
@@ -0,0 +1,65 @@
...
+Indexer 02: State at slot 6, Delta for slots 7 and 8 (if the indexer is only synchronised up to slot 8)
+Indexer 03: State at slot 8, Delta for slots 9 and 10 (if the indexer is synchronised up to or past slot 10, but only has events for slots 9 and 10 stored in the database)
+Indexer 04: No result (the indexer is not synchronised to slot 6)
+Best result (defined as the oldest available slot or all indexers), if calculated for indexers 1, 2 and 3 would be the data at slot 8. If we include the 4th indexer, then there is no result available as the 4th indexer is still synchronising.
+
+
+Argument
+--------
+
+A fairly simple solution to the synchronisation problem would be to qualify responses by the current slot number. This is a reasonable first thought, but it does have a high probability of returning results that are qualified by different slot numbers (if the indexers are fully synchronised, it would be a small offset, but still dependent on the various indexers processing speeds). The advantage of this solution is that at least, we now know that the indexers were not fully synchronised. If the application does not require them to be, then this is a reasonable solution.
```
koslambrou, line 51: Similar to the comment about the semaphor, I would add here an argument explaining how the chosen approach is better. Might seem obvious, but it needs to be clearly stated.

koslambrou (commented):
Very clear and well-written. Just some minor comments.

berewt:
I'm sick, so I couldn't sleep and thought about this ADR (yes, it's as odd as it sounds). And I was wondering if we aren't over-architecturing it.
The main issue we want to address is desynchronising the different indexer. As said by Radu, it's already partially managed by the coordinator but indexer may be processing blocks at a different speed leading to an off-by-one error, or we can query them at a slightly different time, leading as well to a difference in the latest processed slot.

I'd like us to consider the use of the coordinator to ensure that we're querying the indexers at a slot they have all processed. It's basically considering the coordinator itself as an indexer, capable of answering a simple request: "what is the last slot that is fully indexed". Users should pass through this query first if they want to query different indexers, and we can keep sending a unique result, which will be synced with other indexers by design.

I think it makes sense, but it may just be the lack of sleep.

raduom:
> I'm sick, so I couldn't sleep and thought about this ADR (yes, it's as odd as it sounds). And I was wondering if we aren't over-architecturing it. The main issue we want to address is desynchronising the different indexer. As said by Radu, it's already partially managed by the coordinator but indexer may be processing blocks at a different speed leading to an off-by-one error, or we can query them at a slightly different time, leading as well to a difference in the latest processed slot.
> 
> I'd like us to consider the use of the coordinator to ensure that we're querying the indexers at a slot they have all processed. It's basically considering the coordinator itself as an indexer, capable of answering a simple request: "what is the last slot that is fully indexed". Users should pass through this query first if they want to query different indexers, and we can keep sending a unique result, which will be synced with other indexers by design.
> 
> I think it makes sense, but it may just be the lack of sleep.

I am not sure if it is over-architecting. I am not even sure what that means. Let's assume that it means that it offers a solution to problems we don't really have right now.

What this change does:
Allows for SAFELY deploying indexers in parallel for the purpose of modularity, scalability and managing resource consumption. Technically we don't need this, no one asked for it in an official capacity. In my personal discussions with clients they said that the chain-index / PAB is difficult to extend and it needs a lot of resources to run.

If we wanted to kill all parallel computation / queries etc.; How could we do it?

Have the `coordinator` synchronise inserts and queries on some mutex. That would pretty much do it. If people decide that is all we need for now, then I suspect it can be done fairly easily. This is even simpler than using a coordinator, which would be another software component. Besides, even if we use a coordinator there would have to be changes to indexers to report the curent slot, etc.

What we would loose is any scenario where we have a good story for modularity, scalability and resource control. But arguably maybe we don't really care about those.

So I don't know :)

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
raduom: I am assuming that different applications will have different synchronisation requirements, so putting that number in the query seems like a good way to provide that.

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
raduom: I moved the `Currently...` paragraph to an alternatives section.

berewt:
> I am not sure if it is over-architecting. I am not even sure what that means. Let's assume that it means that it offers a solution to problems we don't really have right now.

I meant: providing an API that is too complex to handle on the client side.

> What this change does: Allows for SAFELY deploying indexers in parallel for the purpose of modularity, scalability and managing resource consumption. Technically we don't need this, no one asked for it in an official capacity. In my personal discussions with clients they said that the chain-index / PAB is difficult to extend and it needs a lot of resources to run.
> 
> If we wanted to kill all parallel computation / queries etc.; How could we do it?
> 
> Have the `coordinator` synchronise inserts and queries on some mutex. That would pretty much do it. If people decide that is all we need for now, then I suspect it can be done fairly easily. This is even simpler than using a coordinator, which would be another software component. Besides, even if we use a coordinator there would have to be changes to indexers to report the curent slot, etc.

That's the point, I didn't talk about mutex here, and I don't even think we need to synchronise on insert. Let's suppose that each indexer exposes the last slot it has processed (or notify an external component each time it has done with a block), we could query the last common indexed slot for our indexers (or a subset of it) first and then query our indexers for this specific slot (or any slot of interest prior to it).

Compared to the current approach of this ADR, I see one pro and one con:

- pro: we don't have to rebuild the aggregation on the client side;
- con: we have to request the last common slot first.

(I may miss a lot of pros and a lot of cons)

The good thing is that both solution are quite lightweight on server side, we can easily switch from one solution to the other, or even develop and maintain both of them.

> So I don't know :)

Neither do I :)

raduom:
I think one of the cons I see is that now you have a 2 step query.

Step 1. Get the latest set of processed blocks, and
Step 2. Query them.

This is a bit more complicated than having one query to rule them all, but there is less data serialised and transferred. I see your point. It's not a bad idea.

Things would be much easier if we had the notification system designed for indexers, then you could listen to notifications and figure out the `excelllentest` block number to query at.

Maybe we could even look at this solution as a temporary one, until notifications are added, or we could just add notifications.

raduom:
I have added the idea of having notifications inform the client of the last processed block to the ADR.

berewt:
Yes, that's what I meant with my con as well. 

But with the current design with the coordinator, we can avoid it if we keep the last fully processed slot in the coordinator. Suppose that you want to access the latest available info for indexer `x`, `y` and `z`. You ask for the latest info for `x`, the (one and unique) request pass through the coordinator to get the latest fully synced slot `s` and then forwards the request for this slot to `x`. The answer, along with slot `s`, is sent to the client. The client can then safely query `y` and `z` for slot `s`.

Basically, we providing something alike `query :: Maybe slot -> Query indexer -> (slot, Result indexer)` would be good enough.
What we're losing then is the opportunity to ask all the indexers in parallel, instead, we have to ask a first indexer alone and then the rest of them.

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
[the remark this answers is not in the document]
james-iohk: I'd argue only one is correct :trollface:

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
```diff
@@ -0,0 +1,72 @@
...
+
+This solution also forgoes the possibility of splitting up the indexers into different processes or machines. Our customers complained about the excessive resource consumption, performance and the inability to easily add their own indexers to the chain index. Giving users the possibility of distributing their indexers will partially address the first concern by splitting resource consumption on several machines (which is also addressed at the indexer design level by an inherently scalable design) and indirectly the second (since indexers are distributable, they have a simple, standard API that allows any customer to roll their own version, implicitly solving the hard problems like rollbacks and synchronisation).
```
eyeinsky, line 70: forgoes => allows?

AFAIK the thing that takes time is building up the index, and after that the indexers just tick along one block at a time while consuming little resources.

I would also contest the idea that rollbacks are a hard problem (to us), as to me it seems you just roll back (= delete newer rows from the sqlite) and that is the end of it, this is what we do, no?

It seems to me that it is our clients instead who are the ones that *will continue to have the hard problem* of choosing how bleeding edge they want to be (= how often they will get rollbacks, the probability declines over time) and how will they handle them. But the choice is theirs, no?

E.g, for being totally bleeding edge I guess the solution is some kind of layer 2 like maybe roll-ups. Say if Djed wants to do some atomic thing, then even though that thing might look like multiple things to them, they still collect it all into a single transaction and submit that. That single transaction will either succeed or be rolled back, but in both cases the final state is consistent (this being not much different in how relational databases have transactions).

I wonder how bleeding edge either Mamba or Djed wants to be, and if they are totally bleeding edge then what do they do when they get a rollback? @koslambrou

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
koslambrou: Ah so you're not talking about a result limit (like the `LIMIT` SQL keyword)? Maybe paraphrase?

doc/adr/0018-synchronising-queries-across-indexers-running-in-parallel.rst
koslambrou: For now, the Sidechain team doesn't need bleeding edge. They actually only the stable blocks, so they don't care for rollbacks. For Djed, they are bleeding edge, but don't care for rollbacks because their contracts are one-shot (the transaction will just fail in the user's interface)

PR #1004, opened 2023-03-01
Notifications - Model, implementation and tests.

raduom:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-core/src/Marconi/Core/TracedStorable.hs
+++ b/marconi-core/src/Marconi/Core/TracedStorable.hs
@@ -1,0 +1,391 @@
+module Marconi.Core.TracedStorable
+  ( -- * State
+    Config
+  , memoryBufferSize
+  , State
+  , handle
+  , config
+  , emptyState
+  , Storage
+  , storage
+  , events
+  , cursor
+  , getMemoryEvents
+  , getEvents
+  , StorableEvent
+  , StorablePoint
+  , StorableQuery
+  , StorableResult
+  , StorableMonad
+    -- * API
+  , SyntheticEvent(..)
+  , ControlNotification(..)
+  , Buffered(..)
+  , Queryable(..)
+  , Resumable(..)
+  , Rewindable(..)
+  , HasPoint(..)
+  , syntheticPoint
+  , foldEvents
+  , insert
+  , checkpoint
+  , insertMany
+  , rewind
+  , resume
+  , query
+  ) where
+
+import Control.Lens.Operators ((%~), (.~), (^.))
+import Control.Lens.TH qualified as Lens
+import Control.Monad (void)
+import Control.Monad.Primitive (PrimMonad, PrimState)
+import Data.Foldable (foldl', foldlM)
+import Data.Function ((&))
+import Data.Functor ((<&>))
+import Data.Vector qualified as V
+import Data.Vector.Generic qualified as VG
+import Data.Vector.Mutable qualified as VM
+import GHC.Generics (Generic)
+
+import Control.Tracer (Tracer, traceWith)
+
+{-
+   The extensible parts of the indexers are the way data is stored into some form
+   of persistent storage. The interface used for working blockchain events is defined
+   in this module.
+
+   There are quite a bit of type variables that are involed in defining this interface.
+   Some of them stand for the type of database connection (h), the type of events (e),
+   the type of points along the blockchain (denoted by p), the type of queries (q) and
+   results.
+
+   The following data families implement the observation that most of the variables
+   can be derived from the way information is stored in the database, so there is
+   quite a bit of convenience in reducing the number of type variables to just two.
+   One that stands for the database connection (which should be a newtype for each
+   indexer) and one for the monad in which the indexer runs (usually IO).
+-}
+data family StorableEvent h
+
+-- | The resume and query functionality requires a way to specify points on the chain from which we
+-- want to resume, or points up to which we want to query.
+type family StorablePoint h
+
+data family StorableQuery h
+
+data family StorableResult h
+
+-- TODO: Rename `Storable` to `Indexer`.
+data family StorableNotifications h
+
+data ControlNotification h =
+    CNRollForward !(StorablePoint h)
+  | CNRollBack    !(StorablePoint h)
+  | CNApplication !(StorableNotifications h)
+  deriving (Generic)
+
+type family StorableMonad h :: * -> *
+
+{-
+   The first, `Buffered` class explains what it means for an indexer to be accumulating
+   a set of results that will be flushed to disk when the memory buffer is fully filled.
+
+   The memory layout of the indexer has been simplified by a lot since the previous
+   version and it looks something like this:
+
+   |e|e|e|e|e|e|e|e|e|e|e|e|e|e|e|e|e|e|                                     |
+   |-------------------|---------------|-------------------------------------|
+   |   memory/buffer   |  disk/events  |       disk/aggregate (optional)     |
+
+   When the buffer is filled with events, they all get flushed to disk. We need this
+   operation for performance reasons. Most databases will have significantly improved
+   performance for batch inserts.
+
+   Rollbacks will only happen in the memory or disk/events storage area, so the
+   developer has to make sure that the allocated space for the buffer and the
+   disk/events is greater than the K parameter.
+
+   The last part disk/aggregate represents some aggregated data. At some point the
+   developer should write code to aggregate the disk/events into the aggregate
+   section of the database. We did not provide any special handling of this for the
+   following reasons:
+
+     1) A lot of indexers will not require aggregated data. Only storing events should
+        be enough for most applications.
+     2) Data aggregation can be implemented at database level and probably be more
+        efficient than what Haskell can do here.
+     3) We can always extend the interface to include direct support for this pattern
+        if there is a demand from the users of the API.
+-}
+class Buffered h where
+  -- | This function persists the memory/buffer events to disk when the memory buffer is filled.
+  persistToStorage :: Foldable f => f (StorableEvent h) -> h -> StorableMonad h h
+
+  {- This function retrieves the events from the disk/events area.
+     If the user chooses to only store events, without accumulating them, this function
+     is expected to return the events over which rollbacks can occur in order to keep
+     things performant -}
+  getStoredEvents :: h -> StorableMonad h [StorableEvent h]
+
+{-
+   All information from indexers should be made accessible through queries. Queries
+   act a lot like folds over the event stream. Each query introduces two indexed types:
+   one for the type of requests (called StorableQuery) and one for responses (called
+   StorableResult).
+
+   All queries include a validity interval. If the data is not available for a specified
+   interval, the returned result should specify that. It is also recommended that the result
+   includes the slot number at which the query was ran.
+-}
+class Queryable h where
+  queryStorage
+    :: Foldable f
+    => StorablePoint h
+    -> f (StorableEvent h)
+    -> h
+    -> StorableQuery h
+    -> StorableMonad h (StorableResult h)
+
+{-
+   One of the reasons why indexers were born was to solve one of the issues very specific
+   to blockchains: rollbacks. The basic idea was simple, keep a history of the blockchain
+   data and whenever we have a rollback restore the version which is the target of the
+   rollback.
+
+   This is what the next class is meant to do.
+-}
+class Rewindable h where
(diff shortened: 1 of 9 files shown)


marconi-core/src/Marconi/Core/TracedStorable.hs
```diff
@@ -0,0 +1,391 @@
...
+data family StorableResult h
+
+-- TODO: Rename `Storable` to `Indexer`.
+-- data family StorableNotifications h
+
+data ControlNotification pt =
+    CNRollForward !pt
+  | CNRollBack    !pt
+ ---- | CNApplication !(StorableNotifications h)
+  deriving (Generic)
```
berewt, line 85: Why is this one indexed with `pt` instead of `h`? Is it because of the type family for point ? My gut feeling is that adding the `CNAppplication` won't be trivial, I wonder if we should assess it now or postpone it. I think we'll have to add it somehow, but we may live without it in a near future.

marconi-core/src/Marconi/Core/TracedStorable.hs
```diff
@@ -0,0 +1,391 @@
...
+   StorableResult).
+
+   All queries include a validity interval. If the data is not available for a specified
+   interval, the returned result should specify that. It is also recommended that the result
+   includes the slot number at which the query was ran.
+-}
+class Queryable h where
+  queryStorage
+    :: Foldable f
+    => StorablePoint h
```
berewt, line 143: Why not `Maybe (StorablePoint h)` here, to allow a simple "give me the freshest thing you have" that may be suitable in some scenarios.

marconi-core/src/Marconi/Core/TracedStorable.hs
```diff
@@ -0,0 +1,391 @@
...
+    -> f (StorableEvent h)
+    -> h
+    -> StorableQuery h
+    -> StorableMonad h (StorableResult h)
```
berewt, line 147: If we allow `Maybe`, as suggested in the previous comment, we may want to add the latest sync `StorablePoint` in the result. So... maybe my previous suggestion is a bad idea and we should instead have a second function in the `Queryable` interface: `queryStorageforNewest  :: Foldable f => f (StorableEvent h) -> h -> StorableQuery h -> StorableMonad h (StorablePoint h, Storable:Result h)`

marconi-core/test/Marconi/Core/Spec/TracedSqlite.hs
```diff
@@ -0,0 +1,372 @@
...
+       then pure . Just $ sePoint $ es' !! n
+       -- If we don't have more events stored in the database, then we simply rollback to
+       -- Genesis.
+       else pure . Just $ Genesis
+  else if length es > n
+  -- If the point we want to rollback to is less then the size of the list of stored
+  -- events, then we want to rollback to the indexed event.
+  then pure $ atMay es' n <&> sePoint
+  -- Otherwise the point we need to rollback to was not found.
+  else pure Nothing
```
berewt, line 341: A few implementation remarks: `compare (length es) n` with a case case would arguably be more readable than nested `if then else` here, we could also use `splitAt` instead of `take` to check that the reminder is not null and avoid an extra call to `length`.

berewt (approved):

marconi-core/src/Marconi/Core/TracedStorable.hs
raduom: One thing I noticed is that using type applications inside a data makes life complicated. I prefer to use type variables and use the type application at the places where i use the data type.

This issue is somewhat incomplete. If we add notification, then we would need to add tracers to all the APIs defined by the type classes (so they can use the notification system as well). If I have the time to do it (and I probably will) I will do it next week.

marconi-core/src/Marconi/Core/TracedStorable.hs
raduom: Nice observation. I debated that myself and @kayvank. I don't think that give me the freshest you have is something we want to support. As you further notice, that would require changing the result type to include an indication of the slot number, but that is inherently unsafe (what if some user just ignores the information?).

Always knowing the last processed block is one of the features allowed by the notification system, so the user will be able to always know what is the latest block, so they can specify it in the query.

marconi-core/test/Marconi/Core/Spec/TracedSqlite.hs
```diff
@@ -0,0 +1,372 @@
...
+monadic = monadicIO
+
+{-
+   This function is needed to lookup a point from the model into the new version of
+   indexers. This is needed as rewinds in the model take a number of blocks to rewind
+   back, while in the new indexers they take a point that we want to rewind to.
+
+   This is not trivial because it has to respect both the rules surrounding rewinds
+   in the original model and the rules surrounding rewinds in the adapter version.
+-}
```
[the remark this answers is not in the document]
raduom: I think that could work. If I understand correctly what you mean is have a framework for validating generic indexers, by filling in only the specific implementations (type class instances) and then pugging it into the tests.

marconi-core/test/Marconi/Core/Spec/TracedSqlite.hs
raduom: I am not seeing all the changes at once right now, but I try to rewrite the code later.

marconi-core/src/Marconi/Core/TracedStorable.hs
raduom: See previous comment ^^

marconi-core/src/Marconi/Core/TracedStorable.hs
```diff
@@ -0,0 +1,391 @@
...
+     things performant -}
+  getStoredEvents :: h -> StorableMonad h [StorableEvent h]
+
+{-
+   All information from indexers should be made accessible through queries. Queries
+   act a lot like folds over the event stream. Each query introduces two indexed types:
+   one for the type of requests (called StorableQuery) and one for responses (called
+   StorableResult).
+
+   All queries include a validity interval. If the data is not available for a specified
```
eyeinsky, line 136: But we have no interval anymore (?).

marconi-core/src/Marconi/Core/TracedStorable.hs
eyeinsky, line 143: Instead of removing the interval, why not just make it work like an interval where you can get all events in that interval without needing to start from the beginning?

eyeinsky (commented):
The difference between Marconi.Core.Storable and Marconi.Core.TracedStorable is very small, why not instead of adding a new file every time where it's hard to tell what changed and what stayed the same why not evolve the indexer API that we already have with added tracing and removed query interval? Needing to change all the indexer implementations would also show what are you meaning to do with this change. As far as I can tell, the current file removes interval and adds tracing while keeping the rest the same.

Second question about using tracing for notifications: in the rest of our code bases tracing is used for logging, so how would an indexer user use it for getting the latest chain point processed, would they need to listen on the indexer's stdout and parse the output? If our query is built with JSON RPC then why not implement querying the latest-chainpoint-processed through it as well?

marconi-core/src/Marconi/Core/TracedStorable.hs
```diff
@@ -0,0 +1,391 @@
...
+-- | The resume and query functionality requires a way to specify points on the chain from which we
+-- want to resume, or points up to which we want to query.
+type family StorablePoint h
+
+data family StorableQuery h
+
+data family StorableResult h
+
+-- TODO: Rename `Storable` to `Indexer`.
+-- data family StorableNotifications h
```
koslambrou, line 79: ?

marconi-core/src/Marconi/Core/TracedStorable.hs
```diff
@@ -0,0 +1,391 @@
...
+
+data ControlNotification pt =
+    CNRollForward !pt
+  | CNRollBack    !pt
+ ---- | CNApplication !(StorableNotifications h)
```
koslambrou, line 84: ?

raduom:
> The difference between Marconi.Core.Storable and Marconi.Core.TracedStorable is very small, why not instead of adding a new file every time where it's hard to tell what changed and what stayed the same why not evolve the indexer API that we already have with added tracing and removed query interval? Needing to change all the indexer implementations would also show what are you meaning to do with this change. As far as I can tell, the current file removes interval and adds tracing while keeping the rest the same.

I made a new file because I will probably not have the time to also rewrite the indexers that use the new API. The way the indexer is right now it does not actually change the API, but I think that in it's final form we will have to do that. I will do some more work. It may be the case that changing the existing indexer to conform to the new API will be a very easy exercise.

> Second question about using tracing for notifications: in the rest of our code bases tracing is used for logging, so how would an indexer user use it for getting the latest chain point processed, would they need to listen on the indexer's stdout and parse the output? 

Traces can be emitted using any kind of `contra-tracer` adapter or even `iohk-monitoring`. One such adapter, for example, could open a websocket and publish updates as the node consumes blocks. Or, simpler, a `TChan` where these updates are forwarded. Some other software component listening to the `TChans` created by all indexers will have an accurate view of synchronisation status across all indexers.

> If our query is built with JSON RPC then why not implement querying the latest-chainpoint-processed through it as well?

It's rather inconvenient for the user. I think this change is functionally the same, but easier for our users (we can provide the component that listens on the `TChan`s and the user does not have to worry about it). It also enables the implementation of further functionalities in indexers.

marconi-core/src/Marconi/Core/TracedStorable.hs
raduom: I think that the `Indexer` name is better, but it is just a suggestion for a future generation. :)

marconi-core/src/Marconi/Core/TracedStorable.hs
raduom: I intend to uncomment that if I have the time for it. Depending on how the reviews go for the new ADR.

marconi-core/src/Marconi/Core/TracedStorable.hs
raduom: In the most generic case you need to start from the beginning, since the events that are stored in the database are actually only `deltas` from the previous state. My attempt at doing just that was met with a lot of confusion. Removing them seems like the easier path.

marconi-core/src/Marconi/Core/TracedStorable.hs
raduom: Cut & Paste error :)

berewt (approved):
No specific remarks, the ADR looks good. We may need to complete it with a Marconi specific architecture for tracers, but out of the scope here.

marconi-core/src/Marconi/Core/TracedStorable.hs
eyeinsky: Many indexers don't have events as deltas, e.g epoch stakepoolsize, ScriptTx and MintBurn don't (?). Our events are the things that "business needs", i.e what Djed or Mamba wants. ScriptTx and MintBurn events aren't even folds -- they can be started from anywhere in the chain and the results would be valid for the section of blockchain they indexed.

raduom: Unless you can make that statement for all indexers my reasoning still stands.

eyeinsky: The design should be flexible then in that dimension.

marconi-core/src/Marconi/Core/TracedStorable.hs
```diff
@@ -0,0 +1,399 @@
...
+
+-- | The resume and query functionality requires a way to specify points on the chain from which we
+-- want to resume, or points up to which we want to query.
+type family StorablePoint h
+
+data family StorableQuery h
+
+data family StorableResult h
+
+-- TODO: Rename `Storable` to `Indexer`.
```
kayvank, line 80: this is a good TODO to perform. I don't think it is required for this PR

kayvank (approved):
LGTM

koslambrou:
@kayvank Can you change the ADR name :D

kayvank:
> 

Discussed this with @berewt in detail.  See the JIRA ticket @berewt created capturing the final approach/proposal for handling indexer-queries that areb past the latest sync point

kayvank:
> @kayvank Can you change the ADR name :D

how about: "Indexers query synchronisation primitive"?

koslambrou:
> how about: "Indexers query synchronisation primitive"?

@kayvank LGTM :). The file name _and_ the title need to changed.

PR #1012, opened 2023-03-07
PLT-967 Use reference script in the game example

berewt:
It has required the creation of additional constraint functions to pass the reference script txOutRef along the spend utxos.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-use-cases/src/Plutus/Contracts/GameBabbage.hs
+++ b/plutus-use-cases/src/Plutus/Contracts/GameBabbage.hs
@@ -204,7 +204,7 @@ guess = endpoint @"guess" $ \GuessArgs { guessArgsGameParam, guessArgsGameAddres
                Haskell.<> Constraints.unspentOutputs utxos
                Haskell.<> Constraints.unspentOutputs gameUtxos
         redeemer = clearString guessArgsSecret
-        tx       = Constraints.collectFromTheReferencedScript utxos redeemer gameRef
+        tx       = Constraints.spendUtxosFromTheReferencedScript utxos redeemer gameRef
     unbalancedTx <- mkTxConstraints lookups tx
     yieldUnbalancedTx unbalancedTx
 
--- a/plutus-use-cases/src/Plutus/Contracts/GameBabbage.hs
+++ b/plutus-use-cases/src/Plutus/Contracts/GameBabbage.hs
@@ -198,13 +198,13 @@ guess = endpoint @"guess" $ \GuessArgs { guessArgsGameParam, guessArgsGameAddres
     utxos <- utxosAt $ gameAddress guessArgsGameParam
     let game = gameInstance guessArgsGameParam
         gameHash = validatorHash game
+    gameUtxos <- utxosAt guessArgsGameAddress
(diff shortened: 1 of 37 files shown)


plutus-use-cases/src/Plutus/Contracts/GameBabbage.hs
```diff
@@ -0,0 +1,252 @@
+{-# LANGUAGE AllowAmbiguousTypes        #-}
```
koslambrou, line 1: How about moving the duplication between `Game` and `GameBabbage` into a separate module? It could be something like:

* create the `Plutus.Contract.Game.Types` module and move the GameParams, LockArgs, and UnlockArgs over there.
* create the `Plutus.Contract.Game.PlutusV1` module
* create the `Plutus.Contract.Game.PlutusV2` module

The reason I'm proposing `PlutusV1` and `PlutusV2` instead of `Alonzo` and `Babbage` respectively is because the tx that we are creating are *always* `Babbage`. The difference between the two examples are mostly the Plutus version and features of a specific era.

plutus-use-cases/src/Plutus/Contracts/GameBabbage.hs
```diff
@@ -0,0 +1,252 @@
...
+      $ searchReferenceScript hash utxos
+    where
+        searchReferenceScript :: ValidatorHash -> Map TxOutRef DecoratedTxOut -> Maybe TxOutRef
+        searchReferenceScript (ValidatorHash h) = let
+            getReferenceScriptHash = _2 . decoratedTxOutReferenceScript
+                . _Just . to (getScriptHash . scriptHash)
+                . only h
+            in fmap fst
+            . find (isJust . preview getReferenceScriptHash)
+            . Map.toList
```
koslambrou, line 236: Maybe that should go in `Plutus.Contract.Request`?

plutus-use-cases/src/Plutus/Contracts/GameBabbage.hs
berewt: I like the submodule idea, but I don't if Plutus version are better than era: new features comes with era and are supported by Plutus versions. The original script were a Plutus v2 script, that didn't take advantage of the Babbage feature... I like the era more, because one could look at the era specification and see what's new. But I'm gonna introduce the submodule (and document the meaning of the eras - ie. only use features up to the given era)

plutus-use-cases/src/Plutus/Contracts/GameBabbage.hs
berewt: Actually, I was mentoning it with @james-iohk, as it can probably be reused more widely, I'd like to do it in a separate PR to add tests to it.

plutus-use-cases/src/Plutus/Contracts/GameBabbage.hs
koslambrou: You're right. Forgot that all plutus-use-cases where updated to PlutusV2. In that case, it does make sense.

PR #1017, opened 2023-03-08
PLT-748 Added documentation and placeholder code blocks for the queries needed by `marconi-siechain`.

koslambrou:
* Updated the `marconi-sidechain` README by providing all the queries that are needed by the Sidechain team.

* Added golden tests for the different responses of the JSON-RPC endpoints.

* Changed name of JSONRPC to JsonRpcContentType

* Create the TxMintRow in the MintBurn indexer instead of working with tuples

* Added ToJSON and FromJSON instances for the TxMintRow datatype as well as a roundtrip property test.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


kayvank (approved):

marconi-sidechain/src/Marconi/Sidechain/Api/Routes.hs
```diff
@@ -1,40 +1,132 @@
...
+--  RPC types
+--  methodName -> parameter(s) -> return-type
+----------------------------------------------
 
 -- | JSON-RPC API, endpoint
 type JsonRpcAPI = "json-rpc" :> RawJsonRpc RpcAPI
 
+-- | RPC routes
+type RpcAPI = RpcEchoEndpoint
+         :<|> RpcTargetAddressesEndpoint
```
kayvank, line 33: This is just a preference, and may not apply.
RPCs share the same edpoint, where to me, maybe incorrect, endpoint is is the location from which APIs can access the resources they need to carry out their function. And adding endpoint to the type name may mislead that each type has a different endpoint?

koslambrou: Good point! So maybe use the `Method` suffix?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
```diff
@@ -132,6 +133,46 @@ fromAlonzoData = C.fromPlutusData . LA.getPlutusData -- from cardano-api:src/Car
 
 -- * Sqlite
 
+data TxMintRow = TxMintRow
+    { _txMintRowSlotNo          :: !C.SlotNo
```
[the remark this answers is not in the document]
koslambrou: Ah just remembered why we don't want this. That's because `ChainPoint` has a data constructor which has no parameters (ChainPointAtGenesis). In our scenario, we don't want to represent ChainPointAtGenesis in JSON.

PR #1018, opened 2023-03-08
PLT-214 Refactored the Epoch-StakePoolDelegation indexer to the Marconi interface.

koslambrou:
* Rewrote the Epoch-StakePoolDelegation using the marconi-core interface and to make it support rollbacks and resuming.

* Computed the LedgerState directly with functions in `ouroboros-network` instead of using `cardano-api` and `cardano-streaming` (allows more fined grained control of what to extract).

* Moved orphan instances of EpochStakepoolSize indexer to the Orphans module and added rountrip ToField/FromField tests.

* Added LedgerState serialization/deserialization functions in Orphans module.

IMPORTANT NOTE: 

Not very performant, but not clear on how to make it better. The use of `LedgerState` uses lots of memory. Could not fully sync on my machine as I don't have enough memory (32GB). The node by itself uses about 12GB. I have 6GB for other processes. Then, I only managed to sync the indexer at 85% before using all of my memory. I'll try again by removing some of the other processes, and see if I can complete it.

However, I think I have optimized it as much as I can (given the functions from `ouroboros-network` and `cardano-ledger`). I believe that memory usage is inevitable if we're going to continue using cardano core functions.

Edit 1:

Removed that used 6GB of different processes, and resynced again. This time, it managed to fully sync in about 24h and max RAM usage was just a bit under 16GB.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/marconi-chain-index/performance/monitor-marconi-sync-all.sh
+++ b/marconi-chain-index/performance/monitor-marconi-sync-all.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+
+NETWORKID=764824073
+NETWORK=mainnet
+
+monitor_indexer () {
+  indexer=$1
+  echo "Syncing ${indexer} indexer..."
+  case $indexer in
+    utxo)
+      indexers_cli="--disable-datum --disable-script-tx --disable-address-datum --disable-mintburn --disable-epoch-stakepool-size"
+      ;;
+    address-datum)
+      indexers_cli="--disable-datum --disable-script-tx --disable-utxo --disable-mintburn --disable-epoch-stakepool-size"
+      ;;
+    mintburn)
+      indexers_cli="--disable-datum --disable-script-tx --disable-utxo --disable-address-datum --disable-epoch-stakepool-size"
+      ;;
+    epochsdd)
+      indexers_cli="--disable-datum --disable-script-tx --disable-utxo --disable-address-datum --disable-mintburn"
+      ;;
+  esac
+
+  cabal build marconi-chain-index
+  $(cabal exec -- which marconi-chain-index) \
+    -s "$HOME"/cardano-node/${NETWORK}/cardano-node.socket \
+    --testnet-magic ${NETWORKID} \
+    ${indexers_cli} \
+    -d ~/cardano-node/${NETWORK}/marconi-chain-index \
+    --node-config-path "$HOME"/cardano-node/${NETWORK}/${NETWORK}-config.json >> marconi-chain-index-"${indexer}".log &
+  pid=$!
+  echo $pid
+  ./monitor-marconi-sync.sh marconi-chain-index-"${indexer}".log $pid >> >(tee marconi-chain-index-"${indexer}"-monitor.log)
+
+}
+
+# Uncomment the ones you want to use
+# monitor_indexer "utxo"
+# monitor_indexer "address-datum"
+# monitor_indexer "mintburn"
+# monitor_indexer "epochsdd"
+
+trap 'kill $(jobs -p)' EXIT
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
@@ -325,13 +325,10 @@ epochStakepoolSizeWorker_
               RollBackward cp _ct ->
                   modifyMVar indexerMVar $ \ix -> do
                   newIndex <- fromMaybe ix <$> Storable.rewind cp ix
-                  -- The possible points from which we can possibly rollback should be available
-                  -- in the buffer events and from the resumable points.
-                  -- For that assumption to be correct, we absolutely need
-                  --    * to make sure that 'EpochSPD.open' was called with the correct
-                  --    'SecurityParam' (k) value of connect Cardano network.
-                  --    * that the resumablePoints correctly returns the points from which we
-                  --    have saved a LedgerState on disk.
+                  -- We query the LedgerState from disk at the point where we need to rollback to.
+                  -- For that to work, we need to be sure that any volatile LedgerState are stored
+                  -- on disk. For immutable LedgerStates, they are only stored on disk at the first
+                  -- slot of an epoch.
                   EpochSPD.LedgerStateAtPointResult maybeLedgerState <-
                       Storable.query Storable.QEverything newIndex (EpochSPD.LedgerStateAtPointQuery cp)
                   case maybeLedgerState of
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
@@ -9,51 +9,58 @@
 
 module Marconi.ChainIndex.Indexers where
 
-import Control.Concurrent (MVar, forkIO, modifyMVar_, newMVar, readMVar)
-import Control.Concurrent.QSemN (QSemN, newQSemN, signalQSemN, waitQSemN)
-import Control.Concurrent.STM (atomically)
-import Control.Concurrent.STM.TChan (TChan, dupTChan, newBroadcastTChanIO, readTChan, writeTChan)
-import Control.Exception (catch)
-import Control.Lens (view, (&))
-import Control.Lens.Operators ((^.))
-import Control.Monad (forever, void)
-import Control.Monad.Trans.Class (lift)
-import Data.List (findIndex, foldl1', intersect)
-import Data.Map (Map)
-import Data.Map qualified as Map
-import Data.Maybe (fromMaybe, mapMaybe)
-import Data.Text qualified as TS
-import Database.SQLite.Simple qualified as SQL
-import Streaming.Prelude qualified as S
-
 import Cardano.Api (Block (Block), BlockHeader (BlockHeader), BlockInMode (BlockInMode), CardanoMode,
                     ChainPoint (ChainPoint, ChainPointAtGenesis), Hash, ScriptData, SlotNo, Tx (Tx), chainPointToSlotNo)
 import Cardano.Api qualified as C
-import Cardano.Api.Shelley qualified as Shelley
+import Cardano.Api.Shelley qualified as C
 import Cardano.BM.Setup (withTrace)
 import Cardano.BM.Trace (logError)
 import Cardano.BM.Tracing (defaultConfigStdout)
 import Cardano.Ledger.Alonzo.TxWitness qualified as Alonzo
 import Cardano.Streaming (ChainSyncEvent (RollBackward, RollForward), ChainSyncEventException (NoIntersectionFound),
                           withChainSyncEventStream)
-import Cardano.Streaming qualified as CS
-import Marconi.ChainIndex.Logging (logging)
-import Prettyprinter (defaultLayoutOptions, layoutPretty, pretty, (<+>))
-import Prettyprinter.Render.Text (renderStrict)
-
+import Control.Concurrent (MVar, forkIO, modifyMVar_, newMVar, readMVar)
+import Control.Concurrent.MVar (modifyMVar)
+import Control.Concurrent.QSemN (QSemN, newQSemN, signalQSemN, waitQSemN)
+import Control.Concurrent.STM (atomically)
+import Control.Concurrent.STM.TChan (TChan, dupTChan, newBroadcastTChanIO, readTChan, writeTChan)
+import Control.Exception (catch)
+import Control.Lens (view)
+import Control.Lens.Operators ((^.))
+import Control.Monad (forever, void)
+import Control.Monad.Trans.Except (runExceptT)
+import Data.List (findIndex, foldl1', intersect)
+import Data.Map (Map)
+import Data.Map qualified as Map
+import Data.Maybe (fromMaybe, mapMaybe)
+import Data.Text qualified as TS
+import Data.Word (Word64)
 import Marconi.ChainIndex.Indexers.AddressDatum (AddressDatumDepth (AddressDatumDepth), AddressDatumHandle,
                                                  AddressDatumIndex)
 import Marconi.ChainIndex.Indexers.AddressDatum qualified as AddressDatum
 import Marconi.ChainIndex.Indexers.Datum (DatumIndex)
 import Marconi.ChainIndex.Indexers.Datum qualified as Datum
-import Marconi.ChainIndex.Indexers.EpochStakepoolSize qualified as EpochStakepoolSize
+import Marconi.ChainIndex.Indexers.EpochStakepoolSize (EpochSPDHandle, EpochSPDIndex)
+import Marconi.ChainIndex.Indexers.EpochStakepoolSize qualified as EpochSPD
 import Marconi.ChainIndex.Indexers.MintBurn qualified as MintBurn
 import Marconi.ChainIndex.Indexers.ScriptTx qualified as ScriptTx
 import Marconi.ChainIndex.Indexers.Utxo qualified as Utxo
+import Marconi.ChainIndex.Logging (logging)
+import Marconi.ChainIndex.Node.Client.GenesisConfig (NetworkConfigFile (NetworkConfigFile), initExtLedgerStateVar,
+                                                     mkProtocolInfoCardano, readCardanoGenesisConfig, readNetworkConfig,
+                                                     renderGenesisConfigError)
 import Marconi.ChainIndex.Types (TargetAddresses)
-
 import Marconi.Core.Index.VSplit qualified as Ix
 import Marconi.Core.Storable qualified as Storable
+import Ouroboros.Consensus.Config qualified as O
+import Ouroboros.Consensus.Ledger.Abstract qualified as O
+import Ouroboros.Consensus.Ledger.Extended qualified as O
+import Ouroboros.Consensus.Node qualified as O
+import Prettyprinter (defaultLayoutOptions, layoutPretty, pretty, (<+>))
+import Prettyprinter.Render.Text (renderStrict)
+import Streaming.Prelude qualified as S
+import System.Directory (createDirectoryIfMissing)
+import System.FilePath (takeDirectory, (</>))
 
 -- DatumIndexer
 getDatums :: BlockInMode CardanoMode -> [(SlotNo, (Hash ScriptData, ScriptData))]
@@ -67,13 +74,13 @@ getDatums (BlockInMode (Block (BlockHeader slotNo _ _) txs) _) = concatMap extra
             $ txBody
 
 scriptDataFromCardanoTxBody :: C.TxBody era -> Map (Hash ScriptData) ScriptData
-scriptDataFromCardanoTxBody (Shelley.ShelleyTxBody _ _ _ (C.TxBodyScriptData _ dats _) _ _) =
+scriptDataFromCardanoTxBody (C.ShelleyTxBody _ _ _ (C.TxBodyScriptData _ dats _) _ _) =
     extractData dats
   where
     extractData :: Alonzo.TxDats era -> Map (Hash ScriptData) ScriptData
     extractData (Alonzo.TxDats' xs) =
       Map.fromList
-      . fmap ((\x -> (C.hashScriptData x, x)) . Shelley.fromAlonzoData)
+      . fmap ((\x -> (C.hashScriptData x, x)) . C.fromAlonzoData)
       . Map.elems
       $ xs
 scriptDataFromCardanoTxBody _ = mempty
@@ -90,12 +97,10 @@ scriptDataFromCardanoTxBody _ = mempty
      how many we are waiting.
 -}
 data Coordinator = Coordinator
-  { _channel      :: TChan (ChainSyncEvent (BlockInMode CardanoMode))
-  , _barrier      :: QSemN
-  , _indexerCount :: Int
+  { _channel      :: !(TChan (ChainSyncEvent (BlockInMode CardanoMode)))
+  , _barrier      :: !QSemN
+  , _indexerCount :: !Int
   }
-
-
 
 initialCoordinator :: Int -> IO Coordinator
 initialCoordinator indexerCount =
@@ -134,7 +139,10 @@ utxoWorker_
   :: (Utxo.UtxoIndexer -> IO ())  -- ^ callback function used in the queryApi thread, needs to be non-blocking
   -> Utxo.Depth
   -> Maybe TargetAddresses                       -- ^ Target addresses to filter for
-  -> Coordinator -> TChan (ChainSyncEvent (BlockInMode CardanoMode)) -> FilePath -> IO (IO (), MVar Utxo.UtxoIndexer)
+  -> Coordinator
+  -> TChan (ChainSyncEvent (BlockInMode CardanoMode))
+  -> FilePath
+  -> IO (IO (), MVar Utxo.UtxoIndexer)
 utxoWorker_ callback depth maybeTargetAddresses Coordinator{_barrier} ch path = do
   ix <- Utxo.open path depth
   mIndexer <- newMVar ix
@@ -203,7 +211,7 @@ addressDatumWorker_ onInsert targetAddresses depth Coordinator{_barrier} ch path
         RollForward (BlockInMode (Block (BlockHeader slotNo bh _) txs) _) _ -> do
             -- TODO Redo. Inefficient filtering
             let addressFilter =
-                    fmap (\targetAddrs -> \addr -> addr `elem` targetAddrs)
+                    fmap (flip elem)
                          targetAddresses
                 addressDatumIndexEvent =
                     AddressDatum.toAddressDatumIndexEvent addressFilter txs (C.ChainPoint slotNo bh)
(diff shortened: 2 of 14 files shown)


marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochSPD.hs
```diff
@@ -0,0 +1,442 @@
...
+        -- Only keep the event with the largest slot number of each epoch.
+        let eventsToStore =
+                mapMaybe (\es -> listToMaybe
+                               $ List.sortOn (\e -> Down $ epochSPDEventSlotNo e) es)
+                  $ List.groupBy (\e1 e2 -> epochSPDEventEpochNo e1 == epochSPDEventEpochNo e2)
+                  eventsList
+
+        forM_ (concatMap eventToEpochSPDRows eventsToStore) $ \row ->
+          SQL.execute c
+              [r|INSERT OR REPLACE INTO epoch_spd
```
kayvank, line 243: the Replace works when there is a unique constraint on the table, as a [constraint violation resolution](https://www.sqlite.org/lang_conflict.html).  
I don't think this is what we want here

marconi-chain-index/performance/monitor-marconi-sync-all.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
```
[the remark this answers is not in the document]
koslambrou: Not sure I want to include this script in the README. Let's actually use it and evolve it before documenting it. Not sure how useful it will be in the end.

marconi-chain-index/src/Marconi/ChainIndex/Node/Client/GenesisConfig.hs
eyeinsky, line 1: Perhaps mark the parts of this file which are a copy-paste from elsewhere, otherwise we won't remember to replace it with imports later on.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochSPD.hs
eyeinsky, line 1: Why not simply change marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs instead of copying some parts? We have the old version in git, so we can always go back and look at the original file if the need arose.

eyeinsky (commented):
When stopping and resuming indexing, are the stakepool sizes the same as what dbsync gives?

marconi-chain-index/src/Marconi/ChainIndex/Node/Client/GenesisConfig.hs
koslambrou: :+1:

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochSPD.hs
koslambrou: Good point :) Will change.

eyeinsky:
(Just in case noting a non-code comment from above, maybe it went unnoticed

> When stopping and resuming indexing, are the stakepool sizes the same as what dbsync gives?

That is to say, the result of the indexing process after adding the resuming feature should continue matching what dbsync gives. Working on optimizing performance before that is achieved will only make it harder to fix later on.)

marconi-chain-index/performance/monitor-marconi-sync-all.sh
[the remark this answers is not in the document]
koslambrou: Good point. I'll create a story for doing that.

koslambrou:
@eyeinsky Haven't compared it to db-sync. Since we have another story for that, let's do that in a future PR. I won't optimize further this PR. Is that alright?

koslambrou:
@eyeinsky Applied comments. Approve PR?

eyeinsky (commented):
The overall value with this pull request is figuring out how to store and load ledger state.

But what would be great to know is whether indexing with stopping and resuming produces the same result as it did before. Until that is done we don't even know if this refactoring was successful or not. And moving to marconi-core in addition to adding saving and loading ledger state adds a second layer for uncertainty.

As there is another PR pending then it would be much better to do one thing from start to finish instead of doing two half way (in addition to adding rountrip property tests for e.g EpochNo, newlines to type signatures -- which aren't all that important). This way we can be sure that the code we merge works and thus all of what is in main branch continues to work without regression (as far as we can tell). This is a good practice to have in general, then we know how far we have gotten with the project, what we can vouch for, etc.

---

Ideas for how to test without needing to run it for the ~20h:
- instead of mainnet use preview, preprod
- run node on mainnet [up to a certain slot](https://input-output.atlassian.net/browse/PLT-1542) and index only that
- compare indexing results with the previous version of the indexer (as I did verify the results of it to what dbsync gave)
- use Koios, which exposes dbsync over a REST API: https://api.koios.rest/#overview

koslambrou:
> As there is another PR pending then it would be much better to do one thing from start to finish instead of doing two half way

I mean, these are 2 separate functionality (one handles indexing, and the other is JSON-RPC querying). Makes sense to do them separately, especially for easier code reviewing. We _should_ try to make PRs as small as possible.

My regret is not breaking this one up a bit more, as the logic is a bit complex and error-prone.

> Ideas for how to test without needing to run it for the ~20h:

Good point. I'll fully sync Marconi, run some manual tests with a local db-sync on preprod and see if I get the same result on some queries. Then, I'll stop and resume, and redo the manual test.

marconi-chain-index/src/Marconi/ChainIndex/Node/Client/GenesisConfig.hs
```diff
@@ -0,0 +1,435 @@
+{-# LANGUAGE DataKinds                  #-}
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE GADTs                      #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE RankNTypes                 #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+
+{-| This module is a duplication of functionality from the `Cardano.Api.LedgerState`
```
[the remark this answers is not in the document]
koslambrou: :+1:

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
```diff
@@ -1,187 +1,442 @@
...
+--
+-- The main issue with this indexer is that building the LedgerState and saving it on disk for being
+-- able to resume is VERY resource intensive. Syncing time for this indexer is over 20h and uses
+-- about ~16GB of RAM (which will keep increasing as the blockchain continues to grow).
+--
+-- Here is a synopsis of what this indexer does.
+--
+-- We assume a few things:
+--
+--   * the construction of 'LedgerState' is done outside of this indexer.
```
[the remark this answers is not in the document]
koslambrou: Actually, what I meant was that the LedgerState needs to be compute outside this module (therefore outside the indexer)

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
```diff
@@ -1,187 +1,442 @@
...
+--   last 'LedgerState' before starting a new epoch). We do that because we only care about the SPD
+--   (Stake Pool Delegation) from the last block before a new epoch.
+--
+-- Once the 'Storable.StorableEvent' is stored on disk, we perform various steps:
+--
+--   1. we save the SPD for the current epoch in the `epoch_spd` table
+--   2. we save the 'LedgerState' in the filesystem as a binary file (the ledger state file path has
+--   the format: ledgerState_<SLOT_NO>_<BLOCK_HEADER_HASH>_<BLOCK_NO>.bin
+--   3. we delete old 'LedgerState' binary files and only keep the latest one which can't be
+--   rollbacked anymore (this step is necessary for resuming).
```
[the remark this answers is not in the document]
koslambrou: > It's step 2 which is necessary for resuming?

Yeah. I'll specify that.

> Deleting old ledger states (which makes sense) is for saving space.

Exactly.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
```diff
@@ -1,187 +1,442 @@
...
+                          $ filter (\(_, _, isImmutableBlock) -> isImmutableBlock)
+                          $ NE.tail
+                          $ NE.sortWith (\(_, (_, _, blockNo), isImmutableBlock) -> Down (blockNo, isImmutableBlock))
+                          $ fmap (\(fp, (slotNo, bhh, blockNo@(C.BlockNo bn))) -> (fp, (slotNo, bhh, blockNo), latestBlockNo - bn > securityParam))
+                          nonEmptyLedgerStateFilePaths
+                  forM_ oldLedgerStateFilePaths $ \fp -> removeFile $ ledgerStateDirPath </> fp
+
+        pure h
+
+    -- TODO For now, this indexer does not support storing volatile events on disk, as we assume
```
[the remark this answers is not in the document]
koslambrou: Old TODO. Forgot to change it.

I'll change the comment to "buffering is not in use and we don't need to retrieve stored events in our implementation".

marconi-chain-index/src/Marconi/ChainIndex/Node/Client/GenesisConfig.hs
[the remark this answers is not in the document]
koslambrou: No point in creating an issue about it on `cardano-node`. This module is probably temporary as we might get rid of the use of LedgerState.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
```diff
@@ -1,187 +1,442 @@
...
+-- 'LegderState', we can easily compute the epochNo as well as the stake pool delegation for that
+-- epoch.
+--
+-- The main issue with this indexer is that building the LedgerState and saving it on disk for being
+-- able to resume is VERY resource intensive. Syncing time for this indexer is over 20h and uses
+-- about ~16GB of RAM (which will keep increasing as the blockchain continues to grow).
+--
+-- Here is a synopsis of what this indexer does.
+--
+-- We assume a few things:
```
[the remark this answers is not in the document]
koslambrou: I agree for the 2nd point. However, the first _is_ an assumption. I'm assuming that the LedgerState is contructed _outside_ the module.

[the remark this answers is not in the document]
eyeinsky: One assumes something that they don't know to be true (but assume it to be true), so this would mean that we don't know that the code is in a different file. But I think we do know where the code is.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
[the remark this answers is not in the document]
eyeinsky: The code that implements computing ledger state is not written within the indexer, sure, but the actual computation of it is done in the indexer during runtime. And there is a second ledger state that is computed within the node. (And I would assume, i.e haven't checked :), that the code that is used to compute ledger state is same for both and lives somewhere in the ouroboros repo.)

koslambrou:
@eyeinsky Ok, I'm definitely not getting the same results as db-sync. However, I'm using the same logic as before to compute the stake pool delegation. Going to dig deeper and even take a look at db-sync's implementation.

koslambrou:
@eyeinsky Ok, just confirmed that the old implementation is incorrect. My refactoring produces similar result to what we had before. However, the results are different than db-sync. They are the same for early epochs. However, at later epochs (for example epoch 32), I'm getting different stake pool delegations. That's probably what the slices of db-sync are about.

Given that this PR gives similar results to the previous implementation and that it's a refactoring PR which adds  resume and rollback, I propose to:

* merge this PR (unless you have other serious issues with it)
* work on the testing issue for comparing db-sync and marconi
* fix the discrepancy between the two

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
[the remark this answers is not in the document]
koslambrou: Actually, the actual computation of the LedgerState is done in the `Marconi.ChainIndex.Indexers` module, and not in `Marconi.ChainIndex.Indexers.EpochStakepoolSize`. So my comment is basically saying that the call to `tickThenReapplyLedgerResult` (which computes the new LedgerState given a block), is done outside this module. 

Should I rephrase?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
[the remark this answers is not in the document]
koslambrou: Ahh :D Got it. So I should be talking about a precondition instead?

eyeinsky:
If I remember correctly, the `genericStakeSlice` is where from I got the idea on how to get the stake map from `NewEpochState`. But if the results (when stopping and resuming) match with the old indexer then I guess what was already implemented in the old indexer is maintained and there isn't more to test in that regard.

Were you comparing stakepool sizes from mainnet?

koslambrou:
> If I remember correctly, the genericStakeSlice is where from I got the idea on how to get the stake map from NewEpochState. But if the results (when stopping and resuming) match with the old indexer then I guess what was already implemented in the old indexer is maintained and there isn't more to test in that regard.

Yeah, most of the logic comes from there, but there's some stuff about slicing up the delegations which I don't understand, and seem to give different results from db-sync.

However, just noticed I'm not getting 100% the same results as the old implementation. The reason is because I'm getting the stake pool delegations from the last LedgerState of an epoch, whereas the old implementation computes from the first LedgerState of a new epoch. Do you know which one is right?

> Were you comparing stakepool sizes from mainnet?

No from preprod. Does that change anything?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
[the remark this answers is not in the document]
eyeinsky: Why mention it at all? The actual implementation of applying blocks to ledger state is in the ouroboros repo anyway and all of us, including marconi, cardano-node, cardano-api and perhaps others import it from there.

koslambrou:
@eyeinsky Alright, changed the implementation to get the sdd from the first LedgerState of an epoch. Getting the same results as the old implementation. Fully synced on preprod, stopped and resumed, and getting same result as the old implementation.

Can you do a last quick PR review?

> To do this, using the streaming library ...

Unfortunately, I already tried using the streaming library for that. The issue were rollback handling, as we can't rollback on a previous state of a Stream, unless we kept the LedgerStates in memory with `History` (which I didn't want). As the indexer itself handled rollbacks, I had to do without `streaming`.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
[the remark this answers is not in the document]
eyeinsky: Why not keep indexer code inside the same module then? Not only do you keep code for a single indexer in a single module you can also skip this comment.

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -254,27 +262,104 @@ scriptTxWorker onInsert coordinator path = do
...
+                            currentLedgerState
+
+              RollBackward C.ChainPointAtGenesis _ct -> do
+                  modifyMVar_ indexerMVar $ \ix -> fromMaybe ix <$> Storable.rewind C.ChainPointAtGenesis ix
+                  pure initialLedgerState
+              RollBackward cp _ct ->
+                  modifyMVar indexerMVar $ \ix -> do
+                  newIndex <- fromMaybe ix <$> Storable.rewind cp ix
+                  -- The possible points from which we can possibly rollback should be available
+                  -- in the buffer events and from the resumable points.
```
eyeinsky, line 329: If we store only the first ledger state from epoch on disk but the rollback doesn't go to exactly that chain point then I think we actually need to roll back *to that first slot of the epoch* for which we have ledger state for, and then also reconnect local chainsync and start reading blocks from that exact slot.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
```diff
@@ -1,187 +1,502 @@
...
+                    _
+                    slotNo
+                    blockHeaderHash
+                    blockNo
+                    chainTip
+                    isFirstEventOfEpoch) -> do
+            case (maybeEpochNo, maybeLedgerState) of
+              (Just _, Just ledgerState) -> do
+                  let isRollbackable = isBlockRollbackable securityParam blockNo chainTip
+                  when (isRollbackable || isFirstEventOfEpoch) $ do
```
eyeinsky, line 290: Right, so we write ledger state to disk when `isFirstEventOfEpoch` (this is what happens most when starting indexing from scratch, as none of the ledger states are volatile there), and then when we get close to tip, where I guess all the ledger states are volatile. Does this mean that we only get matching ledger state to roll back to when the rollback happens either to the previous slot or to the first slot for the epoch, and otherwise we error?

eyeinsky: I.e I guess rolling back only works in the case where we roll back just one slot?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
```diff
@@ -1,187 +1,502 @@
...
+                    writeLedgerState ledgerState slotNo blockHeaderHash blockNo isRollbackable
+              -- We don't store any 'LedgerState' if the era doesn't have epochs (Byron era) or if
+              -- we don't have access to the 'LedgerState'.
+              _noLedgerStateOrEpochNo -> pure ()
+
+        -- Remove all immutable LedgerStates from the filesystem expect the most recent immutable
+        -- one which is from the last slot of latest epoch.
```
eyeinsky, line 297: Now it's the first slot of the next epoch (right?).

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
```diff
@@ -113,37 +118,63 @@ test = H.integration . HE.runFinallies . TN.workspace "chairman" $ \tempAbsPath
...
+  -- This is the channel we wait on to know if the event has been indexed
+  indexedTxs <- liftIO IO.newChan
+  -- Start indexer
+  coordinator <- liftIO $ Indexers.initialCoordinator 1
+  ch <- liftIO $ STM.atomically . STM.dupTChan $ Indexers._channel coordinator
+  let dbPath = tempAbsPath </> "epoch_stakepool_sizes.db"
+  (loop, _indexerMVar) <- liftIO $ Indexers.epochStakepoolSizeWorker_
+      (TN.configurationFile runtime)
+      (STM.writeChan indexedTxs)
+      123
```
eyeinsky, line 133: The correct security param for the testnets appears to be 10.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
```diff
@@ -1,187 +1,502 @@
...
+              Just _                               -> pure ()
+
+        pure $ Just h
+
+instance Resumable EpochSPDHandle where
+    resumeFromStorage
+        :: EpochSPDHandle
+        -> IO [C.ChainPoint]
+    resumeFromStorage (EpochSPDHandle c ledgerStateDirPath _) = do
+        epochSPDChainPoints :: Set C.ChainPoint <- Set.fromList . fmap (uncurry C.ChainPoint) <$>
```
eyeinsky, line 444: Since we can search in sqlite then it might be better to load the available chain points from the ledger states first, and then search for these in sqlite. Currently it loads *all* chain points from database which is large.

eyeinsky:
> Unfortunately, I already tried using the streaming library for that. The issue were rollback handling, as we can't rollback on a previous state of a Stream, unless we kept the LedgerStates in memory with `History` (which I didn't want). As the indexer itself handled rollbacks, I had to do without `streaming`.

It's not like there won't be any challenges at all when using streaming, you obviously need to solve rollbacks. But after you add a component that does that then at a later point in timeline it's safe to fold ledger state, have only one of it in memory, you store that as occasionally as you want etc. (E.g in Marconi.ChainIndex.Logging we have code that demonstrates how to do something every x seconds.) And also zip the stream with its tail to get (previousEpochNo, currentLedgerState).

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
koslambrou: Oops, out of date comment. Sorry :(. See change: 493f6797e65c571592f19db97d6f069b625c8116

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
koslambrou: More precisely, we don't store any LedgerState from ByronEra, because they are no epochs in that era. So we only store the LedgerState starting at each epoch of ShellyEra and onwards.

> Does this mean that we only get matching ledger state to roll back to when the rollback happens either to the previous slot or to the first slot for the epoch, and otherwise we error?

No they work for _any_ rollbackable point. If the block is rollbackable, we store the LedgerState on disk. If securityParam is 2160, then we ultimately store 2160 volatile LedgerState on disk (and eventually delete the ones that are immutable). Therefore, we can always rollback to any point that we can rollback from.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
koslambrou: 289542b0de9313055702bbafa38624783c00032d

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
koslambrou: 8f04c23774d0ce46ea5e6b5ba5350163851ca66c

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
koslambrou: Spent an hour tackling this. Seems like there is no clear way to do something like `WHERE slotNo IN (?) AND blockHeaderHash = (?)` where the `?` is a list of elements. 

Closest I've managed is:

```
    resumeFromStorage
        :: EpochSPDHandle
        -> IO [C.ChainPoint]
    resumeFromStorage (EpochSPDHandle c ledgerStateDirPath _) = do
        ledgerStateFilepaths <- listDirectory ledgerStateDirPath
        let ledgerStateChainPoints =
                fmap (\(_, sn, bhh, _) -> (sn, bhh))
                $ mapMaybe chainTipsFromLedgerStateFilePath ledgerStateFilepaths

        let ledgerStateChainPointsValues =
                fmap (\(slotNo, bhh) -> (slotNo, bhh)) ledgerStateChainPointsledgerStateChainPointsValues
        epochSPDChainPoints :: [C.ChainPoint] <- fmap (uncurry C.ChainPoint) <$>
                SQL.query c
                    [r|SELECT slotNo, blockHeaderHash
                       FROM epoch_spd
                       WHERE slotNo IN (?) AND blockHeaderHash in (?)|]
                    ( Text.intercalate ","
                        $ fmap (\(C.SlotNo slotNo, _) -> Text.pack $ show slotNo) ledgerStateChainPointsValues
                    , Text.intercalate ","
                        $ fmap (\(_, bhh) -> "\"" <> C.serialiseToRawBytesHexText bhh <> "\"") ledgerStateChainPointsValues
                    )
```

However, that seems hacky. Also, I can't match the blockHeaderHash in the table by using the WHERE clause. I thought that was because the `blockHeaderHash` SQL field was a blob. Change it to `TEXT` but doesn't change anything: can't query by blockHeaderHash.

Any pointers or something that I've missed?

koslambrou: Tried this implementation and seems to work: 56a74cfd1999ed2f23206b865b2e558e1d6f29ef
What do you think?

koslambrou:
> It's not like there won't be any challenges at all when using streaming, you obviously need to solve rollbacks. But after you add a component that does that then at a later point in timeline it's safe to fold ledger state, have only one of it in memory, you store that as occasionally as you want etc. (E.g in Marconi.ChainIndex.Logging we have code that demonstrates how to do something every x seconds.) And also zip the stream with its tail to get (previousEpochNo, currentLedgerState).

The point of this PR is to process the LedgerState as soon as possible. I don't want to simply put the LedgerState on stream once it can't be rollbacked anymore. I wanto to process the LedgerState as soon as it's available. The logging module is a bit different as it aggregated the number of number that have occured and keep the result in a TVar. Not what I want to do.

Unless I'm misunderstanding, seems like a complex implementation and out of scope of this PR. If you think that using streaming would totality simplify the implementation, then write a detailed story about it so that we can analyse it.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
eyeinsky: If the indexer catches up with the node such that it will start receiving volatile blocks then we write ledger state for each and every one of these? Is it so that given that in Cardano one block (slot) is one second then in reality we won't actually catch up because we'll be writing the ledger state to disk on every such block, and since writing ledger state to disk takes more than one second then effectively we'll always be 2160 slots behind?

eyeinsky:
> The point of this PR is to process the LedgerState as soon as possible. I don't want to simply put the LedgerState on stream once it can't be rollbacked anymore. I wanto to process the LedgerState as soon as it's available. The logging module is a bit different as it aggregated the number of number that have occured and keep the result in a TVar. Not what I want to do.
> 
> Unless I'm misunderstanding, seems like a complex implementation and out of scope of this PR. If you think that using streaming would totality simplify the implementation, then write a detailed story about it so that we can analyse it.

Right, but we are maybe not getting the fresh ledger state anyway because we are blocked by writing it to disk.

For things in and out of scope, just adding a `writeLedgerState` somewhere in the `foldLedgerState` would have already given us working resume capability. And the rest may have not been required, as we don't use much of the indexer framework anyway, buffer is set to 1, and we are possibly not hitting any rollbacks either, as we're writing ledger sate on every volatile block and doing those writes will stall us).

koslambrou:
> For things in and out of scope, just adding a writeLedgerState somewhere in the foldLedgerState would have already given us working resume capability. And the rest may have not been required, as we don't use much of the indexer framework anyway, buffer is set to 1, and we are possibly not hitting any rollbacks either, as we're writing ledger sate on every volatile block and doing those writes will stall us).

The point of this was to use the indexer framework. Might not be useful right now, but having all indexers be consistent was important. Moreover, we probably won't use LedgerState in the near future for computation that, so that implementation might change and using the Storable interface might be more useful. 

I'd like to merge this ASAP. We can always rework it in the future.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochStakepoolSize.hs
eyeinsky: Seems `in` is not supported :( https://github.com/nurpax/sqlite-simple/issues/91, regarding 56a74cfd1999ed2f23206b865b2e558e1d6f29ef, returning early would be better, and I guess the other idea would be to splice the `where in ...` string together manually.

koslambrou:
> I would break the pieces of code-changes as small as possible and merge things that (to the best of our knowledge) work, so attempting to fix resume and then refactoring the indexer to use the framework could have been better done in two steps, and perhaps using the convert-to-framework step to also evolve the framework at the same time.

Yeah... sorry about that. Totally agree.

PR #1021, opened 2023-03-09
Fix utxo query result and test cases, PLT-1540, and PLT-1536

kayvank:
Fixes are:
- fix the broken property tests in Utxo
- remove the `equivalent` test criteria, replace with sort/compare
- add `isToVacuume` switch indicating SQLite vacuum operation. The switch is not in CLI, and set to True. In test we set that to False.
- fix the Query results
-  Balance the in-memory Utxos in regards to `Spent`
- minor change in the JSON form of the UtxoResponse. 
### Additional notes:
Regression test showing the failure see [PR-1032](https://github.com/input-output-hk/plutus-apps/pull/1032)
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
@@ -144,7 +144,7 @@ utxoWorker_
   -> FilePath
   -> IO (IO (), MVar Utxo.UtxoIndexer)
 utxoWorker_ callback depth maybeTargetAddresses Coordinator{_barrier} ch path = do
-  ix <- Utxo.open path depth
+  ix <- Utxo.open path depth True -- open Sqlite with depth=depth and perform sqltie vacuum
   mIndexer <- newMVar ix
   pure (loop mIndexer, mIndexer)
   where
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
@@ -138,13 +138,13 @@ datumWorker Coordinator{_barrier, _channel} path = do
 utxoWorker_
   :: (Utxo.UtxoIndexer -> IO ())  -- ^ callback function used in the queryApi thread, needs to be non-blocking
   -> Utxo.Depth
-  -> Maybe TargetAddresses                       -- ^ Target addresses to filter for
+  -> Maybe TargetAddresses        -- ^ Target addresses to filter for
   -> Coordinator
   -> TChan (ChainSyncEvent (BlockInMode CardanoMode))
   -> FilePath
   -> IO (IO (), MVar Utxo.UtxoIndexer)
 utxoWorker_ callback depth maybeTargetAddresses Coordinator{_barrier} ch path = do
-  ix <- Utxo.open path depth True -- open Sqlite with depth=depth and perform sqltie vacuum
+  ix <- Utxo.open path depth True -- open SQLite with depth=depth and perform SQLite vacuum
   mIndexer <- newMVar ix
   pure (loop mIndexer, mIndexer)
   where
@@ -154,8 +154,8 @@ utxoWorker_ callback depth maybeTargetAddresses Coordinator{_barrier} ch path = 
       readMVar index >>= callback -- refresh the query STM/CPS with new storage pointers/counters state
       event <- atomically . readTChan $ ch
       case event of
-        RollForward (BlockInMode (Block (BlockHeader slotNo hsh _) txs) _) _ct -> do
-            let utxoEvents = Utxo.getUtxoEvents maybeTargetAddresses txs (C.ChainPoint slotNo hsh)
+        RollForward (BlockInMode block _) _ct -> do
+            let utxoEvents = Utxo.getUtxoEventsFromBlock maybeTargetAddresses block
             modifyMVar_ index (Storable.insert utxoEvents)
             loop index
 
@@ -164,8 +164,8 @@ utxoWorker_ callback depth maybeTargetAddresses Coordinator{_barrier} ch path = 
           loop index
 
 utxoWorker
-  :: (Utxo.UtxoIndexer -> IO ())  -- ^ CPS function used in the queryApi thread, needs to be non-blocking
-  -> Maybe TargetAddresses                       -- ^ Target addresses to filter for
+  :: (Utxo.UtxoIndexer -> IO ())  -- ^ callback function used in the queryApi thread
+  -> Maybe TargetAddresses        -- ^ Target addresses to filter for
   -> Worker
 utxoWorker callback maybeTargetAddresses coordinator path = do
   workerChannel <- atomically . dupTChan $ _channel coordinator
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
@@ -144,7 +144,8 @@ utxoWorker_
   -> FilePath
   -> IO (IO (), MVar Utxo.UtxoIndexer)
 utxoWorker_ callback depth maybeTargetAddresses Coordinator{_barrier} ch path = do
-  ix <- Utxo.open path depth True -- open SQLite with depth=depth and perform SQLite vacuum
+  ix <- Utxo.open path depth False -- open SQLite with depth=depth and DONOT perform SQLite vacuum
+  -- TODO consider adding a CLI param to allow user to perfomr Vaccum or not.
   mIndexer <- newMVar ix
   pure (loop mIndexer, mIndexer)
   where
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
@@ -1,31 +1,38 @@
-{-# LANGUAGE DeriveGeneric        #-}
-{-# LANGUAGE DerivingStrategies   #-}
-{-# LANGUAGE FlexibleInstances    #-}
-{-# LANGUAGE GADTs                #-}
-{-# LANGUAGE NamedFieldPuns       #-}
-{-# LANGUAGE OverloadedStrings    #-}
-{-# LANGUAGE PackageImports       #-}
-{-# LANGUAGE PatternSynonyms      #-}
-{-# LANGUAGE QuasiQuotes          #-}
-{-# LANGUAGE RecordWildCards      #-}
-{-# LANGUAGE TemplateHaskell      #-}
-{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE AllowAmbiguousTypes   #-}
+{-# LANGUAGE DeriveGeneric         #-}
+{-# LANGUAGE DerivingStrategies    #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE GADTs                 #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE NamedFieldPuns        #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE PackageImports        #-}
+{-# LANGUAGE PatternSynonyms       #-}
+{-# LANGUAGE QuasiQuotes           #-}
+{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE UndecidableInstances  #-}
 
-{-
--- | Back-end support for Utxo Indexer
+-- | Module for indexing the Utxos in the Cardano blockchain
 
--- This module will create the SQL tables:
-+ table: unspent_transactions
-|---------+------+-------+-----------+-------+-------+------------------+--------------+------+-------------|
-| Address | TxId | TxIx  | DatumHash | Datum | Value | InlineScriptHash | InlineScript | Slot | BlockNumber |
-|---------+------+-------+-----------+-------+-------+------------------+--------------+------+-------------|
+-- + This module will create the SQL tables:
+--
+-- + table: unspent_transactions
+--
+-- @
+--      |---------+------+-------+-----------+-------+-------+------------------+--------------+------|
+--      | Address | TxId | TxIx  | DatumHash | Datum | Value | InlineScriptHash | InlineScript | Slot |
+--      |---------+------+-------+-----------+-------+-------+------------------+--------------+------|
+-- @
+--
+-- + table: spent
+-- @
+--      |------+------|--------+-----------|
+--      | txId | txIx | slotNo | blockHash |
+--      |------+------|--------+-----------|
+-- @
+-- To create these tables, we extract all transactions outputs from each transactions fetched with
+-- the chain-sync protocol of the local node.
 
-+ table: spent
-  |------+------|--------+-----------|
-  | txId | txIx | slotNo | blockHash |
-  |------+------|--------+-----------|
-
--}
 module Marconi.ChainIndex.Indexers.Utxo where
 
 import Control.Concurrent.Async (concurrently_)
@@ -35,10 +42,14 @@ import Control.Lens.Operators ((^.))
 import Control.Lens.TH (makeLenses)
 import Control.Monad (unless, when)
 import Data.Aeson (FromJSON (parseJSON), ToJSON (toJSON), Value (Object), object, (.:), (.=))
-import Data.Foldable (foldl', toList)
+import Data.Either (fromRight)
+import Data.Foldable (fold, foldl', toList)
 import Data.Functor ((<&>))
-import Data.List (elemIndex)
+import Data.List (groupBy, sortBy)
 import Data.List.NonEmpty (NonEmpty)
+import Data.Map (Map)
+import Data.Map qualified
+import Data.Maybe (mapMaybe)
 import Data.Set (Set)
 import Data.Set qualified as Set
 import Database.SQLite.Simple (Only (Only))
@@ -52,24 +63,26 @@ import Text.RawString.QQ (r)
 
 import Cardano.Api ()
 import Cardano.Api qualified as C
-import Cardano.Api.Shelley qualified as Shelley
+import Cardano.Api.Shelley qualified as C
 import Marconi.ChainIndex.Orphans ()
 import Marconi.ChainIndex.Types (TargetAddresses, TxOut, pattern CurrentEra)
-import Marconi.Core.Storable (Buffered (getStoredEvents, persistToStorage), HasPoint (getPoint),
+
+import Marconi.Core.Storable (Buffered (getStoredEvents, persistToStorage), HasPoint,
                               QueryInterval (QEverything, QInterval), Queryable (queryStorage),
                               Resumable (resumeFromStorage), Rewindable (rewindStorage), StorableEvent, StorableMonad,
-                              StorablePoint, StorableQuery, StorableResult, emptyState, filterWithQueryInterval)
+                              StorablePoint, StorableQuery, StorableResult, emptyState)
 import Marconi.Core.Storable qualified as Storable
 
 type UtxoIndexer = Storable.State UtxoHandle
 
 data UtxoHandle = UtxoHandle
-  { hdlConnection :: SQL.Connection
-  , hdlDpeth      :: Int
+  { hdlConnection :: !SQL.Connection  -- ^ SQLite connection
+  , hdlDpeth      :: !Int             -- ^ depth before flushing to disk storage
+  , toVacuume     :: !Bool            -- ^ weather to perform SQLite vacuum to release space
   }
 
 newtype instance StorableQuery UtxoHandle =
-  UtxoAddress C.AddressAny deriving (Show, Eq)
+  UtxoAddress C.AddressAny deriving (Show, Eq, Ord)
 
 type QueryableAddresses = NonEmpty (StorableQuery UtxoHandle)
 
@@ -80,22 +93,24 @@ type instance StorablePoint UtxoHandle = C.ChainPoint
 newtype Depth = Depth Int
 
 data Utxo = Utxo
-  { _address          :: C.AddressAny
+  { _address          :: !C.AddressAny
   , _txId             :: !C.TxId
   , _txIx             :: !C.TxIx
-  , _datum            :: Maybe C.ScriptData
-  , _datumHash        :: Maybe (C.Hash C.ScriptData)
-  , _value            :: C.Value
-  , _inlineScript     :: Maybe Shelley.ScriptInAnyLang -- ^ Reference script
-  , _inlineScriptHash :: Maybe C.ScriptHash
+  , _datum            :: !(Maybe C.ScriptData)
+  , _datumHash        :: !(Maybe (C.Hash C.ScriptData))
+  , _value            :: !C.Value
+  , _inlineScript     :: !(Maybe C.ScriptInAnyLang)
+  , _inlineScriptHash :: !(Maybe C.ScriptHash)
   } deriving (Show, Eq, Generic)
 
 $(makeLenses ''Utxo)
 
 instance Ord Utxo where
-  left <= right
-    =   _txId left <= _txId right
-    &&  _txIx left <= _txIx right
+  compare (Utxo addr txid txix _ _ _ _ _) (Utxo addr' txid' txix' _ _ _ _ _) =
+    let cmp = addr `compare` addr'
+    in
+      if cmp == EQ then compare (C.TxIn txid txix) (C.TxIn txid' txix')
+      else  cmp
 
 instance FromJSON Utxo where
     parseJSON (Object v) =
@@ -111,10 +126,10 @@ instance FromJSON Utxo where
     parseJSON _ = mempty
 
 instance ToJSON Utxo where
-  toJSON (Utxo addr tId tIx dtum dtumHash val scrpt scrptHash) = object
+  toJSON (Utxo addr txid txix dtum dtumHash val scrpt scrptHash) = object
     [ "address"           .= addr
-    , "txId"              .= tId
-    , "txIx"              .= tIx
+    , "txId"              .= txid
+    , "txIx"              .= txix
     , "datum"             .= dtum
     , "datumHash"         .= dtumHash
     , "value"             .= val
@@ -124,9 +139,9 @@ instance ToJSON Utxo where
     ]
 
 data UtxoRow = UtxoRow
-  { _urUtxo      :: Utxo
-  , _urSlotNo    :: C.SlotNo
-  , _urBlockHash :: C.Hash C.BlockHeader
+  { _urUtxo      :: !Utxo
+  , _urSlotNo    :: !C.SlotNo
+  , _urBlockHash :: !(C.Hash C.BlockHeader)
   } deriving (Show, Eq, Ord, Generic)
 
 $(makeLenses ''UtxoRow)
@@ -155,22 +170,53 @@ data instance StorableEvent UtxoHandle = UtxoEvent
   , ueChainPoint  :: !C.ChainPoint
   } deriving (Eq, Ord, Show, Generic)
 
-eventIsBefore :: StorableEvent UtxoHandle  -> C.ChainPoint -> Bool
-eventIsBefore (UtxoEvent _ _ (C.ChainPoint _ slot)) (C.ChainPoint _ slot') =  slot < slot'
+eventIsBefore :: C.ChainPoint -> StorableEvent UtxoHandle -> Bool
+eventIsBefore (C.ChainPoint slot' _) (UtxoEvent _ _ (C.ChainPoint slot _)) =  slot <= slot'
 eventIsBefore _ _                                                          = False
 
+-- | mappend, combine Unspent utxoEvents:
+--
 instance Semigroup (StorableEvent UtxoHandle) where
-  e@(UtxoEvent u i cp) <> (UtxoEvent u' i' cp') =
-    if cp == cp' then
-      UtxoEvent (Set.union u u') (Set.union i i') cp
-    else e -- do not combine events from different chain points
+  (UtxoEvent us is cp) <> (UtxoEvent us' is' cp') =
+    UtxoEvent utxos txins (max cp cp')
+    where
+      toTxIn :: Utxo -> C.TxIn
+      toTxIn u = C.TxIn (u ^.txId) (u ^. txIx)
+      txins = Set.union is is'
+      utxos
+        = foldl' (\a c -> if toTxIn c `Set.notMember` txins then Set.insert c a;else a) Set.empty
+        . Set.union us
+        $ us'
 
 instance Monoid (StorableEvent UtxoHandle) where
   mempty = UtxoEvent mempty mempty C.ChainPointAtGenesis
 
+-- | The effect of a transaction (or a number of them) on the tx output set.
+data TxOutBalance =
+  TxOutBalance
+    { _tobUnspent :: !(Set C.TxIn)
+    -- ^ Outputs newly added by the transaction(s)
+    , _tobSpent   :: !(Set C.TxIn)
+    -- ^ Outputs spent by the transaction(s)
+    }
+    deriving stock (Eq, Show, Generic)
+
+instance Semigroup TxOutBalance where
+    tobL <> tobR =
+        TxOutBalance
+            { _tobUnspent = _tobUnspent tobR
+                         <> (_tobUnspent tobL `Set.difference` _tobSpent tobR)
+            , _tobSpent = _tobSpent tobL <> _tobSpent tobR
+            }
+
+instance Monoid TxOutBalance where
+    mappend = (<>)
+    mempty = TxOutBalance mempty mempty
+
+
 data Spent = Spent
-  { _sTxInTxId  :: !C.TxId                 -- ^ from TxIn, containts the Spent txId
-  , _sTxInTxIx  :: !C.TxIx
+  { _sTxId      :: !C.TxId
+  , _sTxIx      :: !C.TxIx
   , _sSlotNo    :: !C.SlotNo
   , _sBlockHash :: !(C.Hash C.BlockHeader)
   } deriving (Show, Eq)
@@ -179,19 +225,20 @@ $(makeLenses ''Spent)
 
 instance Ord Spent where
   compare spent spent' =
-      case  (spent ^. sTxInTxId) `compare` (spent' ^. sTxInTxId) of
-        EQ  -> (spent ^. sTxInTxIx) `compare` (spent' ^. sTxInTxIx)
+      case  (spent ^. sTxId) `compare` (spent' ^. sTxId) of
+        EQ  -> (spent ^. sTxIx) `compare` (spent' ^. sTxIx)
         neq -> neq
 
 instance HasPoint (StorableEvent UtxoHandle) C.ChainPoint where
   getPoint (UtxoEvent _ _ cp) = cp
 
----------------------------------------------------------------------------------
---------------- sql mappings unspent_transactions and Spent tables -------------
----------------------------------------------------------------------------------
+------------------
+-- sql mappings --
+------------------
+
 instance ToRow UtxoRow where
-  toRow u =
-    [ toField (u ^. urUtxo . address)
+  toRow u = toRow
+    ( toField (u ^. urUtxo . address)
     , toField (u ^. urUtxo . txId)
     , toField (u ^. urUtxo . txIx)
     , toField (u ^. urUtxo . datum)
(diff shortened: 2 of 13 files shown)


marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -80,29 +93,25 @@ type instance StorablePoint UtxoHandle = C.ChainPoint
...
-  { _address          :: C.AddressAny
-  , _txId             :: !C.TxId
-  , _txIx             :: !C.TxIx
-  , _datum            :: Maybe C.ScriptData
-  , _datumHash        :: Maybe (C.Hash C.ScriptData)
-  , _value            :: C.Value
-  , _inlineScript     :: Maybe Shelley.ScriptInAnyLang -- ^ Reference script
-  , _inlineScriptHash :: Maybe C.ScriptHash
+  { _address          :: !C.AddressAny
+  , _txIn             :: !C.TxIn
```
kayvank, line 97: @koslambrou this will change to JSON presentation to that of Cardano API.

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -135,8 +135,9 @@ utxoWorker_ callback depth maybeTargetAddresses Coordinator{_barrier} ch path =
       readMVar index >>= callback -- refresh the query STM/CPS with new storage pointers/counters state
       event <- atomically . readTChan $ ch
       case event of
-        RollForward (BlockInMode (Block (BlockHeader slotNo hsh _) txs) _) _ct -> do
-            let utxoEvents = Utxo.getUtxoEvents maybeTargetAddresses txs (C.ChainPoint slotNo hsh)
+        RollForward (BlockInMode block _) _ct -> do
+            -- let utxoEvents = Utxo.getUtxoEvents maybeTargetAddresses txs (C.ChainPoint slotNo hsh)
```
koslambrou, line 139: ?

marconi-chain-index/test-lib/Gen/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -136,29 +76,14 @@ utxoAddressOverride
   -> Utxo
 utxoAddressOverride addr utxo =  utxo {_address=C.toAddressAny addr}
 
--- | Generate ShelleyEra Utxo Events
-genShelleyEraUtxoEvents :: Gen [StorableEvent UtxoHandle]
+-- | generate Shelley Era Utxos
```
koslambrou, line 79: Why did the comment change?

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -3,82 +3,77 @@
...
-        "filter UtxoEvent for Utxos with address in the TargetAddress"
+        "marconi-chain-index-utxo filter UtxoEvent for Utxos with address in the TargetAddress"
         "eventsAtAddressTest"
         eventsAtAddressTest
 
     , testPropertyNamed
-        "marconi-utxo event-to-sqlRows property"
-        "eventsToRowsRoundTripTest"
+        "marconi-chain-index-utxo roundtrip event-to-sqlRows conversion property"
+        "UtxoEvents to Sql Row presentation round trip conversion"
```
koslambrou, line 49: IIRC That parameter needs to be the same name at the function. So you need to change for the tests below as name. Only the first parameter of `testPropertyNamed` gives a descriptive name.

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -96,68 +91,91 @@ tests = testGroup "Spec.Marconi.ChainIndex.Indexers.Utxo"
           "propJsonRoundtripUtxoRow"
           propJsonRoundtripUtxoRow
     ]
-
+-- | Round trip UtxoEvents to UtxoRow conversion
+-- The purpose of this test is to show that there is a isomorphism
+--  between `UtxoRow` and UtxoEent. Indeed UtxoRow is a presentation of the UtxoEvent.
```
koslambrou, line 96: Utxoeent -> UtxoEvent

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -96,68 +91,91 @@ tests = testGroup "Spec.Marconi.ChainIndex.Indexers.Utxo"
           "propJsonRoundtripUtxoRow"
           propJsonRoundtripUtxoRow
     ]
-
+-- | Round trip UtxoEvents to UtxoRow conversion
```
koslambrou, line 94: Add line break before this line.

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -96,68 +91,91 @@ tests = testGroup "Spec.Marconi.ChainIndex.Indexers.Utxo"
...
---     liftIO . Utxo.rowsToEvents (Utxo.getTxIns (getConn indexer) ) $ rows
---   Hedgehog.assert (equivalentLists
---                    computedEvent
---                    (filter (\e -> Utxo.ueChainPoint e /= C.ChainPointAtGenesis) events) )
+-- | create m ordered chainpoints
+-- SlotNo starts at 1.
+mkChainPoints :: C.SlotNo -> [C.ChainPoint]
+mkChainPoints m =
+  let
+    bs::ByteString
```
koslambrou, line 142: redundant

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -96,68 +91,91 @@ tests = testGroup "Spec.Marconi.ChainIndex.Indexers.Utxo"
...
---   results <- liftIO . traverse (Storable.query Storable.QEverything indexer) $ qs
---   let rows = concatMap (\(Utxo.UtxoResult rs) -> rs ) results
---   computedEvent <-
---     liftIO . Utxo.rowsToEvents (Utxo.getTxIns (getConn indexer) ) $ rows
---   Hedgehog.assert (equivalentLists
---                    computedEvent
---                    (filter (\e -> Utxo.ueChainPoint e /= C.ChainPointAtGenesis) events) )
+-- | create m ordered chainpoints
+-- SlotNo starts at 1.
+mkChainPoints :: C.SlotNo -> [C.ChainPoint]
```
koslambrou, line 139: If think there's a generator that already does this in the `test-lib*

marconi-sidechain/test/Spec/Marconi/Sidechain/Api/Routes/Golden/address-utxo-response.json
```diff
@@ -8,8 +8,7 @@
             "datumHash": null,
             "inlineScript": null,
             "inlineScriptHash": null,
-            "txId": "ec7d3bd7c6a3a31368093b077af0db46ceac77956999eb842373e08c6420f000",
-            "txIx": 0,
+            "txIn": "ec7d3bd7c6a3a31368093b077af0db46ceac77956999eb842373e08c6420f000#0",
```
koslambrou, line 11: This should go back to previous implementation.

koslambrou (commented):
First pass, I'll take a more in-depth look later. Which test actually makes the query fail without the fix?

marconi-sidechain/test/Spec/Marconi/Sidechain/Api/Query/Indexers/address-utxo-response.json
```diff
@@ -14,8 +14,7 @@
             "scriptLanguage": "PlutusScriptLanguage PlutusScriptV1"
         },
         "inlineScriptHash": "284d60f7e56f5fd54faed4c50fd5cab0307da1c4034d6a92c5dbb940",
-        "txId": "ec7d3bd7c6a3a31368093b077af0db46ceac77956999eb842373e08c6420f000",
-        "txIx": 0,
+        "txIn": "ec7d3bd7c6a3a31368093b077af0db46ceac77956999eb842373e08c6420f000#0",
```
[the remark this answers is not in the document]
kayvank: I'll change it back

marconi-chain-index/test-lib/Gen/Marconi/ChainIndex/Indexers/Utxo.hs
kayvank: Fixed.  It was my error as I had removed some code and this part stayed form the old code.

kayvank: I had re-arranged parts of this module and mixed this line with another function that I had taken out. It is fixed now.

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
kayvank: fixed

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
kayvank: I changed this back to previous presentation

koslambrou:
> First pass, I'll take a more in-depth look later. Which test actually makes the query fail without the fix?

@kayvank

kayvank:
> 

@koslambrou  Roundtrip store & retrieve test(s) were failing for depth greater than the number of events. In particular, [this one](https://github.com/kayvank/plutus-apps/blob/kayvan/1536/utxo-test-cases/marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs#L59) always failed. 
I saved the failure and will attach it to the ticket

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -97,72 +84,84 @@ tests = testGroup "Spec.Marconi.ChainIndex.Indexers.Utxo"
...
+  let depth = Utxo.Depth 2160 -- we set a high valued for depth, to make sure we only test in-memory buffer
   events <- forAll UtxoGen.genUtxoEvents
   (storedEvents :: [StorableEvent Utxo.UtxoHandle]) <-
-    (liftIO . Utxo.open ":memory:") (Utxo.Depth 10)
+    (liftIO $ Utxo.open ":memory:" depth False) -- don't vacuum sqlite
      >>= liftIO . Storable.insertMany events
      >>= liftIO . Storable.getEvents
-  Hedgehog.assert (equivalentLists storedEvents events)
-
--- Insert Utxo events in storage, and retrieve the events by address
```
kayvank, line 123: @koslambrou  this test always failed too and was commented out .  
it is replaced with `propUtxoQueryByAddressAndQueryInterval`

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -97,72 +84,84 @@ tests = testGroup "Spec.Marconi.ChainIndex.Indexers.Utxo"
...
-  indexer <- liftIO $ Utxo.open ":memory:" (Utxo.Depth 2)
-             >>= liftIO . Storable.insertMany [event0, event1, event2, event3]
-  let
-    qs :: [StorableQuery Utxo.UtxoHandle]
-    qs = fmap (Utxo.UtxoAddress . Utxo._address) . concatMap (Set.toList . Utxo.ueUtxos) $ events
-  results <- liftIO . traverse (Storable.query (Storable.QInterval (head chainpoints)(chainpoints !! 1)) indexer) $ qs
-  let rows = concatMap (\(Utxo.UtxoResult rs) -> rs ) results
-  computedEvent <-
-    liftIO . Utxo.rowsToEvents (Utxo.getTxIns (getConn indexer) ) $ rows
-  Hedgehog.assert (equivalentLists computedEvent [event0, event1])
```
kayvank, line 160: @koslambrou this is another example of previously failing tests.  This test was passing since the `equivalentLists` was not really comparing anything

berewt (commented):
Minors comments, I didn't follow the implementation carefully enough to comment the rest of the changes.
(and thanks for adding me to the review, it will indeed impact my current PR.)

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
koslambrou: Actually, my question is a bit different. Most of the existing tests were failing because of the implementation of `equivalentLists`. I know you fixed that in this PR and then correctly changed the property tests. My question is more: if you revert back the fix on the implementation of the querying, is there any of the existing property tests that detect that failure. Basically, have you TDD'd this. I just want to make sure you have a regression test.

kayvank: Yes, I belive so. Here is what I did:
This is the test failure I started with:
```  ┏━━ test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs ━━━
            126 ┃ utxoInsertAndQueryTest :: Property
            127 ┃ utxoInsertAndQueryTest = property $ do
            128 ┃   events <- forAll genUtxoEvents
                ┃   │ [ UtxoEvent
                ┃   │     { ueUtxos =
                ┃   │         fromList
                ┃   │           [ Utxo
                ┃   │               { _address =
                ┃   │                   AddressByron
                ┃   │                     (ByronAddress
                ┃   │                        Address
                ┃   │                          { addrRoot =
                ┃   │                              c75ac0f6c2a478087d68308dbbc5f533395f68dbb6d581ffb8cb562b
                ┃   │                          , addrAttributes =
                ┃   │                              Attributes
                ┃   │                                { data_ =
                ┃   │                                    AddrAttributes
                ┃   │                                      { aaVKDerivationPath = Nothing
                ┃   │                                      , aaNetworkMagic = NetworkMainOrStage
                ┃   │                                      }
                ┃   │                                }
                ┃   │                          , addrType = ATVerKey
                ┃   │                          })
```
The failure was from `bad` event -> utxoRow conversion.   

fixed that, and replaced the equivalent test with `===`, and all tests passed.

fixing that function and replacing the equivalent tests,  fixed the remaining tests

koslambrou: @kayvank Tried reverting your fix to the old implementation, while keeping all your property tests. None fail with the old implementation, therefore doesn't seem like you added a regression test. Can you validate?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -80,22 +93,23 @@ type instance StorablePoint UtxoHandle = C.ChainPoint
...
 $(makeLenses ''Utxo)
 
+
 instance Ord Utxo where
-  left <= right
-    =   _txId left <= _txId right
-    &&  _txIx left <= _txIx right
+  compare (Utxo addr txid txix _ _ _ _ _) (Utxo addr' txid' txix' _ _ _ _ _) =
+    if addr == addr' then compare (C.TxIn txid txix) (C.TxIn txid' txix')
+    else EQ
```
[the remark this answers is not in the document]
koslambrou: If the tests didn't test this as a regression, there should be a test case for that.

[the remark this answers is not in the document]
kayvank: @koslambrou I think this is resolved now with the regression test on [Draft PR-1032](https://github.com/input-output-hk/plutus-apps/pull/1032)

koslambrou:
The other PR has been merged: https://github.com/input-output-hk/plutus-apps/pull/1034, so a rebase is necessary.

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -138,13 +138,13 @@ datumWorker Coordinator{_barrier, _channel} path = do
...
   -> Utxo.Depth
-  -> Maybe TargetAddresses                       -- ^ Target addresses to filter for
+  -> Maybe TargetAddresses        -- ^ Target addresses to filter for
   -> Coordinator
   -> TChan (ChainSyncEvent (BlockInMode CardanoMode))
   -> FilePath
   -> IO (IO (), MVar Utxo.UtxoIndexer)
 utxoWorker_ callback depth maybeTargetAddresses Coordinator{_barrier} ch path = do
-  ix <- Utxo.open path depth
+  ix <- Utxo.open path depth True -- open SQLite with depth=depth and perform SQLite vacuum
```
kayvank, line 147: if this change is accepted, I'd like to added to the CLI parameters for both:
- marconi-chain-index
- marconi-sidechain
on a separate PR

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -201,37 +248,36 @@ instance ToRow UtxoRow where
...
 -- | Open a connection to DB, and create resources
 -- The parameter ((k + 1) * 2) specifies the amount of events that are buffered.
 -- The larger the number, the more RAM the indexer uses. However, we get improved SQL
 -- queries due to batching more events together.
 open
-  :: FilePath -- ^ sqlite file path
-  -> Depth    -- ^ The Depth parameter k, the larger K, the more RAM the indexer uses
+  :: FilePath   -- ^ sqlite file path
+  -> Depth      -- ^ The Depth parameter k, the larger K, the more RAM the indexer uses
+  -> Bool       -- ^ weather to perform vacuum
```
kayvank, line 278: if this change is accepted, I'd like to added to the CLI parameters for both:
- marconi-chain-index
- marconi-sidechain
on a separate PR

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -286,37 +333,57 @@ instance Buffered UtxoHandle where
...
-      SQL.execute_ c [r|DELETE FROM unspent_transactions
-                          WHERE unspent_transactions.rowid IN
-                            (SELECT unspent_transactions.rowid
-                             FROM unspent_transactions
-                               LEFT JOIN spent ON
-                                 unspent_transactions.txId = spent.txInTxId
-                               AND unspent_transactions.txIx = spent.txInTxIx
-                             WHERE spent.txInTxId IS NOT NULL)|]
-      SQL.execute_ c "VACUUM"
+    when (toVacuume h) $ do
```
berewt, line 360: The comment is incorrect: we do it randomly on a mean of once every 100.
As discussed, we need to be more cautious about it, especially when disk contains events that can be rollbacked but may be not in this PR (we can at least use `toVacuume = False` now)

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
berewt, line 147: Either add a parameter for `vacuum` or set it to `False`. It's buggy on rollback if our depth is too low.

berewt (approved):
LGTM aside the requested changes.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -80,22 +93,24 @@ type instance StorablePoint UtxoHandle = C.ChainPoint
...
 
 instance Ord Utxo where
-  left <= right
-    =   _txId left <= _txId right
-    &&  _txIx left <= _txIx right
+  compare (Utxo addr txid txix _ _ _ _ _) (Utxo addr' txid' txix' _ _ _ _ _) =
+    let cmp = addr `compare` addr'
+    in
+      if cmp == EQ then compare (C.TxIn txid txix) (C.TxIn txid' txix')
+      else  cmp
```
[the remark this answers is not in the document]
kayvank: Thank you :)

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
kayvank: good catch, thank you. I fixed the comment.
 I agree that if we're going to vacuum, we need more intelligent way of doing it.
I also have no objection to setting the flag `toVaccume=False` by default, unless the user instruct us otherwise through CLI params that yet to be added!

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
kayvank: I added TODO item to allow user decide to vacuum or not. And yes, at low depth, this is an issue and the tests clearly show that

berewt: Is it that bad performance-wise to disable it by default at the moment?
If we keep it, we may have to add a hack to tests that deal with rollbacks until we find a solution, and it's easy to forget to remove the hack once it's fixed.

berewt: Sorry, I didn't see the previous comment. :)

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -155,22 +170,53 @@ data instance StorableEvent UtxoHandle = UtxoEvent
...
+    -- ^ Outputs spent by the transaction(s)
+    }
+    deriving stock (Eq, Show, Generic)
+
+instance Semigroup TxOutBalance where
+    tobL <> tobR =
+        TxOutBalance
+            { _tobUnspent = _tobUnspent tobR
+                         <> (_tobUnspent tobL `Set.difference` _tobSpent tobR)
+            , _tobSpent = _tobSpent tobL <> _tobSpent tobR
```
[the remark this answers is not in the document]
kayvank: yes, that will save us some RAM

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -155,43 +167,75 @@ data instance StorableEvent UtxoHandle = UtxoEvent
...
   } deriving (Eq, Ord, Show, Generic)
 
-eventIsBefore :: StorableEvent UtxoHandle  -> C.ChainPoint -> Bool
-eventIsBefore (UtxoEvent _ _ (C.ChainPoint _ slot)) (C.ChainPoint _ slot') =  slot < slot'
+eventIsBefore :: C.ChainPoint -> StorableEvent UtxoHandle -> Bool
+eventIsBefore (C.ChainPoint slot' _) (UtxoEvent _ _ (C.ChainPoint slot _)) =  slot <= slot'
 eventIsBefore _ _                                                          = False
 
+-- | mappend, combine Unspent utxoEvents:
+--
```
koslambrou, line 175: Bin?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -155,43 +167,75 @@ data instance StorableEvent UtxoHandle = UtxoEvent
...
-      UtxoEvent (Set.union u u') (Set.union i i') cp
-    else e -- do not combine events from different chain points
+  (UtxoEvent us is cp) <> (UtxoEvent us' is' cp') =
+    UtxoEvent utxos txins (max cp cp')
+    where
+      toTxIn :: Utxo -> C.TxIn
+      toTxIn u = C.TxIn (u ^.txId) (u ^. txIx)
+      txins = Set.union is is'
+      utxos
+        = foldl' (\a c -> if toTxIn c `Set.notMember` txins then Set.insert c a;else a) Set.empty
```
koslambrou, line 184: `Set.insert c a;else a` -> `Set.insert c a; else a`

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -155,43 +167,75 @@ data instance StorableEvent UtxoHandle = UtxoEvent
...
+        . Set.union us
+        $ us'
```
koslambrou, line 186: Minor detail. I know we don't have a coding style guide, but I suggest doing something like `foldl' ... $ Set.union us us'` instead of using function composition, because that is how we usually write it in the plutus-apps codebase.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -201,37 +245,36 @@ instance ToRow UtxoRow where
...
 -- | Open a connection to DB, and create resources
 -- The parameter ((k + 1) * 2) specifies the amount of events that are buffered.
 -- The larger the number, the more RAM the indexer uses. However, we get improved SQL
 -- queries due to batching more events together.
 open
-  :: FilePath -- ^ sqlite file path
-  -> Depth    -- ^ The Depth parameter k, the larger K, the more RAM the indexer uses
+  :: FilePath   -- ^ sqlite file path
+  -> Depth      -- ^ The Depth parameter k, the larger K, the more RAM the indexer uses
+  -> Bool       -- ^ weather to perform vacuum
```
koslambrou, line 275: weather -> whether

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
koslambrou: You can create a story for that.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -286,37 +330,57 @@ instance Buffered UtxoHandle where
...
+                            unspent_transactions
+                          WHERE
+                            unspent_transactions.rowid IN (
+                              SELECT
+                                unspent_transactions.rowid
+                              FROM
+                                unspent_transactions
+                                JOIN spent ON unspent_transactions.txId = spent.txId
+                                AND unspent_transactions.txIx = spent.txIx
+                            )|] >> SQL.execute_ c "VACUUM" -- remove Spent and release space, see https://www.sqlite.org/lang_vacuum.html
```
koslambrou, line 370: For readability, do instead:
```
      when (rndCheck == 42) $ do
        SQL.execute_ c [r|DELETE FROM
                            unspent_transactions
                          WHERE
                            unspent_transactions.rowid IN (
                              SELECT
                                unspent_transactions.rowid
                              FROM
                                unspent_transactions
                                JOIN spent ON unspent_transactions.txId = spent.txId
                                AND unspent_transactions.txIx = spent.txIx
                            )|]
        -- remove Spent and release space, see https://www.sqlite.org/lang_vacuum.html
        SQL.execute_ c "VACUUM"
```

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -407,74 +508,83 @@ instance Queryable UtxoHandle where
     -> f (StorableEvent UtxoHandle)
     -> UtxoHandle
     -> StorableQuery UtxoHandle
-    -> StorableMonad UtxoHandle (StorableResult UtxoHandle)
-  queryStorage qi es (UtxoHandle c _) q@(UtxoAddress addr) = do
-    persisted <- case qi of
+    -> IO (StorableResult UtxoHandle)
+  queryStorage qi es (UtxoHandle c _ _) q@(UtxoAddress addr) = do
+    fromSQL :: [UtxoRow] <- case qi of -- get the unspent transaction from DB
```
koslambrou, line 513: fromSQL -> persistedUtxoRows

marconi-chain-index/src/Marconi/ChainIndex/Orphans.hs
```diff
@@ -137,6 +137,9 @@ mapLeft _ (Right v) = Right v
 instance ToJSON C.ScriptData where
     toJSON v = Aeson.String $ Text.decodeLatin1 $ Base16.encode $ C.serialiseToCBOR v
 
+instance SQL.FromRow Int where
+  fromRow = SQL.field
```
koslambrou, line 141: I wouldn't add that instance. If your select query just returns a single element which is an int, then the type would simply be `[[Int]]`, right? There is already a `FromField Int` instance, so let's use that.

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -3,142 +3,236 @@
...
 import Test.Tasty.Hedgehog (testPropertyNamed)
 
 tests :: TestTree
 tests = testGroup "Spec.Marconi.ChainIndex.Indexers.Utxo"
-    [ testPropertyNamed
-        "filter UtxoEvent for Utxos with address in the TargetAddress"
-        "eventsAtAddressTest"
-        eventsAtAddressTest
+      [ testPropertyNamed
+        "marconi-chain-index-utxo regression-test, store and retrieve UtxoEvents by address, then test that there are no `Spent` in the retrieved Utxo(s)"
```
koslambrou, line 42: No need to say it's a regression test.

Also, no need to put "marconi-chain-index-utxo" for everything. The `testGroup` already mentions what module we're testing.

How about a test description like: "All queried UTXOs by address should be unspent.

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -3,142 +3,236 @@
...
+        "allqueryUtxosShouldBeUnspent"
+        allqueryUtxosShouldBeUnspent
+
+      , testPropertyNamed
+        "marconi-chain-index-utxo compute utxos with address in TargetAddresses property"
```
koslambrou, line 47: Same comment as above.

I'll stop repeating, but all tests below need to be changed.

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -3,142 +3,236 @@
...
+-- The purpose of this test is to show that there is a isomorphism between `UtxoRow` and UtxoEvent.
+propRoundTripEventsToRowConversion :: Property
+propRoundTripEventsToRowConversion  = property $ do
+  events <- forAll UtxoGen.genUtxoEvents
+  let
+    txInsMap :: Map C.SlotNo (Set C.TxIn)
+    txInsMap
+      = Data.Map.fromList
+      . concatMap g
+      $ events
```
koslambrou, line 160: Same comment as above. Don't use function composition if you already have the argument to be applied

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -3,142 +3,236 @@
...
+    f :: C.SlotNo -> IO (Set C.TxIn)
+    f sn = pure $ fromMaybe Set.empty (Data.Map.lookup sn txInsMap)
+    g :: StorableEvent Utxo.UtxoHandle -> [(C.SlotNo, Set C.TxIn)]
+    g (Utxo.UtxoEvent _ ins (C.ChainPoint sn _)) = [ (sn, ins)]
```
koslambrou, line 164: [ (sn, ins)] -> [(sn, ins)]

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -238,37 +335,51 @@ propResumingShouldReturnAtLeastOneNonGenesisPointIfStoredOnDisk = property $ do
...
+  List.reverse (List.sort resumablePoints) === resumablePoints
 
 propJsonRoundtripUtxoRow :: Property
 propJsonRoundtripUtxoRow = property $ do
     utxoEvents <- forAll genUtxoEvents
-    let utxoRows = concatMap Utxo.eventsToRows utxoEvents
+    let utxoRows = concatMap Utxo.eventToRows utxoEvents
     forM_ utxoRows $ \utxoRow -> Hedgehog.tripping utxoRow Aeson.encode Aeson.decode
+
+-- -- getUtxoEven should compute the same event as the event generator
```
koslambrou, line 369: getUtxoEvents

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -3,142 +3,236 @@
...
           "propResumingShouldReturnOrderedListOfPoints"
           propResumingShouldReturnOrderedListOfPoints
 
     , testPropertyNamed
           "ToJSON/FromJSON roundtrip for UtxoRow"
           "propJsonRoundtripUtxoRow"
           propJsonRoundtripUtxoRow
+
+    , testPropertyNamed
+          "marconi-chain-index-utxo getUtxoEvent should get the same event from the Block as the event generator"
```
koslambrou, line 92: getUtxoEvents

marconi-sidechain/test/Spec/Marconi/Sidechain/Api/Query/Indexers/Utxo.hs
```diff
@@ -83,23 +84,24 @@ initRpcTestEnv = do
...
-  :: (AddressUtxoIndexer.UtxoIndexer -> IO ())
-  -> [Utxo.StorableEvent Utxo.UtxoHandle]
+  :: (AddressUtxoIndexer.UtxoIndexer -> IO ()) -- ^  callback to be refreshed
+  -> [Utxo.StorableEvent Utxo.UtxoHandle] -- ^  events to store
   -> IO ()
-mocUtxoWorker callback events  =
-  Utxo.open ":memory:" depth >>= Storable.insertMany events >>= callback
+mocUtxoWorker callback events =
+  let depth :: Utxo.Depth
+      depth = Utxo.Depth (1 + length events) -- use in-memory- store
```
koslambrou, line 96: in-memory- -> in-memory

marconi-sidechain/test/Spec/Marconi/Sidechain/Api/Query/Indexers/Utxo.hs
```diff
@@ -83,23 +84,24 @@ initRpcTestEnv = do
...
+  in
+    Utxo.open ":memory:" depth False >>= Storable.insertMany events >>= callback
+
+-- | generate some Utxo events, store them and fetch the Unspent Utxos, and make sure JSON conversion is idempotent
```
koslambrou, line 100: Unspent Utxos -> Utxos

The Unspent is redundant.

koslambrou (approved):
Approving, but some minor changes that need to be fixed.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
koslambrou: Also, we don't want that for `marconi-sidechain` as they need to know all the UTXOs that ever occurred, even the ones that are spent on the latest slot.

kayvank:
created https://input-output.atlassian.net/browse/PLT-5460 for the toVacuum flag

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
kayvank: the switch is set at utxoworker in indexers.hs and covers this case.
This will change once we add the CLI parameter such that users will control that

marconi-chain-index/src/Marconi/ChainIndex/Orphans.hs
kayvank: Yes, I forgot to remove it. I had added in when I needed to use the [Exist](https://sqlite.org/lang_expr.html) operator to see if an Spent existed.

PR #1038, opened 2023-03-29
PLT-5459 Fix resuming on the MintBurn indexer

koslambrou:
* Fixed a syntax error of the `getStoredEvents` SQL query.

* Fixed the `getStoredEvents` SQL query so that the points returned are in descending order.

* Added a property test which verifies that the ChainPointAtGenesis is always returned by `resumablePoints`.

* Added a property test which verifies that the points returned by `resumablePoints` are sorted in descending order.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/AddressDatum/AddressDatumIndex.hs
+++ b/marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/AddressDatum/AddressDatumIndex.hs
@@ -27,6 +27,7 @@ import Marconi.ChainIndex.Indexers.AddressDatum (AddressDatumDepth (AddressDatum
                                                  StorableQuery (AddressDatumQuery, AddressDatumQuery, AllAddressesQuery),
                                                  StorableResult (AddressDatumResult, AllAddressesResult))
 import Marconi.ChainIndex.Indexers.AddressDatum qualified as AddressDatum
+import Marconi.ChainIndex.TestLib.StorableProperties qualified as StorableProperties
 import Marconi.Core.Storable qualified as Storable
 import Spec.Marconi.ChainIndex.Indexers.AddressDatum.Utils (addressInEraToAddressAny)
 import Test.Tasty (TestTree, localOption, testGroup)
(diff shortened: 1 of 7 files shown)


marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/AddressDatum/AddressDatumIndex.hs
```diff
@@ -341,6 +346,20 @@ propResumingShouldReturnAtLeastOneNonGenesisPointIfStoredOnDisk = property $ do
     resumablePoints <- liftIO $ Storable.resumeFromStorage $ finalIndex ^. Storable.handle
     Hedgehog.assert $ length resumablePoints >= 2
 
+-- | The property verifies that the 'Storable.resumeFromStorage' call returns at least a point which
```
koslambrou, line 349: I just copy-pasted this from the MintBurn indexer. Ideally, we would have a module for common property tests, but we will work on it later.

eyeinsky: We have the `marconi-chain-index-test-lib` library component in marconi-chain-index package, maybe it can already be put and imported from there?

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/MintBurn.hs
```diff
@@ -59,8 +62,17 @@ tests = testGroup "MintBurn"
...
   , testPropertyNamed
-      "Querying a resumed indexer returns only the persisted events"
-      "resume" resume
+      "Querying a recreated indexer should only the persisted events, and not the in-memory events of the initial indexer"
+      "propRecreatingIndexerFromDiskShouldOnlyReturnPersistedEvents "
+      propRecreatingIndexerFromDiskShouldOnlyReturnPersistedEvents
+  , testPropertyNamed
+      "The points the indexer can be resumed from should return at least the genesis point"
+      "propResumingShouldReturnAtLeastTheGenesisPoint"
+      propResumingShouldReturnAtLeastTheGenesisPoint
```
eyeinsky, line 71: Perhaps this is a future story but if genesis is always a valid resume point then why include it at all when returning from `resumeFromStorage` and then check it with a test? Return [] instead, and when that happensis then the indexer can start from genesis. Saves from writing code both in the indexers and these tests.

koslambrou: You're right. Will create a story for this.

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/AddressDatum/AddressDatumIndex.hs
koslambrou: Yep. Generalized the property and moved it to `test-lib` package.

PR #1043, opened 2023-03-30
Query security parameter from the local node, pass it to all indexers

eyeinsky:
Pre-submit checklist:
- Branch
- PR
    - [X] Self-reviewed the diff
    - [X] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
@@ -34,7 +34,6 @@ import Data.Map (Map)
 import Data.Map qualified as Map
 import Data.Maybe (fromMaybe, mapMaybe)
 import Data.Text qualified as TS
-import Data.Word (Word64)
 import Marconi.ChainIndex.Indexers.AddressDatum (AddressDatumDepth (AddressDatumDepth), AddressDatumHandle,
                                                  AddressDatumIndex)
 import Marconi.ChainIndex.Indexers.AddressDatum qualified as AddressDatum
@@ -49,7 +48,7 @@ import Marconi.ChainIndex.Logging (logging)
 import Marconi.ChainIndex.Node.Client.GenesisConfig (NetworkConfigFile (NetworkConfigFile), initExtLedgerStateVar,
                                                      mkProtocolInfoCardano, readCardanoGenesisConfig, readNetworkConfig,
                                                      renderGenesisConfigError)
-import Marconi.ChainIndex.Types (TargetAddresses)
+import Marconi.ChainIndex.Types (SecurityParam, TargetAddresses)
 import Marconi.ChainIndex.Utils qualified as Utils
 import Marconi.Core.Index.VSplit qualified as Ix
 import Marconi.Core.Storable qualified as Storable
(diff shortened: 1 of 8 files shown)


marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -110,11 +110,11 @@ initialCoordinator indexerCount =
 -- The points should/could provide shared access to the indexers themselves. The result
 -- is a list of points (rather than just one) since it offers more resume possibilities
 -- to the node (in the unlikely case there were some rollbacks during downtime).
-type Worker = Coordinator -> FilePath -> IO [Storable.StorablePoint ScriptTx.ScriptTxHandle]
+type Worker = Int -> Coordinator -> FilePath -> IO [Storable.StorablePoint ScriptTx.ScriptTxHandle]
```
koslambrou, line 113: Can we use `Word64` as that's the type used by consensus as well?

Also, how about a newtype for this?

marconi-chain-index/src/Marconi/ChainIndex/Utils.hs
```diff
@@ -1,15 +1,39 @@
...
+    eraInMode :: C.EraInMode C.BabbageEra C.CardanoMode
+    eraInMode = C.BabbageEraInCardanoMode
+
+    queryInMode :: C.QueryInMode C.CardanoMode (Either EraMismatch C.GenesisParameters)
+    queryInMode = C.QueryInEra eraInMode
+      $ C.QueryInShelleyBasedEra (C.shelleyBasedEra @C.BabbageEra) C.QueryGenesisParameters
+
+    -- TODO /Really/ handle the error.
+    showError :: Show a => a -> b
+    showError = error . show
```
koslambrou, line 39: Why not handle it in this PR? Do you think the resulting change will be too big?

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -385,18 +385,19 @@ mintBurnWorker_ bufferSize onInsert Coordinator{_barrier} ch dbPath = do
...
   void $ forkIO loop
   readMVar ix >>= Storable.resumeFromStorage . view Storable.handle
 
 initializeIndexers
-  :: [(Worker, FilePath)]
+  :: Int
+  -> [(Worker, FilePath)]
   -> IO ([ChainPoint], Coordinator)
-initializeIndexers indexers = do
+initializeIndexers securtyParam indexers = do
```
koslambrou, line 398: Typo

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -385,18 +385,19 @@ mintBurnWorker_ bufferSize onInsert Coordinator{_barrier} ch dbPath = do
...
   coordinator <- initialCoordinator $ length indexers
-  startingPoints <- mapM (\(ix, fp) -> ix coordinator fp) indexers
+  startingPoints <- mapM (\(ix, fp) -> ix securtyParam coordinator fp) indexers
```
koslambrou, line 400: Typo

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
berewt: I'd avoid a newtype here: worker is really a type alias, I don't see a scenario where we'd like to distinguish `Worker` from its alias  (but I'd be happy to change my mind if you have one), which is the main (unique/) purpose of a newtype.

marconi-chain-index/src/Marconi/ChainIndex/Utils.hs
eyeinsky: Using `error` is what we do elsewhere as well. So as we don't have any specific way to do it then how do we do it?

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
eyeinsky: @koslambrou Newtype for the security parameter you mean?

koslambrou: Yes :) Newtype for the security param, not the worker.

marconi-chain-index/src/Marconi/ChainIndex/Utils.hs
koslambrou: Alright. Then that means we need a story for error handling

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
berewt: My bad, so yes.

eyeinsky: 21830c7d1

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
eyeinsky: 7d440028c

koslambrou (approved):

PR #1046, opened 2023-03-31
PLT-1397 MintBurnTx indexer now supports querying data of specific points in time

koslambrou:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


berewt (approved):
Just a minor remark about code duplication for sql queries.
In general, if the queries are filtering variations on the same select query, I think it's worthy to implement it this way.

koslambrou:
> In general, if the queries are filtering variations on the same select query, I think it's worthy to implement it this way.

Can you clarify? Are you talking about removing the duplication between sql queries, and just have a single one, but with different filter statements?

berewt:
Yes, I was sure there was a comment attached to one of the query function but it seems like I forgot to include it.
And yes, the idea is to pass a parameter type to the query, that would add the corresponding `where` clause to the query.

koslambrou:
@eyeinsky Anything to add? I'd like to merge this before end of sprint

koslambrou:
@berewt I applied the changes we discussed. Also added two more property tests for verifying that calling the query without a slot is the same as querying with the latest indexed slot number.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
```diff
@@ -56,29 +57,29 @@ import Ouroboros.Consensus.Shelley.Eras qualified as OEra
 -- * Event
 
 data TxMintEvent = TxMintEvent
-  { txMintEventSlotNo          :: C.SlotNo
-  , txMintEventBlockHeaderHash :: C.Hash C.BlockHeader
-  , txMintEventTxAssets        :: NE.NonEmpty (C.TxId, NE.NonEmpty MintAsset)
+  { txMintEventSlotNo          :: !C.SlotNo
```
eyeinsky, line 61: Do we want these bangs? If we do then there is the `StrictData` language extension that does it for all data constructor fields. But as far as I remember, these help with not building up big thunks in deep loops (like summing a very long list, but not evaluating the sum), but all of our fields are evaluated when the buffer is flushed to database. Also, as the buffer is empty after that then there is no possibility of having a thunk there.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
```diff
@@ -292,35 +287,58 @@ type instance RI.StorableMonad MintBurnHandle = IO
...
-                redeemerData
-
-      interval = intervalToWhereClause queryInterval
+  queryStorage queryInterval memoryEvents (MintBurnHandle sqlCon _k) query =
+      case query of
+        QueryAllMintBurn Nothing -> toResult Nothing <$> queryStoredTxMintEvents sqlCon interval
+        QueryAllMintBurn (Just slotNo) -> do
+            let conditions = interval <> (["slotNo <= :slotNo"], [":slotNo" := slotNo])
+            toResult (Just slotNo) <$> queryStoredTxMintEvents sqlCon conditions
+        QueryByAssetId policyId assetName Nothing -> do
```
eyeinsky, line 308: I think this and the following case can be joined by just adding the slotNo element to the conditions list?

eyeinsky (commented):
The Jira issue text is about another indexer, but the feature to be added is the same, right?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
```diff
@@ -292,35 +287,58 @@ type instance RI.StorableMonad MintBurnHandle = IO
...
+            let conditions = interval
+                          <> ( ["policyId = :policyId AND assetName = :assetName"]
+                             , [":policyId" := policyId, ":assetName" := assetName]
+                             )
+            toResult Nothing <$> queryStoredTxMintEvents sqlCon conditions
+        QueryByAssetId policyId assetName (Just slotNo) -> do
+            let conditions = interval
+                          <> ( ["slotNo <= :slotNo AND policyId = :policyId AND assetName = :assetName"]
```
eyeinsky, line 316: Should this be `slotNo = :slotNo`? Because if events up to `slotNo` are required then the interval part already does that if you add only the toSlot part. Or maybe I'm still confused as to what the interval is meant to in the indexer framework and what the jira issue is specifying.

koslambrou: I'm ignoring the QInterval stuff. That should be deleted, as the semantics of it is not clear.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
koslambrou: Good idea

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
koslambrou: Just good practice. I agree they might not be necessary, but there's no harm in using them.

berewt (approved):
My only remark was about factorizing the `Nothing/Just x` case when we process the query, as suggested by @eyeinsky so it's good for me.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
koslambrou: d32ea77fe3de6f145417fa5588b9d82b41760fda

koslambrou:
> The Jira issue text is about another indexer, but the feature to be added is the same, right?

You're right. The Jira ticket description was a copy-paste. It was actually meant for MintBurn, and not the Utxo indexer.

PR #1047, opened 2023-04-03
PLT-273 Add slot to utxo query

berewt:
We can now provide a upper slot when querying the utxos.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -257,16 +262,46 @@ propUtxoQueryByAddressAndQueryInterval = property $ do
 
   last slotNoFromStorage === endIntervalSlotNo
 
+
+-- Insert Utxo events in storage, and retreive the events by address and query interval
```
kayvank, line 266: I think query interval should be changed to slot ?

marconi-sidechain/src/Marconi/Sidechain/Api/Query/Indexers/Utxo.hs
```diff
@@ -41,39 +44,50 @@ initializeEnv targetAddresses = do
 findByAddress
     :: AddressUtxoIndexerEnv           -- ^ Query run time environment
```
kayvank, line 45: remove extra spaces please. This was my fault from precocious commits :(

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -257,16 +262,46 @@ propUtxoQueryByAddressAndQueryInterval = property $ do
 
   last slotNoFromStorage === endIntervalSlotNo
 
+
+-- Insert Utxo events in storage, and retreive the events by address and slot
+-- Note: The property we are checking is:
+--   - Insert many events with at various chainPoints
+--   - Fetch for all the evenent addresses in the ChainPoint interval [lowChainPoint, highChainPoint]
```
koslambrou, line 269: evenent -> event

koslambrou (commented):
I'd like to propose the following alternative:

How about keeping the old query which queries everything and add the new one as a sum type?

Then, in the property test, you can do something similar to what I did in the MintBurn indexer PR which you reviewed, which is just compare the two queries between each other.

Also,

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
berewt: Do you mean:
```
data instance StorableQuery UtxoHandle =
   UtxoAddress C.AddressAny | UtxoAddressAtSlot C.AddressAny C.SlotNo
```

I can, the two representations are isomorphic though.
I can compare a query to `UtxoAddress addr Nothing` and another one to `UtxoAddress addr (Just lastSlot)` with the current query.

I think having only one constructor for both cases made the processing easier. Separating them will add an extra resolution step before building the query, or will lead in query duplication. I won't die on this hill though, so if you think it's valuable, I can change it, your take.

berewt: As we discussed, I borrowed your idea of using the query to generate the where clauses and switched to `NamedQuery`.

berewt: I'll probably add a few tests tomorrow. I let you decide if we can consider this as resolved or not.

PR #1054, opened 2023-04-14
Definition of contract model for the Governance contract

Ali-Hill:
I have defined a contract model to test the governance contract. 

Changes in law do not have a monetary impact so it was not possible to check this without adding an additional off-chain endpoint called `CheckLaw`. This endpoint simply queries the state of the contract and has no on-chain impact. Work will soon be done by Quviq so that this no longer needs to be done using off-chain code.

I have enabled code coverage for this contract so the transaction size has increased. I therefore had to add additional transaction limits to the existing tests. 

I have added comments to the contract model and all tests have passed on my system. 

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


sjoerdvisscher (approved):
Looks great, thanks!

PR #1055, opened 2023-04-17
ContractModel: add the ability to perform observations of the blockchain in `DL` properties

MaximilianAlgehed:
This fixes the need for custom endpoints as used e.g. in the Governance example.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/plutus-contract/src/Plutus/Contract/Test/ContractModel/Interface.hs
+++ b/plutus-contract/src/Plutus/Contract/Test/ContractModel/Interface.hs
@@ -77,6 +77,7 @@ module Plutus.Contract.Test.ContractModel.Interface
     , DL
     , action
     , waitUntilDL
+    , observeChain
     , QCCM.anyAction
     , QCCM.anyActions
     , QCCM.anyActions_
@@ -477,6 +478,14 @@ waitUntilDL = QCCM.waitUntilDL . toSlotNo
 
 assertModel :: String -> (QCCM.ModelState state -> Bool) -> DL state ()
 assertModel s p = QCCM.assertModel s (p . fmap coerce)
+
+-- IMPORTANT NOTE: the first argument needs to be unique among calls to `observeChain`
+-- as it is a proxy for the predicate.
+observeChain :: ContractModel state
+             => String
+             -> ((QCCM.SymToken -> CardanoAPI.AssetId) -> QCCM.ChainState -> Bool)
+             -> DL state ()
+observeChain s p = QCCM.observe s p
 
 type Actions state = QCCM.Actions (CMI.WithInstances (WrappedState state))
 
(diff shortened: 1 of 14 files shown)


MaximilianAlgehed:
@sjoerdvisscher  do you want to have a look at this?

plutus-contract/src/Plutus/Contract/Test/ContractModel/Interface.hs
```diff
@@ -478,6 +479,14 @@ waitUntilDL = QCCM.waitUntilDL . toSlotNo
 assertModel :: String -> (QCCM.ModelState state -> Bool) -> DL state ()
 assertModel s p = QCCM.assertModel s (p . fmap coerce)
 
+-- IMPORTANT NOTE: the first argument needs to be unique among calls to `observeChain`
+-- as it is a proxy for the predicate.
+observeChain :: ContractModel state
+             => String
+             -> ((QCCM.SymToken -> CardanoAPI.AssetId) -> QCCM.ChainState -> Bool)
```
sjoerdvisscher, line 486: The rest of the Plutus `Contract` `ContractModel` API is still using `AssetClass` instead of `CardanoAPI.AssetId`, right? I'm fine with either, but it should be consistent.

sjoerdvisscher (approved):
Thanks, this seems a very useful addition!

sjoerdvisscher:
> One issue we have here is that we are a little bit blocked on this PR: [input-output-hk/quickcheck-contractmodel#13](https://github.com/input-output-hk/quickcheck-contractmodel/pull/13) as it would be nice if we could depend on the version of `quickcheck-contractmodel` in `CHaP` rather than on a source repository stanza. However, I wanted to get this PR up here so that the whole chain of things is at least visible.

It seems that that PR has a possibly complicated build error. I'm also fine with keeping it like this for now, and to start depending on the CHaP version later.

plutus-contract/src/Plutus/Contract/Test/ContractModel/Interface.hs
MaximilianAlgehed: Ah yes... It would be good if as we are transitioning the rest of plutus apps to cardano api we might as well transition the contract model. The current implementation is just a thin layer on top of a cardano api based implementation.

MaximilianAlgehed:
@sjoerdvisscher how close to merging this do you think we are? Do you understand what's up with the build?

MaximilianAlgehed:
@sjoerdvisscher I think this should be ready to merge. Although I don't know what's going on with CI timing out?

sjoerdvisscher:
No problem, it's just the darwin tests, the linux tests are fine. This happens from time to time. 

@koslambrou Could you please merge this?

PR #1066, opened 2023-05-05
Move ContractModel to its own cabal package

UlfNorell:
Currently the `ContractModel` stuff lives in the `plutus-contract` package. This is a little inconvenient if you have a project using `plutus-contract`, but not the `Contract` monad and `ContractModel` for testing and instead want to use `quickcheck-contractmodel` with a different backend (for instance cooked-validators). With the current situation you are forced to use the version of `quickcheck-contractmodel` pinned by `plutus-contract`.

This PR breaks out `ContractModel` to a new package `plutus-contract-model` making it possible to depend on `plutus-contract` independently of `quickcheck-contractmodel`.


MaximilianAlgehed:
@koslambrou can we have some eyes on this?

koslambrou:
> This PR breaks out ContractModel to a new package plutus-contract-model making it possible to depend on plutus-contract independently of quickcheck-contractmodel.

Ok, so the goal of this PR is to remove the `quickcheck-contractmodel` from `plutus-contract`. However, I'm not sure I understand the motivation described in the PR description. More specifically, I don't understand this sentence: "This is a little inconvenient if you have a project using plutus-contract, but not the Contract monad and ContractModel for testing and instead want to use quickcheck-contractmodel with a different backend (for instance cooked-validators)". Can you rephrase it?

If a user wants to use a different backend, how does that help him?

MaximilianAlgehed:
@koslambrou In this repo: https://github.com/Quviq/quickcheck-contractmodel-cooked we have a compatibility layer for quickcheck-contractmodel for cooked validators. Cooked validators depends on parts of apps (including `plutus-contract`), but *not* on the contract-model stuff. Therefore, when we make changes to `quickcheck-contractmodel` to develop the above package we end up having to also make changes to apps to keep everything in line - because `cooked-validators` doesn't build without `plutus-contract` which doesn't handle breaking changes to `quickcheck-contractmodel`. By separating out `plutus-contract-model` we decouple `quickcheck-contractmodel` from `plutus-contract` which in turn means that we can experiment with breaking changes without the overhead of having to fix both the `quickcheck-contractmodel-cooked` and `plutus-contract-model` downstreams at the same time. Likewise, if `cooked-validators` falls behind `plutus-apps` main branch we would like to be able to make changes to `quickcheck-contractmodel` without breaking *old* versions of `plutus-apps`.

koslambrou:
@sjoerdvisscher Anything to add?

sjoerdvisscher (approved):
💯

PR #1074, opened 2023-05-23
PLT-5926: Also output emulator logs on success.

### LINKED ISSUE #1063: Emulator log not working for QuickCheck and Plutus.Contract.Test.ContractModel
### Summary

After updating to plutus-apps v1.2.0, the ContractModel QuickCheck code no longer provides logs of the Emulator trace. This makes it almost impossible to figure out what the problem is with the property tests. I have talked with someone Quviq which said that the bug is likely to have been introduced in the emulator, since their side of the code concerning the ContractModel all seems in order.

- Introduce a bug in some existing model such as [Escrow](https://github.com/input-output-hk/plutus-apps/blob/main/plutus-use-cases/test/Spec/Escrow.hs) ContractModel. E.g. replace line 144 with `Trace.callEndpoint @"pay-escrow" (h $ WalletKey w) (Ada.adaValueOf $ fromInteger 10)`
- Run the code using `nix develop`, `cabal repl`, and then `quickCheck prop_Escrow`

### Actual Result

```
>quickCheck prop_Escrow
*** Failed! Falsified (after 1 test):                  
Actions 
 [Pay (Wallet 5) 17]
Balance changes don't match:
  Predicted symbolic balance changes: Wallet 5: {-17000000 Lovelace}
  Predicted actual balance changes: Wallet 5: {-17000000 Lovelace}
  Actual balance changes: Wallet 5: {-10000000 Lovelace}
  Sum of min Lovelace: Lovelace 1892090
```
The value you get is probably going to be something different than 17, but that doesn't really matter.

### Expected Result

This is for a different error that can be found on [this readthedocs tutorial](https://plutus-apps.readthedocs.io/en/latest/plutus/tutorials/contract-models.html), but a log should be appended at the end with the emulator trace.
```
>quickCheck prop_Escrow
*** Failed! Assertion failed (after 5 tests and 7 shrinks):
Actions
 [Init (Slot {getSlot = 0}) [],
  Pay (Wallet 1) 2]
Expected funds of W[1] to change by
  Value (Map [(,Map [("",-2000000)])])
but they did not change
Test failed.
Emulator log:
[INFO] Slot 0: TxnValidate ee3a44b98e0325e19bc6be1e6f25cdb269301666a3473758296e96cd7ea9a851
[INFO] Slot 1: 00000000-0000-4000-8000-000000000000 {Wallet W[1]}:
                 Contract instance started
[INFO] Slot 1: 00000000-0000-4000-8000-000000000001 {Wallet W[2]}:
                 Contract instance started
...
```

_No response_

### System info

OS: Ubuntu 20.04
Plutus-Apps v1.2.0

sjoerdvisscher:
Fixes #1063 

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-contract/src/Plutus/Contract/Test.hs
+++ b/plutus-contract/src/Plutus/Contract/Test.hs
@@ -305,16 +305,16 @@ checkPredicateInnerStream CheckOptions{_minLogLevel, _emulatorConfig}
         -- Also log on success. The test runner suppresses the output on success so this has no effect on succesful tests.
         -- But this makes `checkPredicateInnerStream` more modular: When another test function calls it,
         -- and that function has a failure, the logs will be shown.
-        _ <- logEmulator
+        void logEmulator
         return res
       Right (False S.:> res) -> do
         annot "Test failed."
-        _ <- logEmulator
+        void logEmulator
         assert False
         return res
       Left err -> do
         annot "Test failed."
-        _ <- logEmulator
+        void logEmulator
         annot "Error:"
         annot (describeError err)
         annot (show err)
--- a/plutus-contract/src/Plutus/Contract/Test.hs
+++ b/plutus-contract/src/Plutus/Contract/Test.hs
@@ -293,7 +293,6 @@ checkPredicateInnerStream CheckOptions{_minLogLevel, _emulatorConfig}
             $ consumedStream
 
     let logEmulator = do
-          annot "Test failed."
           annot "Emulator log:"
           S.mapM_ annot
               $ S.hoist runM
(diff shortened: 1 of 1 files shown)


plutus-contract/src/Plutus/Contract/Test.hs
```diff
@@ -302,12 +301,19 @@ checkPredicateInnerStream CheckOptions{_minLogLevel, _emulatorConfig}
               theStream
 
     case result of
-      Right (True  S.:> res) -> return res
+      Right (True  S.:> res) -> do
+        -- Also log on success. The test runner suppresses the output on success so this has no effect on succesful tests.
+        -- But this makes `checkPredicateInnerStream` more modular: When another test function calls it,
+        -- and that function has a failure, the logs will be shown.
+        _ <- logEmulator
```
berewt, line 308: I tend to prefer `void logEmulator` over `_ <- logEmulator`. If it's fine for you, you can update it in the other branches as well.

PR #1088, opened 2023-06-17
Add inline datum functionality for state machines

jonathanfishbein1:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/plutus-contract/plutus-contract.cabal
+++ b/plutus-contract/plutus-contract.cabal
@@ -243,6 +243,7 @@ test-suite plutus-contract-test
     Spec.Contract.Tx.Constraints.Spec
     Spec.Contract.Tx.Constraints.TimeValidity
     Spec.Emulator
+    Spec.InlineDatum
     Spec.Plutus.Contract.Oracle
     Spec.Rows
     Spec.Secrets
--- a/plutus-contract/src/Plutus/Contract/StateMachine.hs
+++ b/plutus-contract/src/Plutus/Contract/StateMachine.hs
@@ -68,8 +68,9 @@ import Ledger (POSIXTime, Slot, TxOutRef)
 import Ledger qualified
 import Ledger.Tx qualified as Tx
 import Ledger.Tx.Constraints (ScriptLookups, TxConstraints (txOwnInputs, txOwnOutputs), UnbalancedTx,
-                              mustMintValueWithRedeemer, mustPayToTheScriptWithDatumInTx, mustSpendOutputFromTheScript,
-                              mustSpendPubKeyOutput, plutusV2MintingPolicy)
+                              mustMintValueWithRedeemer, mustPayToTheScriptWithDatumInTx,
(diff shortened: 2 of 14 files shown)


berewt (changes requested):
Thanks for the submission! I still need to investigate the golden tests issues, as I don't understand yet how the changes impact the script address. Some minor refactoring can also improve the readability and ease the maintenance.
I asked also a review from @sjoerdvisscher who is more knowledgeable of this part of the codebase.
(while we're investigating the golden value mismatches, you can already run `cabal test --test-option=--accept` in `plutus-use-cases` and commit the changes to add the new outputs to the PR.)

sjoerdvisscher:
It makes sense that the golden tests fail, since this changes on-chain scripts.

After you make the changes @berewt suggests I'll do a final review, but this looks good overal!

berewt:
That's why it puzzled me: when the `inline` flag is `False`, i thought we would end with exactly the same constraints and tx as before, and thus to the same script.

sjoerdvisscher:
The `if`-statement that checks for the flag is also in on-chain code.

sjoerdvisscher:
I would be fine with only having the inline datum version.

berewt:
> I would be fine with only having the inline datum version.
It would indeed simplify the code and avoid this drawback. @jonathanfishbein1 would you be ok to change it accordingly? (it deprecates my review comments)

jonathanfishbein1:
Sure, just want to make sure I understand correctly, you want to replace `mustPayToTheScriptWithDatumInTx` with `mustPayToTheScriptWithInlineDatum` completely thereby making the if expression unnecessary as well as the `smInlineDatum :: Bool` field

berewt:
Yes, that's it. Overall, inlining is better than putting the datum in the tx, and due to the design of the constraint library, providing the choice would kill the potential benefit.

jonathanfishbein1:
ScriptHashes will be different even without the if expression because the constructors below are different

```{-# INLINABLE mustPayToTheScriptWithDatumInTx #-}
mustPayToTheScriptWithDatumInTx :: o -> Value -> TxConstraints i o
mustPayToTheScriptWithDatumInTx dt vl =
    mempty { txOwnOutputs = [ScriptOutputConstraint (TxOutDatumInTx dt) vl Nothing] }

{-# INLINABLE mustPayToTheScriptWithInlineDatum #-}
mustPayToTheScriptWithInlineDatum :: o -> Value -> TxConstraints i o
mustPayToTheScriptWithInlineDatum dt vl =
    mempty { txOwnOutputs = [ScriptOutputConstraint (TxOutDatumInline dt) vl Nothing] }
```

sjoerdvisscher:
Yes, but it’s not a problem that they are different!

berewt:
Yes we solved this misunderstanding on discord. Jonathan has opened a new PR, unfortunately, our CI doesn't want to forward its answer, gonna link to it in the PR.

PR #1097, opened 2023-07-17
Start of node compatible socket interface

sjoerdvisscher:
This is a work in progress, but I want to merge this since the change otherwise wont survive the move to a new repo.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/cardano-node-socket-emulator/src/Cardano/Node/Socket/Emulator/Chain.hs
+++ b/cardano-node-socket-emulator/src/Cardano/Node/Socket/Emulator/Chain.hs
@@ -14,6 +14,7 @@ module Cardano.Node.Socket.Emulator.Chain where
 
 import Cardano.Node.Emulator.Internal.Node (Params)
 import Cardano.Node.Emulator.Internal.Node.Chain qualified as EC
+import Cardano.Protocol.Socket.Type (BlockId (..), Tip, blockId)
 import Control.Concurrent
 import Control.Concurrent.STM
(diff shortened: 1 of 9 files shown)


cardano-node-socket-emulator/src/Cardano/Node/Socket/Emulator/Chain.hs
```diff
@@ -66,17 +69,13 @@ fromEmulatorChainState EC.ChainState {EC._txPool, EC._index, EC._chainCurrentSlo
                       , _txPool      = _txPool
                       , _index       = _index
                       , _currentSlot = _chainCurrentSlot
-                      , _tip         = listToMaybe _chainNewestFirst
+                      , _tip         = O.TipGenesis
```
koslambrou, line 72: Should it always be genesis?

koslambrou (approved):

cardano-node-socket-emulator/src/Cardano/Node/Socket/Emulator/Chain.hs
sjoerdvisscher: Oh good point! It was true in the way it was used, but not in general.

sjoerdvisscher: Fixed. Can you merge this?

PR #1103, opened 2023-08-09
[WIP] Integrate IOGX

zeme-wana:
*This PR is a work in progress, please don't look yet*
-----

As part of the ongoing effort to build a Developer Platform for IOG, the smart-contracts-dev-empowerment team is releasing [IOGX - a Flake Template for Haskell Projects at IOG](https://github.com/input-output-hk/iogx)

The vision is to provide a JSON-like, declarative interface to Nix, so that developers unfamiliar with the language may independently maintain and add to the Nix sources with minimum effort.

The goal is to integrate with all Smart Contracts repositories, and maybe more.

There are currently several other repositories using this: marlowe-cardano, antaeus, quickcheck-dynamic, marconi, stablecoin-plutus, dapps-certification.

-----

This PR is a full refactor of the Nix code. The old `__std__` folder is gone, and all the nix files are now in the `nix` folder.
While CI may be green, there are a few things I unfortunately I cannot test, namely x86 linux stuff.
Most importantly, please do check that the following are still working:
- `haskell-language-server` in your IDE
- That the `devcontainer-docker-image` works:
  `docker load < $(nix build .#devcontainer-docker-image)`

You might find that the shell looks a little different, if you don't like it you can tell me.
Notably, the `fix-*formatter*` scripts have gone, because the same result can be obtained just by using `pre-commit`.
E.g.
```bash
fix-stylish-haskell
fix-shellcheck
# -> Now becomes
pre-commit run stylish-haskell
pre-commit run shellcheck

# To run all formatters
pre-commit run

# To run on all files (not just the committed ones)
pre-commit run --all-files
```

---- 

Various documentation-related files will also have to be updated.
I'm preparing a separate PR for that


plutus-script-utils/plutus-script-utils.cabal
```diff
@@ -110,7 +110,7 @@ library
   if flag(defer-plugin-errors)
     ghc-options: -fplugin-opt PlutusTx.Plugin:defer-errors
 
-test-suite plutus-ledger-test
+test-suite plutus-ledger-test2
```
zeme-wana, line 113: There is another executable with the exact same name in the `plutus-ledger` package.
This is a problem because the new flake exports all Haskell executables by their "short" name (i.e. without the package name prefixed to them). 
This means that one will override the other. 
Please suggest a new name for this executable or the one in `plutus-ledger`

nix/haskell.nix
```diff
@@ -0,0 +1,41 @@
...
+  defaultChangelogPackages = [ ];
+
+  # Because of lack of resources in our darwin CI cluster, the "combined-plutus-apps-haddock" job
+  # often gets timed out, which of course makes the required darwin CI checks fail. In the end, we
+  # just remove it from the required darwin jobs, but still keep it in the linux jobs.
+  enableCombinedHaddock = system == "x86_64-linux";
+
+
+  projectPackagesWithHaddock = [
+    "plutus-core"
```
zeme-wana, line 22: These were copied from plutus.
Which of `plutus-apps`'s cabal projects have haddock in them?

cabal.project
```diff
@@ -11,8 +11,6 @@ repository cardano-haskell-packages
     c00aae8461a256275598500ea0e187588c35a5d5d7454fb57eac18d9edb86a56
     d4a35cd3121aa00d18544bb0ac01c3e1691d618f462c46129271bccf39f7e8ee
 
-index-state: 2023-04-20T00:00:00Z
```
zeme-wana, line 14: This is redundant with newer versions of haskell.nix

nix/per-system-outputs.nix
```diff
@@ -0,0 +1,24 @@
+# This file is part of the IOGX template and is documented at the link below:
+# https://www.github.com/input-output-hk/iogx#35-nixper-system-outputsnix
+
+{ nix, projects, ... }:
+
+rec {
+
+  packages.generated-puc-scripts-output =
+    nix.plutus-apps.generated-puc-scripts-outputs projects.default;
```
zeme-wana, line 9: There is a parameters to the interface files named `nix`, which can be used to reference files in your `./nix` folder by name using attribute sets, instead of using the `import` keyword. See this for documentation https://github.com/input-output-hk/iogx#301-nix

zeme-wana: You can support multiple ghc compilers in your project. (See `./nix/cabal-project.nix:supportedCompilers`).

So the `projects` argument will contain an attribute for each compiler name.
However you can use the `default` project which will use the `./nix/cabal-project.nix:defaultCompiler` (which is set the first compiler in `supportedCompilers` by default)

See here for more docs https://github.com/input-output-hk/iogx#351-projects

zeme-wana:
closing in favour of https://github.com/input-output-hk/plutus-apps/pull/1110

PR #1110, opened 2023-10-23
Integrate IOGX v4

zeme-wana:
As part of the ongoing effort to build a Developer Platform for IOG, the smart-contracts-dev-empowerment team is releasing [IOGX - a Flake Template for Haskell Projects at IOG](https://github.com/input-output-hk/iogx)

The vision is to provide a JSON-like, declarative interface to Nix, so that developers unfamiliar with the language may independently maintain and add to the Nix sources with minimum effort.

The goal is to integrate with all Smart Contracts repositories, and maybe more.

There are currently several other repositories using this: marlowe-cardano, antaeus, quickcheck-dynamic, marconi, stablecoin-plutus, dapps-certification.

-----

This PR is a full refactor of the Nix code. The old `__std__` folder is gone, and all the nix files are now in the `nix` folder.
While CI may be green, there are a few things I unfortunately I cannot test, namely x86 linux stuff.
Most importantly, please do check that the following are still working:
- `haskell-language-server` in your IDE
- That the `devcontainer-docker-image` works:
  `docker load < $(nix build .#devcontainer-docker-image)`

You might find that the shell looks a little different, if you don't like it you can tell me.
Notably, the `fix-*formatter*` scripts have gone, because the same result can be obtained just by using `pre-commit`.
E.g.
```bash
fix-stylish-haskell
fix-shellcheck
# -> Now becomes
pre-commit run stylish-haskell
pre-commit run shellcheck

# To run all formatters
pre-commit run

# To run on all files (not just the committed ones)
pre-commit run --all-files
```

---- 

Various documentation-related files will also have to be updated.
I'm preparing a separate PR for that


.readthedocs.yml
```diff
@@ -1,15 +1,14 @@
-
 version: 2
```
zeme-wana, line 1: These changes are necessary as the read-the-docs site itself has changed its api.
See[ this build failure](https://readthedocs.org/projects/plutus-apps/builds/22430196/) for example

doc/requirements.txt
```diff
@@ -2,7 +2,6 @@ Sphinx==3.1.1
 sphinx-intl==2.0.1
 transifex-client==0.14.4
 testresources==2.0.1
--e git+https://github.com/input-output-hk/sphinx_rtd_theme.git#egg=sphinx_rtd_theme
```
zeme-wana, line 5: not sure how this ended up here or why it didn't break the build

koslambrou (approved):

PR #1112, opened 2023-11-08
Update nix documentation in CONTRIBUTING.adoc and RELEASE.adoc

zeme-wana:
Instructions on how to install & upgrade nix and how to setup the binary caches are now included in the [relevant section](https://github.com/input-output-hk/iogx/blob/main/doc/api.md#flakenix) of the IOGX manual. 
This PR updates the relevant sections of the README/CONTRIBUTING/RELEASE documents.

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -14,13 +14,9 @@ If you _really_ cannot use Nix and still want to contribute to Plutus then xref:
 
 === Installing and setting up Nix
 
-The nix code is based on a template from the the 
-[IOGX flake](https://github.com/input-output-hk/iogx). 
+For instructions on how to install and configure nix (including how to enable access to our binary caches), refer to [this document](https://github.com/input-output-hk/iogx/blob/main/doc/nix-setup-guide.md). 
 
-For instructions on how to install and configure nix (including how to enable 
-access to our binary caches), and how keep the flake inputs up to date, refer to 
-[this section](https://github.com/input-output-hk/iogx/blob/main/doc/api.md#flakenix)
-of the IOGX manual.
+If you already have nix installed and setup, you may enter the development shell by running `nix develop`.
 
 === Note on pre-commit hooks
 
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -14,9 +14,9 @@ If you _really_ cannot use Nix and still want to contribute to Plutus then xref:
 
 === Installing and setting up Nix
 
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.adoc
```diff
@@ -14,87 +14,23 @@ If you _really_ cannot use Nix and still want to contribute to Plutus then xref:
...
 
-This project uses Flakes, so you must have Nix version `>= 2.4` installed.
+The nix code is based on a template from the the 
+[IOGX flake](https://github.com/input-output-hk/iogx). 
 
-Either upgrade or install https://nixos.org/nix/[Nix] following the instructions on https://nixos.org/nix/[its website].
+For instructions on how to install and configure nix (including how to enable 
+access to our binary caches), and how keep the flake inputs up to date, refer to 
+[this section](https://github.com/input-output-hk/iogx/blob/main/doc/api.md#flakenix)
+of the IOGX manual.
```
koslambrou, line 24: I think you removed too much. I would keep basic explaination of how to enter a Nix shell using `nix develop` and how to setup the cache. I feel like the link you provided does not explain it adequatly enough (too general explanation). Also, there's no explaination on how to setup the cache which should really be upfront in the CONTRIBUTING.adoc.

Additionnaly, I like how the following paragraphs help the user get everything in order.

michaelpj: I admit, I didn't check the doc in iogx but I assumed it was at least as comprehensive. That said, I do agree that it's better to duplicate a bit here. It's hard enough to get people to click through to CONTIBUTING, if they search for "shell" and don't find anything they may just give up.

zeme-wana: I have created a separate document with comprehensive instructions on installing and setting up nix (including how to setup the cache) [here](https://github.com/input-output-hk/iogx/blob/main/doc/nix-setup-guide.md), and I have updated this CONTRIBUTING to point to that document and mention how to enter the shell. I think that document can now act as the SSOT, or do you still think I should duplicate some parts of it here?

koslambrou: Yes, some duplication will be necessary. At a minimum, we should keep the warning about the binary cache. So I would keep the following paragraph:

```
However you should pay attention to the output of `nix develop`: if you notice that you are _building_ GHC, it is likely that your caches have not been configured correctly.
```

Additionally, maybe the duplication part should be handled by the people responsible for repo. So I would say I'm fine with the current changes, but we will restructure this document to give more context to users.

